PR #1, opened 2022-02-10
Add skeleton of `xcodeproj_rules_dependencies()`

brentleyjones:


xcodeproj/repositories.bzl
```diff
@@ -0,0 +1,66 @@
...
+                    tag = kwargs.get("tag"),
+                    url = kwargs.get("remote"),
+                )
+                existing = "{url} (tag {tag})".format(
+                    tag = existing_repo.get("tag"),
+                    url = existing_repo.get("remote"),
+                )
+
+                warn("""\
+`com_github_buildbuddy_io_rules_xcodeproj` depends on `{repo}` loaded from \
```
tylerwilliams, line 42: nit: do you want to define a common "name" var somewhere and reference everywhere to ease the move of this into another repo down the line?

tylerwilliams (approved):
LGTM

xcodeproj/repositories.bzl
brentleyjones: I don't think we could use that in `load` or `workspace` statements, and besides those this is the only place we spell it out. Should be easy enough to do the replacement when we do a move.

PR #2, opened 2022-02-10
Add compiler and linker option processing

brentleyjones:


brentleyjones:
The TODOs are going to stay for now. I'm not sure how many I'll address before the initial release. Whatever remain before that will get tracked in GitHub issues.

brentleyjones:
I'm also looking into if there is a way to test these functions. I've been testing them with yet-to-be-PR'ed code, but ideally I could unit test them in some way. Starlark tests are tricky though.

xcodeproj/internal/opts.bzl
```diff
@@ -0,0 +1,428 @@
...
+    base_cxxopts = cc_common.get_memory_inefficient_command_line(
+        feature_configuration = feature_configuration,
+        action_name = "c++-compile",
+        variables = variables,
+    )
+
+    # TODO: Handle perfileopts somehow?
+
+    if SwiftInfo in target:
+        # Rule level swiftcopts are include in action.argv below
```
tylerwilliams, line 108: s/include/included/g

xcodeproj/internal/opts.bzl
```diff
@@ -0,0 +1,428 @@
...
+    "-isysroot": 2,
+    "-mios-simulator-version-min": 1,
+    "-miphoneos-version-min": 1,
+    "-mmacosx-version-min": 1,
+    "-mtvos-simulator-version-min": 1,
+    "-mtvos-version-min": 1,
+    "-mwatchos-simulator-version-min": 1,
+    "-mwatchos-version-min": 1,
+    "-target": 2,
+}
```
tylerwilliams, line 19: This feels a little sketch but you know better than I do. Are there cases where a command might be either, and do we need to handle that?

A la:
 --miphoneos-version-min=11.0
 --miphoneos-version-min 11.0

tylerwilliams (approved):
LGTM w/ 2 small comments

xcodeproj/internal/opts.bzl
brentleyjones: The bazel produced ones are always with `=`. I'll check, but I don't think without is supported.

brentleyjones: Yeah, they only support `=`. I believe that is the case for any of these flags, they only support one way.

PR #4, opened 2022-02-14
Set `SWIFT_VERSION`

brentleyjones:


xcodeproj/internal/opts.bzl
```diff
@@ -289,6 +294,9 @@ def _process_swiftcopts(opts, build_settings):
         if compilation_mode:
             build_settings["SWIFT_COMPILATION_MODE"] = compilation_mode
             return True
+        if opt.startswith("-swift-version="):
+            build_settings["SWIFT_VERSION"] = opt[15:]
```
tylerwilliams, line 298: I wonder if it would be cleaner / less error prone to do something like this:

if opt.startswith("-swift-version="):
            build_settings["SWIFT_VERSION"] = opt.replace("-swift-version=", "")


You could even define a helper that does both in one so you didn't have to write out the string twice:
val, was_set = read_flag("-swift-version=")
if was_set {
  build_settings["SWIFT_VERSION"] = val
}

tylerwilliams (approved):
LGTM

xcodeproj/internal/opts.bzl
brentleyjones: Going to keep as is, after adding a unit test.

PR #6, opened 2022-02-17
Add `PBXProject` creation

brentleyjones:
It's exercised only through tests right now. I'll be adding piece by piece to this before finally hooking it up in a binary.

### DIFF
--- a/tools/generator/src/DTO.swift
+++ b/tools/generator/src/DTO.swift
@@ -1,0 +1,3 @@
+struct Project: Equatable, Decodable {
+    let name: String
+}
--- a/tools/generator/src/Generator.swift
+++ b/tools/generator/src/Generator.swift
@@ -1,0 +1,52 @@
+import XcodeProj
+
+class Generator {
+    static let defaultEnvironment = Environment(
+        createProject: Generator.createProject
+    )
+
+    static func createProject(project: Project) -> (PBXProj, PBXProject) {
+        let pbxProj = PBXProj()
+
+        let mainGroup = PBXGroup()
+        pbxProj.add(object: mainGroup)
+
+        let buildConfigurationList = XCConfigurationList()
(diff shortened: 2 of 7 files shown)


tools/generator/src/DTO.swift
```diff
@@ -0,0 +1,3 @@
+struct Project: Equatable, Decodable {
```
brentleyjones, line 1: This file gets a lot bigger, but I wanted to add to it as more tests are added, to keep the churn down if things change while tests/refactoring is happening.

tools/generator/src/Generator.swift
```diff
@@ -0,0 +1,52 @@
...
+            "LastUpgradeCheck": 1320,
+        ]
+
+        let pbxProject = PBXProject(
+            name: project.name,
+            buildConfigurationList: buildConfigurationList,
+            // TODO: Calculate `compatibilityVersion`
+            compatibilityVersion: "Xcode 13.0",
+            mainGroup: mainGroup,
+            // TODO: Make regions configurable?
```
brentleyjones, line 30: All three of these TODOs are staying for now. Not sure if they will be resolved before MVP (low priority).

tools/generator/src/Generator.swift
```diff
@@ -0,0 +1,52 @@
...
+        pbxProj.add(object: mainGroup)
+
+        let buildConfigurationList = XCConfigurationList()
+        pbxProj.add(object: buildConfigurationList)
+
+        let attributes = [
+            "BuildIndependentTargetsInParallel": 1,
+            // TODO: Generate these
+            "LastSwiftUpdateCheck": 1320,
+            "LastUpgradeCheck": 1320,
```
tylerwilliams, line 21: nit: comment that this is xcode 13.2.0 in int form?

tylerwilliams (approved):
LGTM

PR #7, opened 2022-02-17
Extract `Generator.createProject()`

brentleyjones:
In the future there will be a lot more functions in `Generator` and with the current stucture `Generator.swift` and `GeneratorTests.swift` would explode in size.

So here I'm extracting `Generate.createProject()` to it's own file and `test_createProject()` to it's own test (as `CreateProjectTests.test_basic()`). In the future we can add more tests as various Project customizations become available.


tylerwilliams (approved):
LGTM

PR #10, opened 2022-02-21
Improve display of assert diffs for XcodeProj objects

brentleyjones:


tools/generator/test/XcodeProj+CustomDump.swift
```diff
@@ -0,0 +1,115 @@
+import CustomDump
+
+@testable import XcodeProj
+
+extension PBXObjects: CustomDumpReflectable {
+    private static let ignoredProperties: Set<String> = [
+        // The lock is always different
```
tylerwilliams, line 7: nit: maybe "Ignore lock property which is always different"

tylerwilliams (approved):
LGTM

PR #12, opened 2022-02-21
Add `Target` and `Generator.processTargetMerges()`

brentleyjones:


tools/generator/src/BuildSetting.swift
```diff
@@ -0,0 +1,62 @@
...
+    /// becomes an array of `String`s).
+    var asAny: Any {
+        switch self {
+        case .string(let string):
+            return string
+        case .bool(let bool):
+            return bool ? "YES" : "NO"
+        case .array(let array):
+            if array.count < 2 {
+                return array.first ?? ""
```
tylerwilliams, line 24: ok, i'm dumb, can you explain this to me?

tylerwilliams (approved):
LGTM

tools/generator/src/BuildSetting.swift
brentleyjones: If an array has 1 or 0 elements, Xcode wants it to be the same as a string.

brentleyjones: I'll minimally add a comment.

PR #14, opened 2022-02-22
Don't return `PBXProject` from `createProject()`

brentleyjones:
The `PBXProject` is cannonically accessed through `PBXProj.rootObject`.


tools/generator/src/Environment.swift
```diff
@@ -6,7 +6,7 @@ import XcodeProj
 /// The main purpose of `Environment` is to enable dependency injection,
 /// allowing for different implementations to be used in tests.
 struct Environment {
```
tylerwilliams, line 8: love a good environment

tylerwilliams (approved):
LGTM

PR #16, opened 2022-02-23
Allow merging into multiple destinations

brentleyjones:
A single library can be used by multiple executable targets, so we need to allow merging into multiple destinations.


tylerwilliams (approved):
LGTM

PR #18, opened 2022-02-23
Conform `PreconditionError` to `LocalizedError`

brentleyjones:
`LocalizedError` allows us to easily display a user facing message for errors. I've included a call to action to file a bug if a user sees a `PreconditionError`.

There will be more errors in the future, so I also moved it to its own file.


tylerwilliams (approved):
LGTM

PR #19, opened 2022-02-24
Pass `logger` directly to `Generator`

brentleyjones:
We will want to use the same instance of `Logger` outside of the `Generator` instance, so having it be part of `Environment` makes that tricky.


tylerwilliams (approved):
LGTM

PR #27, opened 2022-02-24
Update XcodeProj to fix dictionary compares

brentleyjones:
https://github.com/tuist/XcodeProj/pull/667

PR #35, opened 2022-02-28
Set target configurations

brentleyjones:
A test fails without #27, but knowing that will be fixed so (one way or another), I think it's fine to land this.


tylerwilliams (approved):
LGTM

PR #45, opened 2022-03-01
Add `xcodeproj` rule

brentleyjones:


brentleyjones:
When reviewing, you can generally ignore what is under `//examples/ios_app`, as that is just an example project that `xcodeproj` is run on. The main stuff to review is outside of that directory.

tylerwilliams (approved):
LGTM

PR #50, opened 2022-03-02
Don't generate a target for rules that don't have outputs

brentleyjones:
This causes rules like `apple_dynamic_framework_import` to not be processed as targets. Projects with rules like that will still be incomplete, as we need still need to add support for adding those links and embeds.

I also added an earlier failure for when this doesn't prevent processing.


tylerwilliams (approved):
LGTM

PR #53, opened 2022-03-04
Fix fixtures for Intel Macs

brentleyjones:
We transition `--cpu` to `darwin_x86_64` for the fixture tests.


brentleyjones:
Workflow failure is because #28 isn't merged yet. This change is needed to unblock it.

PR #54, opened 2022-03-04
Output a relative path for `projectRootDir`

brentleyjones:
This matches Xcode's default behavior. It also makes fixture tests more deterministic.

Also changes the fixtures to use fixture-based `xcodeproj` targets. This allows the `install_path` to be correct when generating the project instead of only when installing it.


brentleyjones:
Workflow failure is because #28 isn't merged yet.

brentleyjones:
Workflow failure is something else that I have to fix.

tylerwilliams (approved):
LGTM

PR #56, opened 2022-03-04
Set `externalDirectory` to a relative path in fixtures

brentleyjones:
We don't use a relative path by default because the symlinks in `bazel-WORKSPACE/external` change with each build, causing the files to become unavailable in Xcode from time to time. I'm fine with making the use of the fixture projects a little more finicky in order to have deterministic fixtures.


tylerwilliams (approved):
LGTM

PR #59, opened 2022-03-07
Add basic files and groups test

brentleyjones:
This shows that when special groups, like the Bazel External Repositories aren't needed, they aren't created.


tylerwilliams (approved):
LGTM

PR #63, opened 2022-03-07
Push `FilePath` creation out to DTO

brentleyjones:
This is in preperation adding more information to `FilePath` from the spec, such as if the input was generated or in an external repository.


tylerwilliams (approved):
LGTM

PR #65, opened 2022-03-07
Add `FilePath.PathType.generated` case

brentleyjones:
Part of https://github.com/buildbuddy-io/rules_xcodeproj/issues/100.

Similar to external repositories, this allows `bazel-out` to be a normal directory in the workspace (as long as `--symlink_prefix` is overridden).

We show the generated files in a special group after the external repositories.


tylerwilliams (approved):
LGTM

woo perl

PR #110, opened 2022-03-08
Refactor return type of `createFilesAndGroups()`

brentleyjones:
This cleans up the logic in `writeXcodeProj()`, which unlocks the ability to write other internal files with non-empty content.


tylerwilliams (approved):
LGTM

PR #112, opened 2022-03-08
Loosen the requirement around top-level deps slightly

brentleyjones:
Older rules_apple versions would put the entitlements processing as a dep of `ios_application`. This works around that.


brentleyjones:
I had someone test this where they were running into issues, and it now gets past this check (though fails on something else).

Issue #83, opened 2022-03-08
Support `cc_binary`


brentleyjones:
https://bazel.build/reference/be/c-cpp#cc_binary

- [x] Add example and fixture test: https://github.com/buildbuddy-io/rules_xcodeproj/pull/133
- [x] Support `additional_linker_inputs`
- [x] Support `includes`
- [ ] Support `linkshared`
- [ ] Support `linkstatic`
- [x] Support `local_defines`
- [ ] Support `malloc`?

brentleyjones:
We are missing an example for `linkstatic`/`linkshared`/`malloc`, but unless someone reports an issue for them, I'll assume we support them.

PR #120, opened 2022-03-10
Fix product path detection for cc_binary

brentleyjones:
Part of #83.

A test showing this working will come later.


tylerwilliams (approved):
LGTM

Issue #82, opened 2022-03-08
Support `cc_library`


brentleyjones:
https://bazel.build/reference/be/c-cpp#cc_library

- [x] Add example and fixture test: https://github.com/buildbuddy-io/rules_xcodeproj/pull/133
- [x] Support `alwayslink`
- [x] Support `defines`
- [x] Support `include_prefix`: https://github.com/buildbuddy-io/rules_xcodeproj/pull/1656
- [x] Support `includes`: #310
- [x] Support `implementation_deps`: https://github.com/buildbuddy-io/rules_xcodeproj/pull/1664
- [ ] Support `linkstatic`
- [x] Support `local_defines`
- [x] Support `nocopts` (support removed in Bazel, so not doing anything for it un rules_xcodeproj)
- [x] Support `strip_include_prefix`: https://github.com/buildbuddy-io/rules_xcodeproj/pull/1656

cgrindel:
@brentleyjones As I mentioned in Slack, the issue with keith/swift-syntax-bazel not working is related to [`rules_swift` support for `cc_library` being consumed as a Swift module, if it is tagged](https://github.com/bazelbuild/rules_swift/blob/2a20daa7ce087195b3fe0dbde78f7c5209526946/swift/internal/swift_c_module.bzl#L108). Do you want me to track that work on this issue? If not, I will create another one.

brentleyjones:
No need for a new issue, you can relate the PR to this one and update the description to link back to that PR. Thanks!

brentleyjones:
We are missing an example for `linkstatic`, but unless someone reports an issue for it, I'll assume we support it.

PR #121, opened 2022-03-10
Use `CcInfo` instead of `apple_common.ObjC`

brentleyjones:
Part of #82 .

This allows us to pick up static libaries produced by `cc_library`.


tylerwilliams (approved):
LGTM

PR #125, opened 2022-03-10
Don't try to merge with srcs or multiple deps

brentleyjones:
Part of #83.

`cc_binary` and `swift_binary` can specify their own sources. When that happens we don't want to merge any of their library depenedencies into them. We might want to go a step futher and only try to merge for the bundling rules, but we can decide on that at a later time.

Also, instead of erroring out when there are multiple static library dependencies, it just no longer tries to merge that top-level target. Though, if it's a bundling target, we still error.


tylerwilliams (approved):
LGTM

Issue #126, opened 2022-03-10
Feature Request: Enable Bazel Starlark Linting and Formatting


cgrindel:
## Tasks
- [x] Add Bazel target that can be used to check for lint and formatting issues using Buildifier
- [x] Add Bazel target that can fix lint and formatting issues using Buildifier.
- [ ] Update the existing codebase to address existing lint and formatting issues.
- [ ] Update the CI pipeline to execute the check target.

## Suggested Implementation
- First PR
  - Add [keith/buildifier-prebuilt](https://github.com/keith/buildifier-prebuilt) as a dev-only dependency.
  - Implement two runnable Bazel targets: one to check the lint issues and one to fix issues.
- Second PR
  - Fix all of the existing lint issues 
  - Add another entry to `bazel_commands` in the `buildbuddy.yml` file to execute the check Bazel target.

PR #127, opened 2022-03-10
Add initial buildifier check and fix commands

cgrindel:
Related to #126.

- Added dev-only dependency on `keith/buildifier-prebuilt`.
- Added two runnable targets to the root of the workspace: `//:buildifier.check` and `//:buildifier.fix`.

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -21,4 +21,4 @@ actions:
   #       branches:
   #         - "main"
   #   bazel_commands:
-  #     - "run //:buildifier.check"
+  #     - "run --config=workflows //:buildifier.check"
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -10,3 +10,5 @@ actions:
           - "main"
     bazel_commands:
       - "test --config=workflows //tools/... //test/..."
+      # GH126: Enable the Buildifier check once the current lint issues are addressed.
+      # - "run //:buildifier.check"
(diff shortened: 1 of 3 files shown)


buildbuddy.yaml
```diff
@@ -10,3 +10,5 @@ actions:
           - "main"
     bazel_commands:
       - "test --config=workflows //tools/... //test/..."
+      # GH126: Enable the Buildifier check once the current lint issues are addressed.
+      # - "run //:buildifier.check"
```
brentleyjones, line 14: We can probably put this under a new action, and not limit it to `darwin`, to allow it to run on linux.

cgrindel: I moved it to its own action. Does that look about right? Of course, we won't enable this until the lint issues are addressed.

buildbuddy.yaml
```diff
@@ -10,3 +10,15 @@ actions:
...
+  # - name: Lint
+  #   triggers:
+  #     push:
+  #       branches:
+  #         - "main"
+  #     pull_request:
+  #       branches:
+  #         - "main"
+  #   bazel_commands:
+  #     - "run //:buildifier.check"
```
brentleyjones, line 24:
```suggestion
  #     - "run --config=workflows //:buildifier.check"
```

brentleyjones (approved):
Thank you @cgrindel!

buildbuddy.yaml
cgrindel: Done.

PR #129, opened 2022-03-11
Refactor `process_opts()` to take in `build_settings`

brentleyjones:
This makes it match the API of the other functions, and allows it to not be the first function to set build settings.

### DIFF
--- a/xcodeproj/internal/opts.bzl
+++ b/xcodeproj/internal/opts.bzl
@@ -464,19 +464,20 @@ def _xcode_std_value(std):
 
 # API
 
-def process_opts(*, ctx, target):
+def process_opts(*, ctx, target, build_settings):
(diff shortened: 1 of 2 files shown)


xcodeproj/internal/opts.bzl
```diff
@@ -464,19 +464,20 @@ def _xcode_std_value(std):
 
 # API
 
-def process_opts(*, ctx, target):
+def process_opts(*, ctx, target, build_settings):
```
cgrindel, line 467: I noticed that the usage later in this PR passes an empty `dict`. Is there a reason that we can't just default `build_settings` to an empty dict?

brentleyjones: Because it's used as an `inout` parameter. If it's defaulted then the user doesn't get the result back.

Issue #84, opened 2022-03-08
Support `cc_test`


brentleyjones:
https://bazel.build/reference/be/c-cpp#cc_test

- [x] Support `additional_linker_inputs`
- [ ] Support `linkstatic`
- [ ] Support `malloc`?
- [x] Support `nocopts` (support removed in Bazel, so not doing anything for it un rules_xcodeproj)

brentleyjones:
We are missing an example for `linkstatic`/`malloc`, but unless someone reports an issue for them, I'll assume we support them.

Issue #69, opened 2022-03-08
Support `objc_library`


brentleyjones:
https://bazel.build/reference/be/objective-c#objc_library

- [x] https://github.com/buildbuddy-io/rules_xcodeproj/issues/68
- [x] https://github.com/buildbuddy-io/rules_xcodeproj/pull/147
- [x] Support `alwayslink`: #608 
- [x] Support `pch`: https://github.com/buildbuddy-io/rules_xcodeproj/pull/280
- [ ] Support `runtime_deps`
- [x] Support `sdk_dylibs`: #372
- [x] Support `sdk_frameworks`: https://github.com/buildbuddy-io/rules_xcodeproj/pull/278
- [x] Support `sdk_includes`: https://github.com/buildbuddy-io/rules_xcodeproj/pull/1660
- [x] Support `weak_sdk_frameworks`: https://github.com/buildbuddy-io/rules_xcodeproj/pull/278

maxwellE:
@brentleyjones I think some of this issues' sub tasks are now completed

brentleyjones:
I checked off the ones I'm sure of. We need examples of the remaining ones.

brentleyjones:
While we don't have an example for `runtime_deps`, unless a bug is reported, I'm claiming this is done.

PR #132, opened 2022-03-11
Set user header search paths for non-Swift targets

brentleyjones:
Part of #69, #82, #83, and #84.

This will be shown off in a future change when a `cc_binary` example is checked in.

### DIFF
--- a/tools/generator/src/Generator+SetTargetConfigurations.swift
+++ b/tools/generator/src/Generator+SetTargetConfigurations.swift
@@ -1,3 +1,4 @@
+import PathKit
 import XcodeProj
 
 extension Generator {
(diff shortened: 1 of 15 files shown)


tools/generator/src/Generator+SetTargetConfigurations.swift
```diff
@@ -76,3 +88,45 @@ Test host with id "\(testHostID)" not found
...
+
+private extension Path {
+    /// Wraps the path in quotes if it needs it
+    var quotedString: String {
+        guard string.rangeOfCharacter(from: .whitespaces) != nil else {
+            return string
+        }
+        return #""\#(string)""#
+    }
+}
```
tylerwilliams, line 132: nice

tylerwilliams (approved):
LGTM

PR #138, opened 2022-03-11
Refactor libraries collection

brentleyjones:
We now pass around the linker inputs until we need to grab the static libraries out of it. In a future change this allows for links to be propagated through targets that don't produce their own Xcode target, like `objc_library` targets that only contain headers.

Also, in the future we will probably want to use the other values in the linker inputs, such as the dylib and framework references.


tylerwilliams (approved):
LGTM

PR #141, opened 2022-03-12
Properly handle targets that don't create an Xcode target

brentleyjones:
The existing logic in `_should_process_target()` was written to handle the case of `apple_resource_bundle`, since we currently aren't creating an Xcode target for it. It was updated to also consider the case of import targets, as those won't be Xcode targets, and just provide information.

Both of those cases were leaf targets, so we didn't propagate information from their dependencies. There are intermediary targets that we need to consider though, an easy example being a headers only `objc_library`. In those cases, we want to collect information from the target, but not create an Xcode target for it. With this change, we now do just that.

This is different from the `_passthrough_target()` (now renamed to `_skip_target()`) case, where we don't want any information from the target being processed. Maybe that case should no longer exist though, in favor of the new functionality of  `_process_non_xcode_target()`.


tylerwilliams (approved):
LGTM

Issue #115, opened 2022-03-09
Feature Request: Add Bazel integration tests


cgrindel:
## What

Implement Bazel integration tests that confirm that the functionality in this project works with select versions of Bazel.

## How

- [x] Add [rules_bazel_integration_test](https://github.com/cgrindel/rules_bazel_integration_test) as a dev-only dependency.
- [x] Implement an integration test based upon the `examples/ios_app` in a similar manner to the existing fixture tests.
- [x] Define an integration test suite to execute the integration tests.

I will put together one or more PRs with the above functionality. I will reference this issue.

cgrindel:
@brentleyjones How important is it to support paths with spaces? It causes Bazel many headaches as referenced by bazelbuild/bazel#4327. Do you think that we could recommend a project layout that does not include spaces? It looks like the primary issue is with `Preview Content` and `Preview Assets.xcassets`. I am not sure if these are magical names for Xcode. However, if we can avoid them, that would be awesome. 

BTW, I can move forward with the integration test by renaming the paths. [Here](https://github.com/buildbuddy-io/rules_xcodeproj/pull/144) is the WIP.

cgrindel:
Great. I will plan to rename the directories for the integration test. I will not change the directories for the example.

Issue #134, opened 2022-03-11
Address Bazel Starlark lint warnings about module visibility


cgrindel:
There are a number of `# buildifier: disable=bzl-visibility` directives sprinkled throughout the codebase. The primary issue is that by convention code under `private` or `internal` directories should not be accessible in other parts of the codebase. We should discuss how we want visibility to work in the repository and then update the code so that these directives can be removed.

cgrindel:
I spent a little time looking at ways to address the visibility warnings. If we want to legitimately remove the directives, we would have to 

1. Move some of the packages to not be under private.
2. Expose them by re-exporting them like we did with `testing.bzl`. 
3. Move the tests under the `internal` package.

Ideally, Buildifier would support references to internal APIs under a top-level `test` or `tests` directory. In the meantime, we can do one of the following:

1. Leave the directives.
2. Disable the bzl-visibility warning (once #176 is addressed).

PR #144, opened 2022-03-14
Add Bazel integration test

cgrindel:
Related to #115.

- Added dev-only dependency to [cgrindel/rules_bazel_integration_test](https://github.com/cgrindel/rules_bazel_integration_test).
- Implemented a basic Bazel integration test by copying the `examples/ios_app` files to `test/integration/ios_app`. Right now, this test runs `info`, `test`, and `run` actions on the child workspace. We can add a custom test runner that verifies certain output as a follow up PR.


buildbuddy.yaml
```diff
@@ -9,7 +9,7 @@ actions:
         branches:
           - "main"
     bazel_commands:
-      - "test --config=workflows //tools/... //test/..."
+      - "test --config=workflows //tools/... //test/... //test/integration:all_integration_tests"
```
brentleyjones, line 12: This shouldn't be needed, right? Doesn't `//test/...` cover it?

bazel_versions.bzl
```diff
@@ -0,0 +1,11 @@
+"""Specifies the supported Bazel versions."""
+
+CURRENT_BAZEL_VERSION = "5.0.0"
+
+OTHER_BAZEL_VERSIONS = [
+    "4.2.2",
```
brentleyjones, line 6: We don't support below 5.0.0. Is there a way to add support for HEAD/rolling?

brentleyjones (commented):
Is there a way to re-use `//examples/` instead of copying them over? Ideally I could edit `//examples` and the integration tests will be updated, similar to the fixture tests.

buildbuddy.yaml
cgrindel: Right now, I the tests are configured as `manual` and `exclusive`. This can be changed.  I keep them `manual` for my projects so that `bazel test //...` remains responsive. The integration tests can take some time to execute. When you multiply that with the different versions of Bazel, it may be best to execute them when requested. I am happy to make them execute with other tests.  Just let me know.

bazel_versions.bzl
cgrindel: I will remove `4.2.2`. There is not an easy way to specify HEAD/rolling. In other projects, I specify `6.0.0-pre.20220223.1` and periodically update it with whatever is the latest.

buildbuddy.yaml
brentleyjones: No, that makes sense. Thanks!

cgrindel:
With regard to using the examples directly, yes, we can use them for the integration tests. That is what I do in my projects. However, they would become part of the child workspace and would not be in the parent workspace. Hence, you would need to run the integration tests to exercise them. 

I looked at what it would take to copy them over. We would need to write a bunch of code to modify target references (e.g. `//examples/ios_app/Example` to `//Example`). This seems pretty brittle though. The copy code could easily get out of sync with the examples tree.

I would recommend having the example be the integration test. This would prevent the duplicate code that we would be introducing with this PR. The one thing to keep in mind is that we would need to remove the spaces from the paths under the examples tree.

Let me know what you think.

cgrindel:
@brentleyjones When you get a chance, can you take a look at the shape of this PR. I want to be sure that it looks reasonable before I tidy it up. Here are some of the highlights:
- I converted the `examples/ios_app` to a child workspace with two integration tests (Bazel 5.0.0, Bazel 6.0.0-pre-release).
- To make the fixture related rules/macros visible to the child worksapce, I copied `fixtures.bzl` and `updater.template.sh` to `xcodeproj/internal` and then made the rules publicly available in `xcodeproj/defs.bzl`.

Questions:
1. If you are cool with the location of the fixture files, I will create a separate PR to move them to their new location and remove the files at `test/fixtures`. How does that sound?
2. A common pattern with rules repositories is to have all public definitions in a `defs.bzl`. Would you be open to me putting up a PR that moves the implementations under `internal` and having the publicly available items defined in `defs.bzl`? This should address most if not all of the visibility issues from #134.

cgrindel:
It looks like the specs generated by Bazel 5.0.0 vs 6.0.0-pre-release can differ. I created #174 to track enabling Bazel integration tests for other Bazel versions. This PR only runs an integration test for the current Bazel version.

examples/BUILD
```diff
@@ -0,0 +1,51 @@
...
+    workspace_files = integration_test_utils.glob_workspace_files("ios_app") + [
+        "//:local_repository_files",
+    ],
+    workspace_path = "ios_app",
+)
+
+# By default, the integration test targets are tagged as `manual`. This
+# prevents the targets from being found from most target expansion (e.g.
+# `//...`, `:all`). To easily execute a group of integration tests, you may
+# want to define a test suite which includes the desired test targets.
```
brentleyjones, line 41: Do we need this comment?

examples/ios_app/Utils/BUILD
```diff
@@ -1,3 +1,5 @@
+load("@rules_cc//cc:defs.bzl", "objc_library")
```
brentleyjones, line 1: I don't want to depend on `rules_cc` as it's actually bad to use it right now, until it starts to get updated with recent Starlarkification efforts.

third_party/com_github_tuist_xcodeproj/BUILD
```diff
@@ -0,0 +1,7 @@
+# MARK: - Integration Test Related Targets
```
brentleyjones, line 1:
```suggestion
# Integration test related targets
```

Can you apply the same re-formatting to all of these similar comments?

xcodeproj/internal/BUILD
```diff
@@ -3,6 +3,7 @@ load("@bazel_skylib//rules:common_settings.bzl", "string_setting")
 exports_files([
     "installer.template.sh",
     "validator.template.sh",
+    "updater.template.sh",
```
brentleyjones, line 6:
```suggestion
    "installer.template.sh",
    "updater.template.sh",
    "validator.template.sh",
```

xcodeproj/internal/xcodeproj_tests.bzl
```diff
@@ -14,10 +14,21 @@ def _from_fixture(
     if target_under_test == None:
         fail("Need to specify the target under test.")
 
-    target_under_test_label = Label(target_under_test)
-    target_prefix = ""
+    # The target_under_test is converted to a Label. If no workspace is
+    # provided in the target, it evaluates the target relative to the
+    # rules_xcodeproj workspace.  To avoid this, we prefix the specified target
```
brentleyjones, line 19:
```suggestion
    # rules_xcodeproj workspace. To avoid this, we prefix the specified target
```

examples/BUILD
cgrindel: I will remove it.

third_party/com_github_tuist_xcodeproj/BUILD
cgrindel: Done.

xcodeproj/internal/xcodeproj_tests.bzl
cgrindel: Done.

xcodeproj/internal/BUILD
cgrindel: Done.

examples/ios_app/Utils/BUILD
cgrindel: Buildifier added it. I am looking into how we can avoid this being added in the future.

cgrindel: I think that I understand how this appeared now. I run Buildifier in my editor with the flag `-warnings all`. That added the load statement. The Buildifier check/fix that we added in a previous PR currently does not run with all of the warnings. I added #176 to track this. For now, I added the file with the `objc_library` to the exclude list.

cgrindel: I just found a warning level that appears to apply to the load statement (`cc-native`). I will apply it in the PR for #176.

brentleyjones (approved):

PR #149, opened 2022-03-15
Add non-Xcode target files to `extra_files`

brentleyjones:
Also adds a header-only `objc_library` target as an integration test.

### DIFF
--- a/examples/cc/lib/BUILD
+++ b/examples/cc/lib/BUILD
@@ -1,9 +1,19 @@
 cc_library(
-    name = "lib",
+    name = "actual_lib",
     srcs = [
         "lib.c",
         "private.h",
     ],
+)
+
+cc_library(
+    name = "headers_only",
     hdrs = ["lib.h"],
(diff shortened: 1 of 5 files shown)


examples/cc/lib/BUILD
```diff
@@ -1,9 +1,19 @@
...
-    name = "lib",
+    name = "actual_lib",
     srcs = [
         "lib.c",
         "private.h",
     ],
+)
+
+cc_library(
+    name = "headers_only",
```
tylerwilliams, line 10: maybe impl_lib and header_only_lib?

(then you avoid the name collision of "actual = headers_only" below)

tylerwilliams (approved):
LGTM (1 nit)

examples/cc/lib/BUILD
brentleyjones: The alias is on purpose to prove that they don't mess with things. I'll rename these though in a followup PR (auto-merged).

PR #152, opened 2022-03-15
Add `macos_command_line_application` example 

brentleyjones:
Part of https://github.com/buildbuddy-io/rules_xcodeproj/issues/99.

This just shows that we have basic support for `macos_command_line_application`. https://github.com/buildbuddy-io/rules_xcodeproj/issues/99 lists additional things we need to add support for.


examples/command_line/lib/dir with space/lib.h
```diff
@@ -0,0 +1,2 @@
+// function declaration that returns a string
+char *greeting();
```
cgrindel, line 2: Curious. Was there special handling that had to be added to support spaces in the path?

cgrindel (approved):
Sorry that this review took so long. I missed the review request. 😳

PR #156, opened 2022-03-17
Refactor `FilePath` resolution

brentleyjones:
It was getting unwieldy passing around all of the paths needed to resolve `FilePath`s.


tylerwilliams (approved):
this is great

PR #159, opened 2022-03-17
Provide `isSwift` to the generator

brentleyjones:
In a future change we will use this value to conditionalize setting some build settings.


tylerwilliams (approved):
if we change this to an enum later you owe me a 🍺  :P

PR #166, opened 2022-03-17
Remove `TARGET_NAME` override

brentleyjones:
Now that we fully customize `CONFIGURATION_BUILD_DIR`, we no longer need to worry about what `TARGET_NAME` is.


tylerwilliams (approved):
ace

PR #172, opened 2022-03-18
Fix processing of `copts` based `-I` setting

brentleyjones:
Some projects set simulate a `local_includes` via `copts`, which is like `includes`, but is only set for the single target, and doesn't propagate to dependent targets.

We need to special case this, as the `bazel-out/` and `external/` in the the `copts` might not resolve to the correct path from within Xcode. Thus we look for those magic prefixes (ew) and translate them to `FilePath`s.

Since this introduces the concept of local quote headers and includes, I also adjusted `_process_search_paths()` to only propagate search paths for public headers (`hdrs`), and not if the target only uses headers in `srcs`/`non_arc_srcs`. This makes Xcode behave more like sandboxed Bazel.

Finally, I added an external repo example to `//examples/cc`, to help expose these issues better in the future.


brentleyjones:
I'll also add starlark tests for the new `opts` processing soon.

tylerwilliams (approved):

PR #194, opened 2022-03-22
Added flag to child workspace to not create convenience symlinks.

### LINKED ISSUE #181: Ensure that the Bazel symlinks are not created for child workspaces.
This avoids errors when attempting to build/test from the parent workspace.

cgrindel:
Closes #181.

### DIFF
--- a/examples/ios_app/.bazelrc
+++ b/examples/ios_app/.bazelrc
@@ -1,2 +1,5 @@
 # Import Shared settings
 import %workspace%/../../shared.bazelrc
+
+# Remove convenience symlinks 
+build --experimental_convenience_symlinks=clean


brentleyjones (commented):
I'll have to think on this a bit. The downside is that it makes the fixtures no longer work as they rely on the symlinks for external and generated files resolution. Normal projects don't care about that because they use absolute paths. This might have to roll into the changes I described here: https://github.com/buildbuddy-io/rules_xcodeproj/issues/174#issuecomment-1074114828

cgrindel:
That is fine. This was a nice to have in the event that someone was manually building in the child workspace and then ran the integration test. I can close this for now.

PR #199, opened 2022-03-24
Refactor input collection

brentleyjones:
We no longer use an additional aspect to collect input files. This is needed in order to properly classify `extra_files`. Before this change, while trying to implement resource bundle targets, the resources would always be classified as extra files, because of the transitive nature of  how `input_files_apsect` collected the `other` field. Now we are able to leverage the information in `XcodeProjInfo` to determine the correct classification of all input files.

This also now includes intermediate generated files. I personally think this is a regression (though others might disagree), but I don't know of an easy way to exclude them at this time (if we even want to).

Finally, `generated_inputs` and `extra_files` have been moved into the `input_files` `struct`.


xcodeproj/internal/target.bzl
```diff
@@ -262,13 +227,12 @@ def _processed_target(
...
-            file
-            for file in modulemaps.files
-            if not file.is_source
-        ],
-        inputs_info = inputs_info,
+        # generated = generated + [
+        #     file
+        #     for file in modulemaps.files
+        #     if not file.is_source
+        # ],
```
brentleyjones, line 234: I need to address this till. Forgot about it.

It's interesting that the specs haven't changed, which means they are being found elsewhere, but maybe not reliably.

xcodeproj/internal/target.bzl
```diff
@@ -728,7 +702,102 @@ def _process_library_target(*, ctx, target, transitive_infos):
...
+            links = [],
+            dependencies = dependencies,
+            outputs = _process_outputs(target),
+        ),
+    )
+
+# Resource targets
+
+def _process_resource_target(*, ctx, target, transitive_infos):
+    """Gathers information about a resource target.
```
brentleyjones, line 715: This should be in the next PR. Removing.

tylerwilliams (approved):
comment _merge?

PR #219, opened 2022-03-29
Improve diff on `createdFiles`

brentleyjones:
We shouldn't need to create the intermediate struct and sort, but currently `CustomDump` mixes up values and keys, making the diff output unreadable. This is at least parsable now when there is a diff failure here.


tylerwilliams (approved):
every good project needs a std::pair type

Issue #99, opened 2022-03-08
Support `macos_command_line_application`


brentleyjones:
https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-macos.md#macos_command_line_application

- [x] #152
- [x] Support `infoplists`
  - [x] For targets with AppleBundleInfo: #225
  - [x] For targets with ObjcProvider: #234 
- [x] Support `provisioning_profile`
- [x] Support `codesignopts` #593
- [x] Support `exported_symbols_lists`: #649 
- [ ] Support `launchdplists`: #684

maxwellE:
Sounds good @brentleyjones

luispadron:
@brentleyjones we can mark the last task here completed as well.

PR #234, opened 2022-04-01
Add support for retrieving the `Info.plist` from an `ObjcProvider`

cgrindel:
Related to #99.

- Updated the `command_line` example to include an `Info.plist`.
- Added `link_opts.bzl` API to parse linker options and retrieve information from the options.
- Added `info_plists.bzl` API to encapsulate the logic of retrieving an `Info.plist` from a target.
- Updated `target.bzl` to use `info_plists.bzl` to retrieve an `Info.plist` if present.


test/internal/info_plists/BUILD.bazel
```diff
@@ -0,0 +1,6 @@
+load(":get_file_from_objc_provider_tests.bzl", "get_file_from_objc_provider_test_suite")
+load(":get_file_tests.bzl", "get_file_test_suite")
+
+get_file_from_objc_provider_test_suite()
```
brentleyjones, line 4: Can we pass the name of the suites here like in the other tests?

cgrindel: Sure. Curious. What is the benefit of doing that way?

brentleyjones: Mainly following this: https://github.com/bazelbuild/rules_apple/commit/e3ebb8f4b7b8ed8b155b49d434bc7e13e41ba092

cgrindel: Interesting. I am surprised that Buildifier did not yell at me. Thanks.

brentleyjones (approved):
I really like the encapsulation here. I need to do that more elsewhere in the code.

FYI, I applied some small formatting changes to the branch instead of leaving them as suggested changes here.

Issue #174, opened 2022-03-19
Enable integration test with Bazel 6.0.0-pre.XXXX


cgrindel:
The specs for the `ios_app` example differ when generated using Bazel 5.0.0 and 6.0.0-pre.20220223.1. See below for an example. To enable testing with a different version of Bazel, we can do one of the following:

1. Only perform the spec test on the "current" version of Bazel. The other Bazel versions will skip the spec test.
2. Perform the spec test against the checked-in specs using the current Bazel version. For the other Bazel versions, we would execute the fixture update command before running the spec test. The benefit of this approach is that it exercises the spec testing code on different Bazel versions.
3. Transition away from spec tests and write tests that assert the existence of certain elements in the output.

## Example Diff Comparing Spec Generated using Bazel 5.0.0 and Bazel 6.0.0-pre.20220223.1

```
==================== Test output for //test/fixtures:xcodeproj_fixtures:
Spec doesn't match expected:
39c39
<         "//Example:Example.library ios-x86_64-min15.0-applebin_ios-ios_x86_64-fastbuild-ST-4c6d2b462eb0",
---
>         "//Example:Example.library ios-x86_64-min15.0-applebin_ios-ios_x86_64-fastbuild-ST-d3e756bfe7fd",
41c41
<             "//Example:Example applebin_ios-ios_x86_64-fastbuild-ST-4c6d2b462eb0"
---
>             "//Example:Example applebin_ios-ios_x86_64-fastbuild-ST-d3e756bfe7fd"
43c43
<         "//ExampleTests:ExampleTests.library ios-x86_64-min15.0-applebin_ios-ios_x86_64-fastbuild-ST-4c6d2b462eb0",
---
>         "//ExampleTests:ExampleTests.library ios-x86_64-min15.0-applebin_ios-ios_x86_64-fastbuild-ST-d3e756bfe7fd",
45c45
<             "//ExampleTests:ExampleTests.__internal__.__test_bundle applebin_ios-ios_x86_64-fastbuild-ST-4c6d2b462eb0"
---
>             "//ExampleTests:ExampleTests.__internal__.__test_bundle applebin_ios-ios_x86_64-fastbuild-ST-d3e756bfe7fd"
47c47
<         "//ExampleUITests:ExampleUITests.library ios-x86_64-min15.0-applebin_ios-ios_x86_64-fastbuild-ST-4c6d2b462eb0",
---
[...]

PR #246, opened 2022-04-05
Fix compiling with Xcode 13.3 on macOS 12.2

brentleyjones:
Xcode 13.3 sets the SDK version to 12.3, which results in a runtime error if the host macOS isn't 12.3. We resolve this by using `--{host_,}macos_minimum_os`, and setting our minimum Bazel version to 5.1.0.


test/fixtures/cc/project.xcodeproj/project.pbxproj
```diff
@@ -341,7 +341,7 @@
 		74237F30610FC68FC69D29AF /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				BAZEL_PACKAGE_BIN_DIR = "bazel-out/darwin_x86_64-fastbuild-ST-1b9bd654f600/bin/examples/cc/lib2";
+				BAZEL_PACKAGE_BIN_DIR = "bazel-out/darwin_x86_64-fastbuild-ST-d53d69b6b8c1/bin/examples/cc/lib2";
```
brentleyjones, line 344: #174

PR #253, opened 2022-04-06
Fix `PRODUCT_MODULE_NAME` calculation for external repos

brentleyjones:
We needed to use `swift_common.derive_module_name()`. We now only set a product module name for Swift targets, since the module name for Objective-C targets comes from the modulemap, which is supplied to us from Bazel.


erikkerber (approved):
🙇🏼

PR #257, opened 2022-04-06
Update invalid merge logging removing the force unwrap expressions

cgrindel:
Fixes an illegal instruction failure.

### DIFF
--- a/tools/generator/src/Generator.swift
+++ b/tools/generator/src/Generator.swift
@@ -61,14 +61,14 @@ class Generator {
             guard let srcTarget = targets[invalidMerge.source] else {
                 throw PreconditionError(
                     message: """
-                    Expected the invalid merge source to exist.  \(String(reflecting: invalidMerge))
-                    """)
+Source target "\(invalidMerge.source)" not found in `targets`
+""")
             }
             for destination in invalidMerge.destinations {
                 guard let desTarget = targets[destination] else {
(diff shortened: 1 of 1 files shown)


tools/generator/src/Generator.swift
```diff
@@ -58,12 +58,23 @@ class Generator {
         )
 
         for invalidMerge in invalidMerges {
+            guard let srcTarget = targets[invalidMerge.source] else {
+                throw PreconditionError(
+                    message: """
+                    Expected the invalid merge source to exist.  \(String(reflecting: invalidMerge))
+                    """)
```
brentleyjones, line 65:
```suggestion
                    message: """
Source target "\(invalidMerge.source)" not found in `targets`
""")
```

tools/generator/src/Generator.swift
```diff
@@ -58,12 +58,23 @@ class Generator {
...
+            }
             for destination in invalidMerge.destinations {
+                guard let desTarget = targets[destination] else {
+                    throw PreconditionError(message: """
+                    Expected the destination target to exist. \(String(reflecting: destination))
+                    """)
```
brentleyjones, line 71:
```suggestion
                    throw PreconditionError(message: """
Destination target "\(destination)" not found in `targets`
""")
```

brentleyjones (approved):

Issue #228, opened 2022-04-01
Bug: Custom `AppleResourceInfo` aren't respected


brentleyjones:
## Describe the bug

If a custom rule provides an `AppleResourceInfo`, those resources aren't picked up and bundled correctly.

brentleyjones:
I already have a fix in progress. I just wanted to report the bug in case anyone runs into this issue.

Issue #80, opened 2022-03-08
Fully support `apple_resource_bundle`


brentleyjones:
- [x] https://github.com/buildbuddy-io/rules_xcodeproj/pull/208
- [x] Support `infoplists`: #225

brentleyjones:
This is done at the base level. A known bug is tracked in #228.

PR #263, opened 2022-04-08
Fix resource targets that are included via `deps`

brentleyjones:
Part of #80, #81, #103. 

This doesn't make sense, but it's clearly supported with rules_apple. This also is more fuel for me to eventually rewrite this in a way that fixes another resource completeness bug (#228).


erikkerber:
We'll update our builds to use `data` for resources as well

Issue #95, opened 2022-03-08
Support `{ios,macos,tvos,watchos}_unit_test`


brentleyjones:
- [x] `ios_unit_test`
- [x] `macos_unit_test` (#267)
- [x] `tvos_unit_test` (#282)
- [x] `watchos_unit_test`

erikkerber:
`macos_unit_test` not supported as of 0.1.0

brentleyjones:
I've updated the comments.

cgrindel:
At this point, I believe that we have rudimentary support for the listed unit test rules. Closing this issue.

Issue #136, opened 2022-03-11
Split up `target.bzl`


brentleyjones:
https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/xcodeproj/internal/target.bzl has grown to be quite large. While everything in there doesn't really belong in any of the other `.bzl` files, it can still be split up a bit more.

maxwellE:
@brentleyjones Is this ticket complete or should we continue to move out logic from target.bzl?

PR #267, opened 2022-04-08
Add `macos_unit_test` support

cgrindel:
Related to #95.

- Added `targets.bzl` to encapsulate API for inspecting and acting upon a target. The hope is to move some additional functionality from `target.bzl` to this API.
- Added check for MacOS test bundle target when deciding whether to skip a target.


brentleyjones:
Can you pull out the `target.bzl` refactor (without changes needed for this PR) to it's own PR, an associate it with #136?

cgrindel:
@brentleyjones I removed the refactor and added the simplest implementation.

brentleyjones:
So it's clear: I do like the refactor, I just feel we should do those in their own PRs 😄.

brentleyjones (approved):

PR #274, opened 2022-04-09
Fix function name typo in warning message

a2:
I saw this warning locally while working on my patch in [bazelbuild/rules_apple](https://github.com/bazelbuild/rules_apple). Basically the warning calls the function `rules_xcodeproj_dependencies` whereas the function's name is `xcodeproj_rules_dependencies`.


brentleyjones:
Thanks!

It's the function that has the wrong name though 😅

a2:
Oh no! Okay, is there a concern that fixing the function's name could break in clients' code?

brentleyjones:
Not really, as we are pre-1.0, so there will be breaking changes. We will call it out in the release notes.

brentleyjones (approved):
So, it seems that the other repos follow this pattern as well. Let's keep it named this way. I'll accept this change as is.

Thanks again!

PR #276, opened 2022-04-11
Use `CcInfo` for `defines` and `local_defines`

brentleyjones:
This also fixes some parsing issue with escaped `"`s. And an oddity(?) where `swift_library` sets defines for cc dependents.


brentleyjones:
(I really need to get the `xcodebuild` based integration tests up and running.)

PR #286, opened 2022-04-12
Add `watchos_app` example

cgrindel:
Related to #95.

- Added example that includes an iOS app with a corresponding watchOS app.
- Updated `targets.is_test_bundle` to support `WatchosXcTestBundleInfo`.

### DIFF
--- a/examples/watchos_app/ExampleWatchKitExtension/BUILD
+++ b/examples/watchos_app/ExampleWatchKitExtension/BUILD
@@ -1,0 +1,52 @@
+load("@build_bazel_rules_apple//apple:resources.bzl", "apple_resource_group")
+load("@build_bazel_rules_apple//apple:versioning.bzl", "apple_bundle_version")
+load("@build_bazel_rules_apple//apple:watchos.bzl", "watchos_extension")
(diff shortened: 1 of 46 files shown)


examples/watchos_app/ExampleWatchKitExtension/BUILD
```diff
@@ -0,0 +1,50 @@
...
+    values = {
+        "compilation_mode": "opt",
+    },
+)
+
+apple_bundle_version(
+    name = "Version",
+    build_version = "1.0",
+    visibility = ["//examples/watchos_app:__subpackages__"],
+)
```
brentleyjones, line 17: Seems like we support https://github.com/buildbuddy-io/rules_xcodeproj/issues/74 already? There is this comment https://github.com/buildbuddy-io/rules_xcodeproj/blob/db7363a95346c39d33beb26972ca0b97a6040e17/xcodeproj/internal/xcodeproj.bzl#L55 related to `CURRENT_PROJECT_VERSION` and `MARKETING_VERSION`. Maybe we should removed those now that we always set the Info.plist which has those built in?

Not in this PR though, just made me think of it.

brentleyjones (approved):

PR #298, opened 2022-04-15
Pre-generate files during initial project creation

brentleyjones:
To make initial experiences nicer, we now run the `Bazel Generated Files` scheme as part of initial project creation. This won't be run each time the project is generated, only if the `gen_dir` isn't found.

To disable this behavior, you can set `pre_generate_files = False` on your `xcodeproj` target. In that case, it will only run the `Setup` scheme, which will ensure that Xcode behaves nicely once the generated files are created, without having to close and re-open the project.

Longer term we might want to reduce the number of files pre-generated, as this step might take a long time. Also, Build with Bazel might have different requirements.

### DIFF
--- a/xcodeproj/internal/fixtures.bzl
+++ b/xcodeproj/internal/fixtures.bzl
@@ -76,11 +76,6 @@ _update_fixtures = rule(
         "_allowlist_function_transition": attr.label(
             default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
         ),
-        "_installer_template": attr.label(
-            allow_single_file = True,
-            executable = False,
-            default = Label("//xcodeproj/internal:installer.template.sh"),
-        ),
         "_updater_template": attr.label(
             allow_single_file = True,
             default = ":updater.template.sh",
--- a/xcodeproj/internal/installer.template.sh
+++ b/xcodeproj/internal/installer.template.sh
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 readonly src="$PWD/%source_path%"
-readonly dest="${BUILD_WORKSPACE_DIRECTORY}/%output_path%"
+readonly dest="$BUILD_WORKSPACE_DIRECTORY/%output_path%"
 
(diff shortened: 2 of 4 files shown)


xcodeproj/internal/fixtures.bzl
```diff
@@ -117,6 +112,7 @@ def xcodeproj_fixture(
         name = name,
         external_dir_override = "bazel-{}/external".format(workspace_name),
         generated_dir_override = "bazel-out",
+        # pre_generate_files = False,
```
BalestraPatrick, line 115: Is this needed?

brentleyjones: Oops, no.

xcodeproj/internal/installer.template.sh
```diff
@@ -3,7 +3,7 @@
 set -euo pipefail
 
 readonly src="$PWD/%source_path%"
-readonly dest="${BUILD_WORKSPACE_DIRECTORY}/%output_path%"
+readonly dest="$BUILD_WORKSPACE_DIRECTORY/%output_path%"
```
cgrindel, line 6: I believe that the curly brace word expansion is the preferred form. Was there a reason to remove it (e.g. template expansion)?

cgrindel (approved):

xcodeproj/internal/installer.template.sh
brentleyjones: It's only needed when you need to disambiguate the name, or do something special like substrings or pattern matching, right?

Here is an example of @keith requesting a similar change in rules_apple: https://github.com/bazelbuild/rules_apple/pull/1417#discussion_r850955300

cgrindel: I agree that in this case, it is not necessary. It is more of a style issue, I suppose.

keith: Yea I guess we don't really have a specific style for this. My informal style guide for this internally, which I am projecting, is to only use these when it's necessary.

PR #301, opened 2022-04-15
Include `BUILD` files in Xcode's Project navigator

jpsim:
Resolves https://github.com/buildbuddy-io/rules_xcodeproj/issues/300

|Before|After|
|-|-|
|image|image|

### DIFF
--- a/test/fixtures/command_line/spec.json
+++ b/test/fixtures/command_line/spec.json
@@ -27,7 +27,6 @@
             "_": "macos-x86_64-min11.0-applebin_macos-darwin_x86_64-fastbuild-ST-8946c8252059/bin/examples/command_line/lib/lib_swift.swift.modulemap",
             "t": "g"
         },
-        "apple/testing/BUILD",
         {
             "_": "build_bazel_rules_apple/apple/testing/DefaultTestBundle.plist",
             "t": "e"
@@ -35,8 +34,7 @@
         {
             "_": "applebin_macos-darwin_x86_64-fastbuild-ST-8946c8252059/bin/examples/command_line/Tests/LibSwiftTests.__internal__.__test_bundle-intermediates/Info.plist",
             "t": "g"
-        },
-        "apple/testing/default_runner/BUILD"
+        }
     ],
     "invalid_target_merges": [],
     "label": "//test/fixtures/command_line:xcodeproj",
--- a/test/fixtures/command_line/spec.json
+++ b/test/fixtures/command_line/spec.json
@@ -10,20 +10,24 @@
         "VALIDATE_WORKSPACE": false
     },
     "extra_files": [
+        "examples/command_line/lib/BUILD",
         {
             "_": "macos-x86_64-min12.0-applebin_macos-darwin_x86_64-fastbuild-ST-f2aadcfcab57/bin/examples/command_line/lib/lib_impl.swift.modulemap",
             "t": "g"
         },
         "examples/command_line/lib/dir with space/lib.h",
+        "examples/command_line/tool/BUILD",
         "examples/command_line/tool/Info.plist",
         {
             "_": "macos-x86_64-min11.0-applebin_macos-darwin_x86_64-fastbuild-ST-8946c8252059/bin/examples/command_line/lib/lib_impl.swift.modulemap",
             "t": "g"
         },
+        "examples/command_line/Tests/BUILD",
         {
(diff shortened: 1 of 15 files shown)


xcodeproj/internal/input_files.bzl
```diff
@@ -32,9 +32,24 @@ def _collect_transitive_extra_files(info):
     inputs = info.inputs
     transitive = [inputs.extra_files]
     if not info.target:
-        transitive.append(inputs.srcs)
-        transitive.append(inputs.non_arc_srcs)
-        transitive.append(inputs.hdrs)
+        new_srcs = []
```
jpsim, line 35: This is super verbose, I'm sure there's a nicer way to do this.

test/fixtures/command_line/spec.json
```diff
@@ -10,28 +10,33 @@
...
         {
             "_": "macos-x86_64-min11.0-applebin_macos-darwin_x86_64-fastbuild-ST-8946c8252059/bin/examples/command_line/lib/lib_impl.swift.modulemap",
             "t": "g"
         },
+        "examples/command_line/Tests/BUILD",
         {
             "_": "macos-x86_64-min11.0-applebin_macos-darwin_x86_64-fastbuild-ST-8946c8252059/bin/examples/command_line/lib/lib_swift.swift.modulemap",
             "t": "g"
         },
+        "apple/testing/BUILD",
```
jpsim, line 30: Not sure what the best way is to avoid this from being included.

tools/generator/src/Path+Extensions.swift
```diff
@@ -12,6 +12,12 @@ extension Path {
 
     var isLocalizedContainer: Bool { self.extension == "lproj" }
 
+    var isBazelBuildFile: Bool { self.lastComponent == "BUILD" && self.extension == nil }
```
keith, line 15: Probably want to allow BUILD.bazel too which some folks use

xcodeproj/internal/input_files.bzl
brentleyjones: Hmm, I guess this is why I just had them as `File`.

I wonder if it would be better to have a second `extra_file_paths` instead of changing the type of `extra_files`? Then in `xcodeproj.bzl` we append `extra_file_paths` to the transformed `extra_files` and call `to_dto` on that?

test/fixtures/command_line/spec.json
brentleyjones: Not a good way right now. We can open an issue to clean it up later, which I'll do when I refactor the input collection as part of fixing #228.

xcodeproj/internal/input_files.bzl
jpsim: I don't feel strongly either way. I have a slight preference for keeping it as-is because I think this mapping is still simpler than having two ways to add extra files.

I was more asking if there's a more concise way to perform this mapping of a depset. Like `inputs.srcs.map(file_path)`.

test/fixtures/command_line/spec.json
brentleyjones: Actually, we can just only include BUILD files for the project instead of the external repos:
```
if not target.label.workspace_root:
    # Add BUILD file here
```

jpsim: 7aebba5

xcodeproj/internal/input_files.bzl
brentleyjones: The downside to this way is that we are cracking open the depset for each target, and if there are files that are duplicated in the inputs it's more efficient to only it once later.

I'm fine with it like this though, since I'll be refactoring stuff later. Better to keep it localized for now.

As for your question:
```
transitive.extend([
    file_path(file)
    for file in inputs.srcs.to_list()
])
```

jpsim: Ah I didn't notice this syntax elsewhere, it's surprising to me because it's wrapped in square brackets already. [e53f82f](https://github.com/buildbuddy-io/rules_xcodeproj/pull/301/commits/e53f82ffd9d10de58a5090a9e5b0e339787ee786)

brentleyjones (approved):

brentleyjones (approved):

brentleyjones (approved):

brentleyjones (approved):

brentleyjones:
(Trying to debug a CI issue, sorry for the noise.)

jpsim:
Thanks for guiding me through this @brentleyjones! 🙌

brentleyjones:
Thanks for the contribution!

PR #310, opened 2022-04-16
Support `includes` in cc_library

jpsim:
By specifying `SYSTEM_HEADER_SEARCH_PATHS` in Xcode build settings.

Partially addresses https://github.com/buildbuddy-io/rules_xcodeproj/issues/82.

Allows [Yams](https://github.com/jpsim/Yams) to compile.

### DIFF
--- a/test/internal/opts/process_compiler_opts_tests.bzl
+++ b/test/internal/opts/process_compiler_opts_tests.bzl
@@ -72,7 +72,7 @@ def process_compiler_opts_test_suite(name):
             *,
             name,
             expected_build_settings,
-            expected_search_paths = {"quote_includes": [], "includes": []},
+            expected_search_paths = {"quote_includes": [], "includes": [], "system_includes": []},
             conlyopts = [],
             cxxopts = [],
             full_swiftcopts = [],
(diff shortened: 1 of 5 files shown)


brentleyjones:
This _should_ be handled by the other mechanisms that collect header include paths. Ideally we don't look at `ctx.attr` if possible, to allow custom rules to "just work".

Can you debug a bit to see where the disconnect is? We look at `CcInfo.compilation_context`, which should have these.

brentleyjones:
Looks like `cc_library` sets these in [`CcInfo.compilation_context.system_includes`](https://docs.bazel.build/versions/main/skylark/lib/CompilationContext.html#system_includes). I thought that was going to be for SDK paths, but thankfully it's not.

So, the more complete fix is to parse that in `_process_search_paths()` and set it to `SYSTEM_HEADER_SEARCH_PATHS` in`generator`. Also, we should parse `-isystem` in `opts.bzl` similar to `-iquote` and have `opts_search_paths` have a `system_includes`.

jpsim:
Thanks for the guidance (again!). I applied what I think you asked for, and confirmed that Yams is still compiling in Xcode with the changes. Let me know if I misinterpreted something.

brentleyjones (approved):
Perfect! Thanks!

Issue #311, opened 2022-04-16
Feature Request: Support external Swift packages in "Build with Xcode" projects.


cgrindel:
## Problem

External Swift packages are, typically, represented in Bazel projects in one of two ways:

1. As an external repositry using `http_archive` providing a custom build file.
2. As an external repository using `spm_repositories` and `spm_pkg` from
   [cgrindel/rules_spm](https://github.com/cgrindel/rules_spm/). 

If an Xcode project is generated to build with Bazel (BWB), no additional support is required in the
Xcode project for the project to build properly. Whatever is defined in the Bazel project will
handle the building and linking of the external dependencies.

If an Xcode project is generated to build with Xcode (BWX), the Xcode project needs to understand
how to incorporate the external dependency.

PR #312, opened 2022-04-16
Design document for adding external Swift packages to "Build with Xcode" projects

cgrindel:
Related to #311.

The document in this PR is meant to start a discussion about how to handle external Swift packages. If folks prefer to conduct the discussion in Google Docs, I would be glad to move the document.


design_docs/support_external_swift_packages.md
```diff
@@ -0,0 +1,204 @@
...
+
+- The generation code can leverage much of its existing logic mapping Bazel targets.
+- This maps well for Bazel projects that download the external package using `http_archive` and
+  define a custom build file.
+
+The cons for this approach are:
+
+- The generated project will be cluttered with targets that are not relevant to the developer.
+  These will include not just the direct dependencies, but all of the transitive dependencies for
+  the external packages.
```
BalestraPatrick, line 34: I assume we might have a different way to filter out targets that aren't relevant to a particular developer for their project in the future, so this would be a problem only until we have a solution for it that could be applicable for this as well.

design_docs/support_external_swift_packages.md
```diff
@@ -0,0 +1,204 @@
...
+- The generation code will need to handle the source for these external dependencies, differently.
+  It will need to download the source for the external dependencies, map it and make it available in
+  the user interface.
```
BalestraPatrick, line 37: How is this different from today?

design_docs/support_external_swift_packages.md
```diff
@@ -0,0 +1,204 @@
...
+    ],
+)
+
+xcodeproj(
+    name = "xcodeproj",
+    project_name = "My App",
+    tags = ["manual"],
+    targets = XCODEPROJ_TARGETS,
+    external_deps = [
+        ":external_packages",
```
BalestraPatrick, line 141: So, from my understanding today we handle external packages with option 1 already. If we were to go with option 2, this would significantly increase the amount of work to adopt this rule for projects made up of only external `http_archive`s. Today, someone can simply drop a `xcodeproj` and the project should build just fine in Xcode. Instead, with this option the amount of rules required would be higher and we're basically duplicating the dependencies definitions once for building, and once for the IDE integration. Is my interpretation correct?

tinder-maxwellelliott:
I really enjoyed reading this proposal and it gave me a better sense of how the rules work. Thanks!

design_docs/support_external_swift_packages.md
cgrindel: Perhaps. If the files are used for compilation, they need to exist in the project somewhere. Is there a way in Xcode to hide targets, but still use them for compilation.

design_docs/support_external_swift_packages.md
cgrindel: The xcodeproj rules currently create targets for source files that exist in the target project (i.e., where you are running the generator code). The source code for the external dependencies does not live in the target project. So, the generator code will need to retrieve the source probably via `bazel fetch` and map the files. Perhaps, this will not be a heavy lift. However, I am sure that there will be some nuances. Also, I am not sure that most developers will want to have it included in the project's navigator pane.

cgrindel: I just noticed that we do have an example that uses an external lib (`//examples/cc/tool`). So, perhaps, there aren't as many nuances as I had anticipated.

design_docs/support_external_swift_packages.md
cgrindel: In retrospect, I should not have proposed this as an either/or situation. In my mind, option 1 and option 2 offer different user experiences. (I wonder if there may be an Xcode performance benefit by reducing the number of targets in the project.) If I reframe the proposal as an alternative means to import external dependencies, I think that the extra mapping may be desirable for some projects, especially those with a large number of external transitive dependencies. 

In terms of the work adding the mapping for the direct dependencies for a project, it will vary. In general, the number of direct dependencies for a project will be lower than the number of transitive dependencies. In some cases, it will be much lower. For instance, there is [a Vapor application example in the rules_spm project](https://github.com/cgrindel/rules_spm/blob/main/examples/vapor/WORKSPACE#L43) that has 3 direct dependencies. When fully resolved, it results in 24 transitive dependencies. When writing this proposal, I was thinking about whether I would want to work with an Xcode project that has 5 targets for my project or 29 targets (5 source targets + 24 dependency targets). So, for this example, one would need to map 3 direct dependencies (i.e., ~25 lines of Starlark code) that would result in a project with 5 targets all of which are targets from the project. None of them would be for the external dependencies.

Back to the reframing, what if I update the design to reflect the following scenarios?

### Scenario 1: Initial Usage of `rules_xcodeproj`
- Add `xcodeproj` rule to your project and go.
- Everything works
- All of the transitive dependencies appear in Xcode.

### Scenario 2: Hide External Swift Packages
- Add `xcodeproj_external_swift_packages` mapping for the direct dependencies.
- Only the source targets appear in the resulting Xcode project.

design_docs/support_external_swift_packages.md
brentleyjones: Yes, but it involves a more sophisticated Bazel Generated Files step (or additional similar targets). I'll detail something like that soon, as it's needed for other use cases anyway.

brentleyjones: But also in Build with Bazel, we won't need most files to be part of the project.

design_docs/support_external_swift_packages.md
BalestraPatrick: To me it sounds like this design doc should be focused on how to offer a native experience for projects using SPM through `rules_spm`. The fact that external Swift package are "hidden" is just a side-effect of using Xcode's native SPM functionality. If that's something that a project would like to offer as a usability or performance improvement, they should be able to do that for sure. For example, it could be another attribute named `uses_native_spm` in order to generate an Xcode project that uses native SPM when building with Xcode. The attribute would default to `False`, but setting that to `True` would open up the possibility of defining the native SPM to be added via the mapping of the dependencies as presented earlier. Does that make sense?

brentleyjones: I think it can go a little further: we can add SPM related values to `XcodeProjInfo`, but the default rules won't fill them in. This means we will default to Scenario 1. Then `rules_spm` (or any other custom rules) can supply `XcodeProjInfo` providers with the SPM values filled in. `rules_spm` could maybe have the attribute to allow users to choose if they would want SPM represented natively or not (though maybe it just always does it, since using `rules_spm` is already stating that it wants to "natively" integrate with SPM).

brentleyjones:
I'm going to respond to this soon, I've just been trying to push out some high priority fixes. A mini response is here though: https://github.com/buildbuddy-io/rules_xcodeproj/pull/312#discussion_r852214151.

brentleyjones:
Quick meta comment though: can this live under `docs/design/` instead of a new top level directory?

brentleyjones:
The doc isn't finished yet, but [this section](https://github.com/buildbuddy-io/rules_xcodeproj/blob/dc0c1c60cd56f7d4141efe6312b66807142367f7/doc/design/high-level.md#projects-can-be-customized) of the "High Level Design" goes into more details on how I feel we should allow customization of the generated project. In particular, I think `rules_spm` can provide one of the providers that `rules_xcodeproj` expects. Adjusting this design document to describe how `XcodeProjInfo` and the spec.json/DTO needs to change to support that customization would be the route I would go.

cgrindel:
The part that I think is not currently handled in the high-level design is how to add Xcode project elements that `rules_xcodeproj` does not know how to add. Specifically, Xcode adds external Swift packages as `XCSwiftPackageProductDependency` and then adds a reference to that entity in `packageProductDependencies` for `PBXNativeTarget`. Right now, `rules_xcodeproj` does not know anything about this type of entity. 

At a high-level, the third-party would need to provide information about the items that need to be added to the Xcode project (i.e., serialized version of [tuist models](https://github.com/tuist/XcodeProj)) and how to wire them up (e.g. the UUID of this thing I told you about is a dependency for this other target).

Either way, I will close this PR, for now. As things progress, I will open another PR with a design for adding things like SPM dependencies.

brentleyjones:
> The part that I think is not currently handled in the high-level design is how to add Xcode project elements that `rules_xcodeproj` does not know how to add. Specifically, Xcode adds external Swift packages as `XCSwiftPackageProductDependency` and then adds a reference to that entity in `packageProductDependencies` for `PBXNativeTarget`. Right now, `rules_xcodeproj` does not know anything about this type of entity.

Fields would be added to `XcodeProjInfo` such that they would make their way to `generator` through the `spec.json`/`DTO.swift`. Then there it would translate those to `XCSwiftPackageProductDependency` etc.

Third party rules would depend on `rules_xcodeproj` (or more likely a `rules_xcodeproj_support`) to create and return the `XcodeProjInfo` provider. We would also adjust our logic to not create a `XcodeProjInfo` if one already returned for a target.

PR #316, opened 2022-04-18
Add `bazel` option to `build_mode` and force `dbg` for it

brentleyjones:
Right now the only difference is that in `bazel` mode the `compilation_mode` is currently forced to be `dbg`. This "restriction" will be lifted when we implement proper configuration support. For now though, we need this compilation mode to enable debugging in Build with Bazel.

We now also generate another set of fixtures for Build with Bazel, a.k.a "bwb".

### DIFF
--- a/xcodeproj/internal/xcodeproj.bzl
+++ b/xcodeproj/internal/xcodeproj.bzl
@@ -188,6 +188,29 @@ def _write_installer(
 
     return installer
 
+# Transition
+
+def _xcodeproj_transition_impl(settings, attr):
+    """Transition that applies command-line settings for the xcodeproj rule."""
+    if attr.build_mode == "bazel":
+        compilation_mode = "dbg"
+    else:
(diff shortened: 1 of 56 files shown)


xcodeproj/internal/xcodeproj.bzl
```diff
@@ -267,6 +290,9 @@ def make_xcodeproj_rule(*, transition = None):
             allow_empty = False,
             aspects = [xcodeproj_aspect],
         ),
+        "_allowlist_function_transition": attr.label(
+            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+        ),
```
cgrindel, line 295: I don't see this being used anywhere. Is this needed in this PR?

brentleyjones: Yes. It enables the use of the transition that is set on `cfg`.

BalestraPatrick (approved):

PR #319, opened 2022-04-18
Determine `bazel-out` and `external` paths at build time

### LINKED ISSUE #201: Bug: Project with generated or external files fails to build when workspace is moved
## Describe the bug

When you move the workspace that contains a generated project, and it uses Bazel Generated Files, the project can fail to compile if the project isn't regenerated. This is because the project has hardcoded references to the `--output_base`, and moving the workspace changes the default location for that.

## Expected behavior

One of three things can be done:

- A Run Script target can fail saying that the project needs to be regenerated
    - We can hard code the workspace path into the script and check against it, or
    - We can run `bazel info` and check if the paths are different than what were calculated
- A Run Script target can create a symlink in derived data for `bazel-out/` and `external/` and file references can use that instead; this gets around the same issue as the hardcoded paths (it's resilient to the ephemeral nature of `bazel-WORKSPACE` symlinks)

All of these can be disabled if both `external_dir` and `generated_dir` are set (like in fixtures). We also need to account for custom `--output_base`.

brentleyjones:
Fixed #201.

This removes the need for test fixtures to be slightly different from normal projects. It will also be needed for future Indexing improvements, which will make the symlinks in the workspace unreliable. Finally, this fixes the issue builds breaking after relocating the workspace.


examples/ios_app/test/fixtures/bwb.xcodeproj/project.pbxproj
```diff
@@ -1796,7 +1801,7 @@
 				ENABLE_TESTING_SEARCH_PATHS = YES;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(PROJECT_DIR)/third_party",
-					"$(PROJECT_DIR)/bazel-ios_app/external/examples_ios_app_external",
+					"$(BAZEL_EXTERNAL)/examples_ios_app_external",
```
brentleyjones, line 1804: Besides for being shorter, I like that these are more readable, especially in diffs.

cgrindel (approved):

PR #330, opened 2022-04-19
Use a different `--output_base` for Index Build

brentleyjones:


tools/generator/src/Generator+AddTargets.swift
```diff
@@ -156,26 +156,36 @@ Product for target "\(id)" not found in `products`
...
 set -eu
 
+if [ "$ACTION" == "indexbuild" ]; then
+  output_base="$OBJROOT/bazel_output_base"
+fi
+
 output_path=$(\#(bazelExec) \
+  ${output_base:+--output_base "$output_base"} \
   info \
+  ${output_base:+--experimental_convenience_symlinks=ignore} \
```
brentleyjones, line 166: Since we don't use the convenience links in either version of this build, I could do this unconditionally?

BalestraPatrick (approved):

cgrindel (approved):

PR #333, opened 2022-04-19
Use an "output map" for the `generated_inputs` output group

brentleyjones:
With this change a file that lists the locations of all of the generated input files is included in the `generated_inputs` output group instead of the files themselves. This results in two improvements:

1. The bazel output is cleaner. With lots of generated inputs the output from `bazel build` could be very noisy.
2. The more important thing: only the output map is listed in the BEP as an "important output". This keeps the size of the BEP small (sometimes on the order of hundreds or more megabytes smaller).

Since the files themselves are no longer listed as top-level outputs, we have to do some "trickery" to ensure they are still downloaded when using `--remote_download_toplevel`. We write a unique value to a file inside of project's internal directory, and use that file as an input to the output map generation. This, along with specific execution infos and input declarations, cause the files to be downloaded to the host as part of writing the output map.


xcodeproj/internal/output_group_map.bzl
```diff
@@ -0,0 +1,161 @@
+"""Module containing functions dealing with output group map files."""
+
+# Each "file" might actually represent many files (if the "file" is an output
+# directory). This value was optimized around indexstores. The larger this value
```
BalestraPatrick, line 4: What exactly is the connection with indexstores here (or is there none and it’s just from your past experience)?

xcodeproj/internal/xcodeproj.bzl
```diff
@@ -280,7 +280,21 @@ _xcodeproj = make_xcodeproj_rule()
 def xcodeproj(*, xcodeproj_rule = _xcodeproj, **kwargs):
     testonly = kwargs.pop("testonly", True)
 
+    project = kwargs.get("project_name", kwargs.get("name"))
+
+    if kwargs.get("toplevel_cache_buster"):
+        fail("`toplevel_cache_buster` is for internal use only")
```
BalestraPatrick, line 286: Should this attribute be prefixed with an underscore to highlight it’s internal/private scope?

xcodeproj/internal/output_group_map.bzl
brentleyjones: Past experience. This code can be used for fetching indexstores in the future though, in Build with Bazel mode, though it isn't right now.

xcodeproj/internal/xcodeproj.bzl
brentleyjones: Underscores for attributes mean they are implicit, and I can't pass in a value. I think this error is good enough. I can add a doc comment to the attribute itself so when we generate docs it says it there too.

BalestraPatrick (approved):

Issue #283, opened 2022-04-11
Add detailed feature list to README


cpsauer:
Hey build buddy folks,

Looks like a neat new tool you're working on here! For those of us who already use, e.g., Tulsi (or XCHammer) and would consider switching, could I ask how it compares--in goals and status--to the previous Xcode<>Bazel efforts? 

[Sorry if I've such a comparison already exists and I've missed it somewhere.]

I'm also going to tag @BalestraPatrick and @erikkerber -- since from the release notes it sounds like they've been switching over and playing around with it on their codebases!

Thanks so much!
Chris

brentleyjones:
It's on my TODO list to make a comparison to Tulsi, rules_ios, and XCHammer, along with a design/vision document

 In the meantime, this Twitter thread, and the blog post I link to at the start, should be a good holdover: https://twitter.com/brentleyjones/status/1512137776782794753.

cpsauer:
Sweet! Thanks for replying so fast. Good luck and thanks @brentleyjones. Have really enjoyed your 👍🏻s and interacting with you around the GitHub bazel verse, btw. Also, I **_love_** the logo :) Have showed several people just for that.

[Super helpful, btw. Can't use right now, I think, because I'm pretty sure I need the actual build to be in Bazel. But I remain very excited.]

brentleyjones:
@cpsauer FYI, we added a high level design goals document: https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/doc/design/high-level.md.

Thinking more on this, I don't think it would be in good taste to create a document that does a direct comparison to any of the other generators. I think it would instead be best to celebrate all of the things that this generator does well. The README has a small section on this already, but because it's before 1.0, it's a little lite. If that was beefed up, would that be good enough in your eyes? If not, what could we do without directly comparing to other projects?

cpsauer:
Thanks, Brentley! You're great :)

That linked doc especially helps my understanding of how things are evolving around build modes (building with Xcode vs Bazel vs proxy), and the goals behind each mode. Some of my main (implicit) questions were around better understanding those build modes and feeling reassured about being able to produce outputs consistent with Bazel even with custom rules (and select()s) up the dependency chain. 

[...]

PR #334, opened 2022-04-21
Add "High Level Design" document

brentleyjones:


brentleyjones:
I need to flesh out the "Build with X" sections, but I would appreciate some initial feedback for what exists 🙏.

jpsim:
This was a very informative read, I don't have any suggestions to change what's been written.

That being said, it could be helpful to state what the _non-goals_ are to further guide the principles of what would be in or out of scope of this project.

Similarly, comparing the high level design to alternative projects or approaches could also help convey what this project aims to do and how to accomplish it.

brentleyjones:
@jpsim I added a couple more goals and a couple non-goals. Let me know if you had any more non-goals in mind.

brentleyjones:
> Similarly, comparing the high level design to alternative projects or approaches could also help convey what this project aims to do and how to accomplish it.

I agree this should be done (#283), but I'm not ready to do that now. It probably will live in another document, though it will influence this one once I write it.

jpsim (approved):
Big +1 on what's been written.

doc/design/high-level.md
```diff
@@ -0,0 +1,214 @@
...
+  partially work with "Build with Bazel")
+- Less overhead
+
+With all of these benefits, you may be wondering why "Build with Bazel" will be
+the default build mode instead of this one. There are two main reasons:
+
+- The proxy relies on a private API that Xcode uses to communicate with
+  XCBBuildService. This can have, and has had, breaking changes between Xcode
+  versions.
+- To have Xcode use the proxy, you either need to launch it with an environment
```
jpsim, line 188: I'd add what the consequence of this is, which as I understand it, is that it would force all projects opened with that instance of Xcode to use XCBBuildServiceProxy and build with bazel, which could be much more inconvenient than launching Xcode with a specific command.

jpsim (approved):

doc/design/high-level.md
brentleyjones: The proxy is written to only build with bazel for projects generated with rules_xcodeproj. Currently though proxies are written to only work with a single version of Xcode, so setting the environment variable is probably too large of a change. I'm sure we could make the proxy work with a range of Xcode versions though.

I'll add this to the docs :+1:.

doc/design/high-level.md
```diff
@@ -0,0 +1,220 @@
...
+- The project can be further customized by using additional rules_xcodeproj
+  rules, or by returning certain rules_xcodeproj providers from custom rules
+- Outputs are as similar to `bazel build` as possible
+- The project feels as "native" in Xcode as possible
+- All definition of a project exists in `BUILD` files
+- The project can be configured to use either Xcode or Bazel as the build system
+
+It's also worth mentioning a few non-goals of rules_xcodeproj:
+
+- Providing additional build rules
```
BalestraPatrick, line 15: What exactly is a non-goal? Is something we strongly feel against or just a secondary goal? Because we then mention that the project can be customized with additional rules.

brentleyjones: That we are strongly against/won't do. This non-goal is talking about _build_ rules, which would go in your build graph, as opposed to the additional code generation rules.

I'll try to make that more clear.

BalestraPatrick: Nice, I got that after reading a bit more of the doc towards the end, but I think this change improves the clarity here too!

BalestraPatrick (approved):

doc/design/high-level.md
```diff
@@ -0,0 +1,221 @@
...
+
+Here are a few reasons one may want to build with
+Xcode instead of Bazel:
+
+- If a new, possibly beta, version of Xcode is released with a feature that
+  Build with Bazel doesn't support yet, because one of Bazel, rules_apple,
+  rules_swift, or rules_xcodeproj doesn't support it
+- To work around a bug in Bazel, rules_apple, rules_swift, or rules_xcodeproj
+- To compare the Bazel and Xcode build systems or build commands
+- As a step when migrating to Bazel
```
cgrindel, line 127: Do you anticipate getting SwiftUI previews to work using BWB? If not, is that worth adding here?

cgrindel (approved):
Looks great. Thanks for writing this!

doc/design/high-level.md
brentleyjones: SwiftUI previews will work, but they will require that the build graph support them: targets will need to be frameworks or bundles in some way. While rules_xcodeproj won't modify the build graph in any way to do that for teams, we can mention what needs to be done, probably in a `--config` so it looks right for the project generation.

I mention something like this in some of the detailed sections, so I think it's covered?

Issue #293, opened 2022-04-13
Post-processing of files should happen in Bazel


brentleyjones:
Right now we post-process some modulemaps and Info.plists in Xcode run scripts. We will probably post-process some more in the near future (https://github.com/buildbuddy-io/rules_xcodeproj/pull/288#issuecomment-1098303363) Bazel is better at granularity, and the less logic in `generator` the better.

We should create an action for each file that we want to post-process, and return those files in place of whatever file we processed. this probably has to happen at the `xcodeproj` level instead of in the aspect, to prevent duplicate actions, which might require changing around where `File` -> `FilePath` conversions happen.

PR #336, opened 2022-04-22
Extract Bazel Dependencies target creation

brentleyjones:
Code around creating the Bazel Dependencies target was taking up a lot of what was in `Generator+AddTargets.swift`. This will allow adding specific tests for it in the future as well.


tools/generator/src/Generator+AddBazelDependenciesTarget.swift
```diff
@@ -0,0 +1,398 @@
...
+        contains(where: { filePath, _ in
+            return filePath.type == .generated
+                && filePath.path.extension == "modulemap"
+        })
+    }
+
+    var containsInfoPlists: Bool {
+        contains(where: { filePath, _ in
+            return filePath.type == .generated
+                && filePath.path.lastComponent == "Info.plist"
```
jpsim, line 395: Are info plists guaranteed to have this exact file name? Isn't it possible to specify an info plist with a different name? If that's the case, maybe a comment to that effect could be helpful.

Since this file is generated, maybe we control that and it's not a problem?

brentleyjones: They are generated with this name in rules_apple. Also, when #293 is tackled we won't depend on this heuristic anymore.

BalestraPatrick (approved):

PR #338, opened 2022-04-23
Add support for propagating `SwiftInfo` from `cc_library` targets.

cgrindel:
- Added `swift_clang_module_aspect` to the requires list for `xcodeproj_aspect`. This is a `rules_swift` aspect that is used to handle modules that come from `cc_library` targets.
- Updated `xcodeproj_aspect` to collect the `SwiftInfo` providers for the direct dependencies for a rule. The modules provided by `cc_library` targets do not appear in a `SwiftInfo` ’s `direct_modules` or `transitive_modules`. Collecting them in the aspect allows the code access the module information.
- Updated the `is_swift` logic to check for the existence of `SwiftInfo` and check whether it has at least one direct module.
- Updated `command_line` example to include exercise code that leverages `SwiftSyntaxParser` that is loaded via [keith/swift-syntax-bazel](https://github.com/keith/swift-syntax-bazel). This demonstrates the propagation of module information from a `cc_library` to Swift code.


cgrindel:
@brentleyjones I wanted to get your feedback on a couple of items before putting this PR up for review.
1. [The build is failing](https://app.buildbuddy.io/invocation/dec16f76-300e-4b8d-8e19-c0a619d82a0d) in CI, but works on my RH laptop which is running 13.2.1. The compilation error is in [apple/swift-syntax CNodes.swift](https://github.com/apple/swift-syntax). I suspect that this is an Xcode version issue. Can we upgrade the CI machines?
2. You will notice in this PR that I am collecting the `SwiftInfo` from a rule's dependencies in the `xcodeproj_aspect`. Right now, I am plumbing it into the target processing code separate from the `XcodeProjInfo` transitive infos. I was/am tempted to add it to the `XcodeProjInfo` provider. Alternatively, I was thinking that we could collect all of the desired providers in a dictionary or struct (i.e. key would be the provider name, value list of collected providers). Thoughts?
3. You will also notice that I collect the `SwiftInfo` in a struct created by `_create_dependent_provider_info`. I propose that we follow this pattern for the `XcodeProjInfo` collection. I find it a confusing passing around tuples. Thoughts?

examples/ios_app/test/fixtures/bwb_spec.json
```diff
@@ -402,7 +418,21 @@
                 },
                 "third_party/ExampleFramework.framework/ExampleFramework"
             ],
-            "modulemaps": [],
+            "modulemaps": [
+                "third_party/ExampleFramework.framework/Modules/module.modulemap",
```
brentleyjones, line 422: I don't think it's valid that these targets now get modulemaps?

examples/ios_app/test/fixtures/bwb_spec.json
```diff
@@ -429,7 +459,12 @@
                 }
             ],
             "search_paths": {},
-            "swiftmodules": [],
+            "swiftmodules": [
+                {
+                    "_": "ios-x86_64-min15.0-applebin_ios-ios_x86_64-dbg-ST-d8c59b644fe5/bin/Example/Example.swiftmodule",
```
brentleyjones, line 464: I don't think it's valid that this target now gets a swiftmodule?

examples/ios_app/test/fixtures/bwb_spec.json
```diff
@@ -1459,10 +1633,20 @@
                     "Utils/Foo.m"
                 ]
             },
-            "is_swift": false,
+            "is_swift": true,
```
brentleyjones, line 1636: Utils is not a swift target

xcodeproj/internal/target.bzl
```diff
@@ -1147,6 +1152,11 @@ def _process_search_paths(*, cc_info, objc, opts_search_paths):
     search_paths = {}
     if cc_info:
         compilation_context = cc_info.compilation_context
+        header_paths = sets.to_list(sets.make([
+            header.dirname
+            for header in compilation_context.headers.to_list()
```
brentleyjones, line 1157: Why was this needed? Seems disjoint to the rest, so ideally it would be a different PR. It feels very suspect though, given all of the other ones are from attributes telling us the paths, and this is deriving it.

brentleyjones (commented):
Just some pre-review comments and questions.

WORKSPACE
```diff
@@ -74,6 +74,17 @@ swift_library(
     url = "https://github.com/pointfreeco/swift-custom-dump/archive/15fae98653ab3573b458bdb416cbba06e95348a4.tar.gz",
 )
 
+http_archive(
+    name = "com_github_keith_swift_syntax_bazel",
```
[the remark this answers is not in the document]
cgrindel: Sure. In `rules_spm`, I added a number of real world examples using popular dependencies that other folks would use. It helped me feel comfortable that the ruleset worked. If you still don't want to the dependency, let me know.

examples/command_line/lib/BUILD
```diff
@@ -22,16 +23,46 @@ objc_library(
...
+#     srcs = ["SwiftParserExample.swift"],
+#     generated_header_name = "private/SwiftParserExample-Swift.h",
+#     generates_header = True,
+#     module_name = "SwiftParserExample",
+#     visibility = ["//examples/command_line/Tests:__subpackages__"],
+#     deps = [
+#         "@com_github_keith_swift_syntax//:SwiftSyntax",
+#         "@com_github_keith_swift_syntax//:SwiftSyntaxParser",
+#     ],
+# )
```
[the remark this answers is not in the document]
cgrindel: You caught me while I was still cleaning up. 🙂

examples/ios_app/test/fixtures/bwb_spec.json
cgrindel: I wanted to talk about this. What does `is_swift` mean for you? Before this PR, it was a target that provided `SwiftInfo`. I updated the logic to only mark it as true if it also directly provided a module. What do you think it should be?

xcodeproj/internal/target.bzl
cgrindel: The `_CSwiftSyntax` module will not build in Xcode without it. It cannot find the header files. Do you think that it should occur elsewhere?

WORKSPACE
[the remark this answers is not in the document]
brentleyjones: I prefer to not exercise stuff through external deps for this project. By crafting a specific example we control it and the example can't change underneath us, potentially no longer exercising the thing we want from it.

examples/ios_app/test/fixtures/bwb_spec.json
brentleyjones: It means that it is compiling Swift code. In Xcode there are certain things that only make sense for a "Swift target", though we aren't properly doing some of that stuff today (like conditionally setting the last swift migration attributes).

Basically `swift_module.swift` needs to not be `None`.

xcodeproj/internal/target.bzl
brentleyjones: Yeah, one of the search paths in `compilation_context` should handle this.

cgrindel: The following is a dump of the `compilation_context` information for that target:

```
*** CHUCK compilation_context.headers.to_list():
*** CHUCK 0 : <source file Sources/_CSwiftSyntax/include/atomic-counter.h>
*** CHUCK 1 : <source file Sources/_CSwiftSyntax/include/c-syntax-nodes.h>
*** CHUCK compilation_context.quote_includes.to_list():
*** CHUCK 0 : external/com_github_keith_swift_syntax
*** CHUCK 1 : bazel-out/macos-x86_64-min11.0-applebin_macos-darwin_x86_64-fastbuild-ST-cd2b3b8d4835/bin/external/com_github_keith_swift_syntax
*** CHUCK compilation_context.includes.to_list():
*** CHUCK compilation_context.system_includes.to_list():
```

We could use the search paths from the `quote_includes`, but we would need to set them to be recursive. Today, they are set to be non-recursive. However, I question whether that is the best strategy as we can provide a much narrower search list by creating our search path from the actual headers. In this scenario, there is no need to make the search paths recursive.

WORKSPACE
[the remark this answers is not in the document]
cgrindel: OK. I will rework the example to not use an external dep.

examples/ios_app/test/fixtures/bwb_spec.json
cgrindel: OK. So, I will update `targets.get_swift_info` to do the following:

1. Check for the existence of `SwiftInfo`.
2. Check if any of the `direct_modules` has its `swift` attribute set.

Does that work for you?

brentleyjones: Yep!

xcodeproj/internal/target.bzl
brentleyjones: What I mean is, bazel only uses stuff from `compilation_context`, or maybe `ObjcProvider`, and it doesn't derive the paths from `headers`, and it doesn't do recursive searching. I _think_ the real issue here is that it's a `private_dep`, and that is impacting things. Regardless, it's best to see how Bazel sets search paths in this exact case, and we will be able to do the same thing.

cgrindel:
Instead of retrieving the missing private dependency information in this project, @brentleyjones [updated `rules_swift`](https://github.com/bazelbuild/rules_swift/pull/815) to pass along a compilation context much like CcInfo. He is pulling that data through in #342.

Closing this PR.

Issue #94, opened 2022-03-08
Support all `rules_swift` rules


brentleyjones:
[`rules_swift` Rules](https://github.com/bazelbuild/rules_swift/blob/master/doc/rules.md):

- [x] `swift_library`
- [x] `swift_binary`
- [x] `swift_test`
- [x] #93 
- [x] #90 
- [ ] #91 
- [ ] #92

brentleyjones:
While we don't have the "best" support `swift_proto_library` and `swift_grpc_library`, we technically support them. I feel this is finished.

PR #342, opened 2022-04-25
Use `SwiftInfo` `compilation_context`

brentleyjones:
Related to #94.

This adds support for `private_deps`.

| Before | After |
| - | - |
| CleanShot 2022-04-24 at 20 57 45@2x | CleanShot 2022-04-24 at 20 58 39@2x |

### DIFF
--- a/examples/command_line/Tests/SwiftGreetingsTests.swift
+++ b/examples/command_line/Tests/SwiftGreetingsTests.swift
@@ -5,6 +5,6 @@ import XCTest
 class SwiftGreetingsTests: XCTestCase {
 
   func test_greeting() throws {
-    XCTAssertEqual("Swifty", SwiftGreetings.greeting)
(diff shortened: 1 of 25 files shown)


cgrindel:
@brentleyjones So that we are on the same page, I believe that this PR will obsolete #338. The only item that is not covered is the change to the `is_swift` logic. However, that may not be relevant as you are not applying the `swift_clang_module_aspect` now that `rules_swift` passes along the compilation context. Does this sound right to you?

brentleyjones:
Yeah, I think this fully obsoletes #338. We can leave the `is_swift` logic alone for now.

cgrindel:
Are you still working this PR? It is marked as WIP.

cgrindel (approved):

cgrindel:
FYI. I tested your changes against a real-world example using the SwiftSyntax stuff. It compiles.

PR #345, opened 2022-04-25
Fix the order of linking

brentleyjones:
The order archives are linked can matter, so now we set the correct `order` on the link related `depset`s, and we preserve the order in `generator`.


tools/generator/test/Target+Testing.swift
```diff
@@ -19,7 +20,7 @@ extension Target {
         swiftmodules: [FilePath] = [],
         resourceBundles: Set<FilePath> = [],
         inputs: Inputs = Inputs(),
-        links: Set<FilePath> = [],
+        links: OrderedSet<FilePath> = [],
```
BalestraPatrick, line 23: Hopefully in the stdlib one day!

BalestraPatrick (approved):

Issue #73, opened 2022-03-08
Support `objc_import`


brentleyjones:
https://bazel.build/reference/be/objective-c#objc_import

brentleyjones:
I'm pretty sure we already support this, and just need an example in the fixtures. Would you like to do that @maxwellE?

BalestraPatrick:
Isn't this covered by this example already? https://github.com/buildbuddy-io/rules_xcodeproj/blob/e1c0f4f3aa32c5679a888714a2b40f512beed3a2/examples/integration/CommandLine/external/BUILD.tpl#L19-L24

brentleyjones:
This one is, thanks!

PR #346, opened 2022-04-25
Handle frameworks better

brentleyjones:
Part of #73.

This updates `linker_input_files` to be a proper DTO, moves link avoidance into it, removes the manual collection of dynamic frameworks in favor of using `ObjcProvider`, manually removes collection of frameworks into `extra_files` until that is rewritten, and sets up support for `objc_import`.


tools/generator/src/Generator+SetTargetConfigurations.swift
```diff
@@ -106,13 +110,13 @@ Target "\(id)" not found in `pbxTargets`
...
-            if !target.links.isEmpty {
-                let linkFileList = filePathResolver
+            if !target.linkerInputs.staticLibraries.isEmpty {
+                let linkFileList = try filePathResolver
                     .resolve(try target.linkFileListFilePath())
                     .string
                 try targetBuildSettings.prepend(
                     onKey: "OTHER_LDFLAGS",
-                    ["-filelist", "\(linkFileList),$(BUILD_DIR)".quoted]
+                    ["-filelist", linkFileList.quoted]
```
brentleyjones, line 119: This change was to setup support for `cc_import` and `objc_import`. Both of those will have `PROJECT_DIR` relative paths.

tools/generator/src/Generator+CreateFilesAndGroups.swift
```diff
@@ -286,6 +286,7 @@ extension Generator {
         var allInputPaths = extraFiles
         for target in targets.values {
             allInputPaths.formUnion(target.inputs.all)
+            allInputPaths.formUnion(target.linkerInputs.nonGenerated)
```
brentleyjones, line 289: We use `.nonGenerated` instead of `.all` because generated files will be collected via product outputs, or `extraFiles`.

xcodeproj/internal/default_input_file_attributes_aspect.bzl
```diff
@@ -87,6 +91,9 @@ def _default_input_file_attributes_aspect_impl(target, ctx):
             "runtime_deps": [target_type.compile],
             "data": [target_type.resources],
         }
+    elif ctx.rule.kind == "objc_import":
+        xcode_targets = {}
+        excluded = ("archives")
```
brentleyjones, line 96: All changes in here are temporary, until this is rewritten to default to under-collecting instead of over-collecting.

xcodeproj/internal/linker_input_files.bzl
```diff
@@ -21,84 +22,189 @@ def _collect(*, cc_info, objc, is_xcode_target):
...
+        _dynamic_framework_files = dynamic_framework_files,
         _static_framework_files = static_framework_files,
         _is_xcode_library_target = False,
         xcode_library_targets = xcode_library_targets,
     )
 
-def _get_files_to_link(linker_inputs):
-    """Returns the files to link for a target.
+def _get_static_libraries(linker_inputs):
+    """Returns the static libraries needed to link the target.
```
brentleyjones, line 116: I want to remove this in the future. It's duplicative of work done in `_to_dto`, and it's only used by the `required_merges` calculations.

tools/generator/src/Generator+CreateFilesAndGroups.swift
cgrindel: This seems like a salient point. Should we add the above as a code comment?

cgrindel (approved):

PR #349, opened 2022-04-26
Add objc_import example

BalestraPatrick:
Now that we support the `objc_import` rule, this PR adds an example to verify that it works correctly.


examples/ios_app/external/BUILD
```diff
@@ -31,3 +32,24 @@ apple_bundle_import(
...
+    archives = ["ImportableLibrary/libImportableLibrary.a"],
+    visibility = ["//visibility:public"],
+)
+
+# Don't use this directly, instead use it to construct ImportableLibrary.a and
+# depend on that instead:
+# $ bazel build @examples_ios_app_external//:ImportableLibrary --cpu=ios_sim_arm64 --apple_platform_type=ios --ios_sdk_version="15.0"
+# $ bazel build @examples_ios_app_external//:ImportableLibrary --cpu=x86_64 --apple_platform_type=ios --ios_sdk_version="15.0"
+# $ lipo -create -output ... ... libImportableLibrary.a
+objc_library(
```
brentleyjones, line 48: We should probably add `tags = ["manual"]`?

BalestraPatrick: Good idea

brentleyjones (approved):

PR #350, opened 2022-04-26
Write a custom `.lldbinit` for Build with Bazel

brentleyjones:
This will be used in schemes when building with Bazel, to allow debugging to work when we relative paths in our debug information.


tools/generator/src/Generator+AddBazelDependenciesTarget.swift
```diff
@@ -23,7 +23,35 @@ env -i \
...
+    private static func setup(buildMode: BuildMode) -> String {
+        let lldbInit: String
+        if buildMode.requiresLLDBInit {
+            lldbInit = #"""
+
+if [[ -f "$HOME/.lldbinit" ]]; then
+  home_init="command source ~/.lldbinit
+
+"
+else
```
cgrindel, line 35: Is this extra line for formatting in the final output below?

cgrindel (approved):

tools/generator/src/Generator+AddBazelDependenciesTarget.swift
brentleyjones: Yep. Causes a new line to appear after the output, otherwise no new lines when omitted.

PR #352, opened 2022-04-27
Fix incorrect passing of `attrs_info`

brentleyjones:
Passing `None` means to take the inputs as is from `transitive_infos`, which is what we want for `_skip_target`.

### DIFF
--- a/xcodeproj/internal/target.bzl
+++ b/xcodeproj/internal/target.bzl
@@ -900,17 +900,16 @@ def _target_info_fields(
         "xcode_targets": xcode_targets,
     }
 
-def _skip_target(*, target, deps, transitive_infos):
+def _skip_target(*, deps, transitive_infos):
     """Passes through existing target info fields, not collecting new ones.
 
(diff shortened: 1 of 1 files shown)


xcodeproj/internal/target.bzl
```diff
@@ -922,7 +921,7 @@ def _skip_target(*, target, deps, transitive_infos):
             transitive_infos = transitive_infos,
         ),
         inputs = input_files.merge(
-            attrs_info = target[InputFileAttributesInfo],
+            attrs_info = None,
```
cgrindel, line 924: It might be helpful to add the comment from the PR to the code.

cgrindel (approved):

xcodeproj/internal/target.bzl
brentleyjones: I plan to completely revamp input collection "soon" which does away with most of this, so I think it's fine to go without a comment for now.

PR #353, opened 2022-04-27
Add output file collection

brentleyjones:
Similar to our generated inputs collection, we can now specify output groups to cause the building of desired outputs. This is needed for Build with Bazel.

We currently support two modes: build (`b`) and indexing (`i`). The build mode will collect outputs needed by Xcode for a build, run, or test to work. The indexing mode will collect outputs needed by Xcode's indexing processes.

For example, this will build the files needed for indexing of `PathKit`:
```
bazel build --output_groups='i @com_github_kylef_pathkit//:PathKit darwin_arm64-fastbuild'
```

### DIFF
--- a/xcodeproj/internal/output_files.bzl
+++ b/xcodeproj/internal/output_files.bzl
@@ -1,0 +1,192 @@
+"""Module containing functions dealing with target output files."""
+
+load(":output_group_map.bzl", "output_group_map")
+
+# Utility
+
+def _create(*, direct_outputs = None, attrs_info = None, transitive_infos):
+    """Creates the internal data structure of the `output_files` module.
+
+    Args:
+        direct_outputs: A value returned from `_get_outputs`, or `None` if
+            the outputs are being merged.
+        attrs_info: The `InputFileAttributesInfo` for the target.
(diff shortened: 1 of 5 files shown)


xcodeproj/internal/output_files.bzl
```diff
@@ -0,0 +1,158 @@
...
+def _get_outputs(*, bundle_info, id, swift_info):
+    build = []
+    index = []
+
+    if bundle_info:
+        build.append(bundle_info.archive)
+
+    # TODO: Collect headers for CC targets
+
+    # TODO: Determine which of these are actually needed for build vs index
```
brentleyjones, line 60: These TODOs I'll address when iterating on BwB.

xcodeproj/internal/output_files.bzl
```diff
@@ -0,0 +1,158 @@
+"""Module containing functions dealing with target output files."""
+
+load(":output_group_map.bzl", "output_group_map")
+
+# Utility
+
+def _create(*, direct_outputs = None, attrs_info = None, transitive_infos):
```
cgrindel, line 7: Not sure if the signature for this has solidified yet. If they are, we may want to add doc comments.

cgrindel (approved):

Issue #101, opened 2022-03-08
Create schemes


brentleyjones:
Currently we let Xcode create the schemes for us. This is finicky, and we will want to customize how the schemes are created, especially when [Building with Bazel](https://github.com/buildbuddy-io/rules_xcodeproj/milestone/2).

PR #370, opened 2022-05-01
Add XcodeProj extensions and models for Xcode scheme generation

cgrindel:
Related to #101, #371.

This is part 2 of 3 parts for the Xcode scheme generation. The follow-on PR contains the meat of the logic. Breaking out these peripheral changes will make it easier to review the next PR.

- Add `isTestBundle` to `PBXProductType`
- Add `getBuildableName`, `createBuildableReference`, `isTestable`, and `isLaunchable` to `PBXTarget`
- Add `getContainerReference` to `FilePathResolver`


tools/generator/src/PBXProductType+Extensions.swift
```diff
@@ -90,38 +90,49 @@ extension PBXProductType {
...
+            case .xpcService: return "wrapper.xpc-service"
+            case .ocUnitTestBundle: return "wrapper.cfbundle"
+            case .xcodeExtension: return "wrapper.app-extension"
+            case .instrumentsPackage: return "com.apple.instruments.instrdst"
+            case .intentsServiceExtension: return "wrapper.app-extension"
+            case .onDemandInstallCapableApplication: return "wrapper.application"
+            case .metalLibrary: return nil
+            case .driverExtension: return "wrapper.driver-extension"
+            case .systemExtension: return "wrapper.system-extension"
+            case .none: return nil
```
cgrindel, line 135: The formatting for this didn't match the `switch` statement below. So, I updated this one to look like it.

tools/generator/src/PBXProductType+Extensions.swift
```diff
@@ -90,38 +90,49 @@ extension PBXProductType {
...
-        case .ocUnitTestBundle: return "wrapper.cfbundle"
-        case .xcodeExtension: return "wrapper.app-extension"
-        case .instrumentsPackage: return "com.apple.instruments.instrdst"
-        case .intentsServiceExtension: return "wrapper.app-extension"
-        case .onDemandInstallCapableApplication: return "wrapper.application"
-        case .metalLibrary: return nil
-        case .driverExtension: return "wrapper.driver-extension"
-        case .systemExtension: return "wrapper.system-extension"
-        case .none: return nil
+            case .application: return "wrapper.application"
```
brentleyjones, line 106: This shouldn't have an indent.

tools/generator/src/PBXProductType+Extensions.swift
brentleyjones: The one below is wrong then 😄.

tools/generator/src/PBXProductType+Extensions.swift
cgrindel: I will remove the indents.

tools/generator/src/PBXTarget+Extensions.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+import XcodeProj
+
+public extension PBXTarget {
+    var buildableName: String {
+        return productName ?? name
```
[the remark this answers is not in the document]
cgrindel: All of the `PBXTarget` instances that I have seen have had a value for `productName`. In terms of whether it should default to name, I think that it is reasonable given the following:
- We need a value for `buildableName`
- The `productName` is optional
- The `productName` seems to best align with the values that are generated by Xcode itself
- The `name` value still conveys the intent given that we opted for one scheme per target

tools/generator/src/PBXTarget+Extensions.swift
```diff
@@ -0,0 +1,37 @@
...
+        )
+    }
+
+    var schemeName: String {
+        // The XcodeProj write logic does not like slashes (/) in the scheme
+        // name. It fails to write with a missing folder error.
+        return buildableName.replacingOccurrences(
+            of: #"[/]"#,
+            with: "_",
+            options: .regularExpression
```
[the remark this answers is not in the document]
cgrindel: I don't think that I follow. The `XcodeProj` code appears to use the name directly as the path to the scheme file. When slashes came through, I saw missing folder errors. I did not dig into the `XcodeProj` code.

Regarding the use of regex, you are correct. I had regex on the brain thinking that I needed to replace more stuff. I switched to a simpler replace.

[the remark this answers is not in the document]
cgrindel: I just added #371 to track addressing the issue in `XcodeProj`. I updated the code to reference this issue.

tools/generator/src/PBXTarget+Extensions.swift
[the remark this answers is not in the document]
brentleyjones: I mean, I think `buildableName` needs to reference a `productName`. I think we can't use `name` because the scheme won't actually work as expected then.

I would rather we return `nil` here and then throw where we try to use the nil value.

tools/generator/src/XcodeProjContainerReference.swift
```diff
@@ -0,0 +1,29 @@
+import PathKit
+
+/// In XcodeProj, a `referencedContainer` in a `XCScheme.BuildableReference`
+/// accepts a string in the format `container:<path-to-xcodeproj-dir>`. This
+/// type represents that structure.
+public struct XcodeProjContainerReference: Equatable {
```
brentleyjones, line 6: It seems odd that we need a whole type for this. Just wondering why it's not a simple computed string property on another type, like `Project`.

cgrindel: I debated this a bit. You will see in the next PR that this type is an argument to a function that creates the schemes.

```swift
    static func createXCSchemes(
        referencedContainer: XcodeProjContainerReference,
        pbxTargets: [TargetID: PBXNativeTarget]
    ) throws -> [XCScheme] {
    // ...
    }
```

I feel that it makes it clearer what is being passed along. Here is what it looked like originally:

```swift
    static func createXCSchemes(
        referencedContainer: String,
        pbxTargets: [TargetID: PBXNativeTarget]
    ) throws -> [XCScheme] {
    // ...
    }
```

Without some doc that could be easily missed, it is not clear what this value should be. I dug around the `XcodeProj` code looking for a factory or an explanation of what the format should be. In the end, all of the values that I found looked like `container:path/to/Foo.xcodeproj`. I defined the type so that it was clear what this value is.

brentleyjones: What about a version that had `Project` as the parameter and `Project` had an extension on it to return this computed property? This just seems a bit heavy, and (oddly in contrast to your desired reason for it existing) too indirect.

It just confused me when I saw it, and even more when I looked at the implementation. I'm not opposed to it, just really wondering if it's needed.

cgrindel:
FYI. I forgot to add `defaultBuildConfigurationName` to `PBXTarget`. I updated this PR to include it along with its tests.

tools/generator/src/XcodeProjContainerReference.swift
brentleyjones: I guess instead of project you would do `FilePathResolver` which has the path to the project as a property already.

tools/generator/test/PBXTarget+ExtensionsTests.swift
```diff
@@ -0,0 +1,84 @@
+import XcodeProj
+import XCTest
+
+@testable import generator
+
+class PBXTargetExtensionsTests: XCTestCase {
+    let pbxTarget = PBXTarget(name: "chicken", productName: "MyChicken")
+
+    func test_buildableName_WithProductName() throws {
```
brentleyjones, line 9: Nit: can we use camel case?
```suggestion
    func test_buildableName_withProductName() throws {
```

tools/generator/src/PBXTarget+Extensions.swift
```diff
@@ -0,0 +1,39 @@
...
+        // name. It fails to write with a missing folder error.
+        return buildableName.replacingOccurrences(of: "/", with: "_")
+    }
+
+    var isTestable: Bool {
+        return productType?.isTestBundle ?? false
+    }
+
+    var isLaunchable: Bool {
+        return productType?.isExecutable ?? false
```
brentleyjones, line 32: We instead need a `isLaunchable` on `PBXProductType` that we reference here, as `isExecutable` includes a lot of non-launchable things, mainly frameworks. `isExecutable` just means it has an executable Mach-O binary (versus static or none).

tools/generator/test/PBXTarget+ExtensionsTests.swift
cgrindel: Done

tools/generator/src/PBXTarget+Extensions.swift
cgrindel: I updated my code to use the `isLaunchable` on `PBXProductType`.

tools/generator/src/XcodeProjContainerReference.swift
cgrindel: I removed `XcodeProjContainerReference` and added `containerReference` to `FilePathResolver`.

tools/generator/src/PBXTarget+Extensions.swift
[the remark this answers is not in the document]
cgrindel: To avoid duplicating the check logic, I replaced `buildableName` with `getBuildableName`. This resulted in `schemeName` changing to `getSchemeName`.

tools/generator/src/PBXTarget+Extensions.swift
```diff
@@ -2,26 +2,32 @@ import Foundation
...
     }
 
-    var schemeName: String {
+    func getSchemeName() throws -> String {
         // GH371: Update XcodeProj to support slashes in the scheme name.
         // The XcodeProj write logic does not like slashes (/) in the scheme
         // name. It fails to write with a missing folder error.
-        return buildableName.replacingOccurrences(of: "/", with: "_")
+        return try getBuildableName()
+        .replacingOccurrences(of: "/", with: "_")
```
brentleyjones: nit:
```suggestion
            .replacingOccurrences(of: "/", with: "_")
```

tools/generator/src/FilePathResolver.swift
```diff
@@ -101,4 +106,5 @@ struct FilePathResolver: Equatable {
             return internalDir + filePath.path
         }
     }
+
```
brentleyjones, line 109: nit:
```suggestion
```

brentleyjones (approved):

tools/generator/src/PBXTarget+Extensions.swift
cgrindel: You must have seen this before I fixed it. 🙂

tools/generator/src/FilePathResolver.swift
cgrindel: Done

PR #372, opened 2022-05-02
Add support for sdk_dylibs

BalestraPatrick:
This adds support for `sdk_dylibs` which are passed to the linker as `-l` flags.

### DIFF
--- a/examples/ios_app/CoreUtilsObjC/Answers.mm
+++ b/examples/ios_app/CoreUtilsObjC/Answers.mm
@@ -1,10 +1,11 @@
 #import <CoreUtils/Answers.h>
+#include <limits>
 
 @implementation Answers
 
 - (NSInteger)answer {
     NSLog(@"%@", [Bar new].baz);
-    return 42;
+    return std::numeric_limits<int>::max();
 }
 
 @end
--- a/xcodeproj/internal/target.bzl
+++ b/xcodeproj/internal/target.bzl
@@ -283,11 +283,7 @@ The xcodeproj rule requires {} rules to have a single library dep. {} has {}.\
         cc_info = cc_info,
         build_settings = build_settings,
     )
-    _process_sdk_frameworks(
-        objc = objc,
-        build_settings = build_settings,
-    )
-    _process_sdk_dylibs(
+    _process_sdk_links(
         objc = objc,
         build_settings = build_settings,
     )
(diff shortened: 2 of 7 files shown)


examples/ios_app/CoreUtilsObjC/Answers.mm
```diff
@@ -1,10 +1,11 @@
 #import <CoreUtils/Answers.h>
+#include <limits>
 
 @implementation Answers
 
 - (NSInteger)answer {
     NSLog(@"%@", [Bar new].baz);
-    return 42;
+    return std::numeric_limits<int>::max();
```
BalestraPatrick, line 8: I'm not sure if there's a better case for exercising this. Even before setting `sdk_dylibs`, the previous code compiled successfully (after changing it to a `.mm`), so I imagine there's some autolinking going on for a `.mm` file. I only noticed that we were missing these flags when comparing two projects (one building with standard Xcode for many years and one with rules_xcodeproj).

xcodeproj/internal/target.bzl
```diff
@@ -937,6 +945,21 @@ def _process_sdk_frameworks(*, objc, build_settings):
          build_settings.get("OTHER_LDFLAGS", [])),
     )
 
+def _process_sdk_dylibs(*, objc, build_settings):
```
brentleyjones, line 948: Can we roll it into `_process_sdk_frameworks`, and rename that function instead? Maybe `_process_sdk_links`?

brentleyjones: Also can we make sure they appear in the same order that Bazel puts them in?

BalestraPatrick: Good idea. I verified it by building the example app:
```
-stdlib=libc++
-std=gnu++11
-arch
arm64
-Xlinker
-objc_abi_version
-Xlinker
2
-fobjc-link-runtime
-ObjC
-framework
Foundation
-framework
ExampleFramework
-framework
ExternalFramework
-weak_framework
SwiftUI
-lc++
-filelist
bazel-out/ios-sim_arm64-min15.0-applebin_ios-ios_sim_arm64-fastbuild-ST-eeab8ca40dab/bin/Example/Example-linker.objlist
```

Looks like it's correct as it is now. First `-framework`, `-weak_framework` and then `-l`.

brentleyjones (approved):

PR #373, opened 2022-05-02
Work around `DYLD_LIBRARY_PATH` issue

brentleyjones:
When launching an application with Xcode, it sets `DYLD_LIBRARY_PATH` to include the `BUILT_PRODUCTS_DIR` of every target in the build graph of the target being run. For our projects, this is a lot of paths, and with enough targets it can cause a crash in `dlyd` here: https://github.com/apple-oss-distributions/dyld/blob/419f8cbca6fb3420a248f158714a9d322af2aa5a/dyld/DyldProcessConfig.cpp#L1027

```
Exception Type:  EXC_BAD_ACCESS (SIGBUS)
Exception Subtype: KERN_PROTECTION_FAILURE at 0x00000001228f8000
Exception Codes: 0x0000000000000002, 0x00000001228f8000
VM Region Info: 0x1228f8000 is in 0x1228f8000-0x122930000;  bytes after start: 0  bytes before end: 229375
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      mapped file                 1228f4000-1228f8000    [   16K] rw-/rw- SM=COW  ...t_id=a04cc64f
--->  mapped file                 1228f8000-122930000    [  224K] r--/r-- SM=COW  ...t_id=92ffeb4f
      GAP OF 0x45474000 BYTES
      Stack Guard                 167da4000-16b5a8000    [ 56.0M] ---/rwx SM=NUL
Exception Note:  EXC_CORPSE_NOTIFY
Termination Reason: SIGNAL 10 Bus error: 10
Terminating Process: exc handler [58981]
```

```
(lldb) bt
* thread #1, stop reason = EXC_BAD_ACCESS (code=2, address=0x1214ac000)
  * frame #0: 0x00000001214320a8 dyld`memmove + 88
    frame #1: 0x0000000121431cb0 dyld`strcpy + 48
    frame #2: 0x00000001214394dc dyld`dyld4::ProcessConfig::PathOverrides::setString(dyld4::ProcessConfig::Process const&, char const*&, char const*) + 180
    frame #3: 0x0000000121438ba8 dyld`dyld4::ProcessConfig::PathOverrides::addEnvVar(dyld4::ProcessConfig::Process const&, dyld4::ProcessConfig::Security const&, char const*, bool, char*) + 544
    frame #4: 0x00000001214388a8 dyld`dyld4::ProcessConfig::PathOverrides::PathOverrides(dyld4::ProcessConfig::Process const&, dyld4::ProcessConfig::Security const&, dyld4::ProcessConfig::Logging const&, dyld4::ProcessConfig::DyldCache const&, dyld4::SyscallDelegate&) + 172
    frame #5: 0x0000000121436fc0 dyld`dyld4::ProcessConfig::ProcessConfig(dyld4::KernelArgs const*, dyld4::SyscallDelegate&) + 140
    frame #6: 0x0000000121434fcc dyld`start + 328
```

To work around this, we set `BUILT_PRODUCTS_DIR` to `$(BUILD_DIR)`. This alone has three issues though:

1. Xcode copies swiftmodules to `BUILT_PRODUCTS_DIR`, which can result in conflicts for XCBuild when multiple configurations (e.g. different minimum OSes) produce the same swiftmodule
2. swiftmodules aren't where they should be for compilation because of "1"
3. Launchable products need to be at `BUILT_PRODUCTS_DIR` for schemes to work, but the binaries and bundles are still at `CONFIGURATION_BUILD_DIR`

For "3", we override `BUILT_PRODUCTS_DIR` to be `$(CONFIGURATION_BUILD_DIR)` only for launchable targets. Since these are the top level targets, it's only one additional path added to `DYLD_LIBRARY_PATH`, which is fine.

For "1" and "2" though, we had to get clever.  By using Xcode's "Deployment Location" functionality we can affect `BUILT_PRODUCTS_DIR` indirectly with some slight overhead of symlinks. When `DEPLOYMENT_LOCATION = YES`, then `BUILT_PRODUCTS_DIR` becomes the value of `$(DSTROOT)/$(INSTALL_PATH)` and products are symlinked (or in the case of swiftmodules, copied) to the value of `$(CONFIGURATION_BUILD_DIR)` during the build. So we set `DSTROOT` to `$(PROJECT_TEMP_DIR)` and `INSTALL_PATH` to `$(BAZEL_PACKAGE_BIN_DIR)`. This allows us to reuse the same `TARGET_TEMP_DIR` folder structure for the actual outputs.

### DIFF
--- a/tools/generator/src/Generator+CreateProject.swift
+++ b/tools/generator/src/Generator+CreateProject.swift
@@ -21,9 +21,16 @@ extension Generator {
         buildSettings.merge([
             "BAZEL_EXTERNAL": "$(LINKS_DIR)/external",
             "BAZEL_OUT": "$(BUILD_DIR)/real-bazel-out",
+            // `BUILT_PRODUCTS_DIR` isn't actually used by the build, since
+            // `DEPLOYMENT_LOCATION` is set. It does prevent `DYLD_LIBRARY_PATH`
+            // from being modified though.
+            "BUILT_PRODUCTS_DIR": "$(BUILD_DIR)",
             "CONFIGURATION_BUILD_DIR": "$(BUILD_DIR)/$(BAZEL_PACKAGE_BIN_DIR)",
+            "DEPLOYMENT_LOCATION": true,
+            "DSTROOT": "$(PROJECT_TEMP_DIR)",
             "GEN_DIR": "$(LINKS_DIR)/gen_dir",
             "LINKS_DIR": "$(INTERNAL_DIR)/links",
+            "INSTALL_PATH": "$(BAZEL_PACKAGE_BIN_DIR)",
             "INTERNAL_DIR": """
 $(PROJECT_FILE_PATH)/\(filePathResolver.internalDirectoryName)
 """,
--- a/tools/generator/src/PBXProductType+Extensions.swift
+++ b/tools/generator/src/PBXProductType+Extensions.swift
@@ -47,6 +47,38 @@ extension PBXProductType {
         }
     }
 
+    var isExecutable: Bool {
+        switch self {
+        case .application,
+             .framework,
+             .xcFramework, // Could be
+             .dynamicLibrary,
+             .bundle, // Could be
+             .unitTestBundle,
+             .uiTestBundle,
+             .appExtension,
+             .commandLineTool,
+             .watchApp,
+             .watch2App,
+             .watch2AppContainer,
+             .watchExtension,
+             .watch2Extension,
+             .tvExtension,
+             .messagesApplication,
+             .messagesExtension,
+             .xpcService,
+             .ocUnitTestBundle,
+             .xcodeExtension,
+             .intentsServiceExtension,
+             .onDemandInstallCapableApplication,
+             .driverExtension,
+             .systemExtension:
+            return true
+        default:
+            return false
+        }
+    }
+
     var isFramework: Bool {
         switch self {
         case .framework,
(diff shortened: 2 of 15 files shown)


tools/generator/src/PBXProductType+Extensions.swift
```diff
@@ -47,6 +47,38 @@ extension PBXProductType {
         }
     }
 
+    var isExecutable: Bool {
```
brentleyjones, line 50: I move this up to be alphabetical.

tools/generator/src/Generator+CreateProject.swift
```diff
@@ -21,9 +21,16 @@ extension Generator {
         buildSettings.merge([
             "BAZEL_EXTERNAL": "$(LINKS_DIR)/external",
             "BAZEL_OUT": "$(BUILD_DIR)/real-bazel-out",
+            // `BUILT_PRODUCTS_DIR` isn't actually used by the build, since
+            // `DEPLOYMENT_LOCATION` is set. It does prevent `DYLD_LIBRARY_PATH`
+            // from being modified though.
+            "BUILT_PRODUCTS_DIR": "$(BUILD_DIR)",
             "CONFIGURATION_BUILD_DIR": "$(BUILD_DIR)/$(BAZEL_PACKAGE_BIN_DIR)",
+            "DEPLOYMENT_LOCATION": true,
+            "DSTROOT": "$(PROJECT_TEMP_DIR)",
```
BalestraPatrick, line 30: Just curious, where did you find these vars?

brentleyjones: I explore https://xcodebuildsettings.com a lot when I'm trying to get Xcode to do something odd.

BalestraPatrick (approved):

PR #374, opened 2022-05-02
Fix off-by-one error for entitlements

brentleyjones:
XCBuild seems to look at the contents of `CODE_SIGN_ENTITLEMENTS` early in the build. We need the path to match what is in `generated.copied.xcfilelist` for XCBuild to use the generated version.

Fixes this error:
> Build input file cannot be found: '/Users/brentley/Library/Developer/Xcode/DerivedData/bwx-dutupusxoomkshbjcojnxbrklaer/Build/Products/bazel-out/applebin_ios-ios_x86_64-fastbuild-ST-d3e756bfe7fd/bin/Example/Example_entitlements.entitlements'


BalestraPatrick (approved):
I ran into this only once today after I ran `bazel clean` so I was gonna look into it soon, thanks!

brentleyjones:
I'm going to push another PR that uses the GEN_DIR for more of these files, even if they aren't technically needed, as a better pattern to follow.

PR #377, opened 2022-05-02
Remove non-existent PrioritityQueue target from Swift collections

erikkerber:
Found via setting `--incompatible_disallow_empty_glob`

`PriorityQueueModule` does not exist in Swift collections 1.0.2


brentleyjones (approved):
I was looking at an unreleased version I guess.

PR #422, opened 2022-05-06
Extract `.lldbinit` creation to a script file

brentleyjones:
This is the start of moving parts of Run Scripts into script files.

There are numerous benefits to having them as files instead of embedded in the Run Script itself:
- Easier to review in PRs
- Easier to iterate on
- Easier for consumers to customize


cgrindel (approved):
Nice!

PR #424, opened 2022-05-06
Support disambiguation of targets with names that only differ by case

cgrindel:


tools/generator/test/DisambiguateTargetsTests.swift
```diff
@@ -23,12 +23,23 @@ final class DisambiguateTargetsTests: XCTestCase {
...
             ),
+            // The following targets only differ by case
+            "C 1": Target.mock(
+                label: "//c:A",
+                product: .init(type: .staticLibrary, name: "A", path: "")
+            ),
+            "C 2": Target.mock(
+                label: "//c:a",
+                product: .init(type: .application, name: "a", path: "")
+            ),
```
brentleyjones, line 34: Can you add an example to `test_productType` to show that it does this for non-label based results as well?

tools/generator/test/DisambiguateTargetsTests.swift
```diff
@@ -60,11 +71,20 @@ final class DisambiguateTargetsTests: XCTestCase {
...
+            "C 2": Target.mock(
+                product: .init(type: .application, name: "c", path: "")
+            ),
         ]
         let expectedTargetNames: [TargetID: String] = [
             "A 1": "A (Library)",
             "A 2": "A (App)",
             "B": "B",
+            "C 1": "C (Library)",
+            "C 2": "c (App)",
```
cgrindel, line 87: @brentleyjones I just want to confirm that this is the output you would expect for the `producType` test. If so, I will need to modify the following conditional 
```swift
if labelsByName[normalizedName]!.count == 1 {
```

brentleyjones: That is the expected output, yes.

brentleyjones: That condition is fine. We just need to store normalized labels in it here:
```
labelsByName[target.name, default: []].insert(target.label)
```

brentleyjones: And then the label test will need to use unique (not just case differences) labels.

brentleyjones (approved):

PR #430, opened 2022-05-09
Embed configuration into `generated_inputs` output group

brentleyjones:
This makes the configuration group similar to the ones used for Build with Bazel outputs. After this change we will be able to grep the output to ensure that the output groups we expected were actually built, leading to better error messages.


tools/generator/src/Generator+AddBazelDependenciesTarget.swift
```diff
@@ -184,6 +188,9 @@ if [ "$ACTION" == "indexbuild" ]; then
   # We use a different output base for Index Build to prevent normal builds and
   # indexing waiting on bazel locks from the other
   output_base="$OBJROOT/bazel_output_base"
+  convenience_symlinks=ignore
+else
+  convenience_symlinks=normal
```
cgrindel, line 193: Is this being used somewhere? Is it supposed to be passed to `--experimental_convenience_symlinks`?

brentleyjones: Accidental commit from when I was trying something else. I'll remove.

cgrindel (approved):

PR #434, opened 2022-05-09
Verify output group maps are actually built

brentleyjones:
This catches the case when something has caused Bazel to use a different configuration for the project, but the project hasn't been regenerated yet.


brentleyjones:
@cgrindel @BalestraPatrick

BalestraPatrick (approved):

PR #440, opened 2022-05-16
Provide Bazel environment variables for launchable schemes

cgrindel:
- Provide `BUILD_WORKSPACE_DIRECTORY` and `BUILD_WORKING_DIRECTORY` for launchable schemes.

### DIFF
--- a/tools/generator/src/XCSchemeEnvrionmentVariables+Extensions.swift
+++ b/tools/generator/src/XCSchemeEnvrionmentVariables+Extensions.swift
@@ -1,6 +1,9 @@
 import XcodeProj
 
 extension Array where Element == XCScheme.EnvironmentVariable {
+    /// Provides the Bazel-specific environment variables for Build with Bazel
+    /// launch action.
+    ///
     /// For more information:
     /// https://docs.bazel.build/versions/main/user-manual.html#run
     static let bazelLaunchVariables: [XCScheme.EnvironmentVariable] = [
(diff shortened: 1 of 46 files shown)


tools/generator/test/CreateXCSchemesTests.swift
```diff
@@ -179,6 +183,13 @@ fi
             file: file,
             line: line
         )
+        XCTAssertEqual(
+            testAction.customLLDBInitFile,
+            expectedCustomLLDBInitFile,
+            "testAction.customLLDBInitFile did not match for \(scheme.name)",
+            file: file,
+            line: line
+        )
```
cgrindel, line 192: This was grouped with the launch action assertions. I just moved it to be with the test action assertions.

tools/generator/src/XCSchemeEnvrionmentVariables+Extensions.swift
```diff
@@ -0,0 +1,25 @@
...
+        .init(
+            variable: "BUILD_WORKSPACE_DIRECTORY",
+            value: "$(SRCROOT)",
+            enabled: true
+        ),
+        .init(
+            variable: "BUILD_WORKING_DIRECTORY",
+            // This is a poor substitute for the working directory.
+            // Preferably, it would be $(PWD) or something similar.
+            // Unfortunately, none of the following worked:
```
brentleyjones, line 19: We can remove this last line. Xcode doesn't have a concept of a working directory. Schemes can have a working directory though, and they default to something inside of derives data. Maybe we should set this value to whatever that defaults to, or change the working directory to be SRCROOT?

cgrindel: I see [a way to specify that you want to set a custom working directory](https://github.com/tuist/XcodeProj/blob/main/Sources/XcodeProj/Scheme/XCScheme%2BLaunchAction.swift#L45). Unfortunately, I don't see a way to set the actual custom working directory. 

When I set it in the Xcode UI, I see the following in the scheme file:

```xml
<LaunchAction
      useCustomWorkingDirectory = "YES"
      customWorkingDirectory = "/Users/chuckgrindel"
```

Do you know if there is a macro for the derives data location? I will poke around to try and find one.

cgrindel: It looks like `BUILD_PRODUCTS_DIR` evaluates to the directory that Xcode uses as the current working directory.

brentleyjones (approved):

PR #442, opened 2022-05-18
Apply filtering logic to `generated` `extra_files`, and `xccurrentversions` as well

brentleyjones:
This prevents the `generated_inputs` output group from having unnecessary files in it. This filtering is very redundant, and error prone (as shown by this and previous changes applying missing filtering). This will get addressed with the forthcoming input file collection refactor.

### DIFF
--- a/xcodeproj/internal/input_files.bzl
+++ b/xcodeproj/internal/input_files.bzl
@@ -313,7 +313,10 @@ https://github.com/buildbuddy-io/rules_xcodeproj/issues/new?template=bug.md
             xccurrentversions,
             transitive = [
                 info.inputs.xccurrentversions
-                for _, info in transitive_infos
+                for attr, info in transitive_infos
+                if (not attrs_info or
+                    info.target_type in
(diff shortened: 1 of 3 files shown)


xcodeproj/internal/input_files.bzl
```diff
@@ -321,7 +321,10 @@ https://github.com/buildbuddy-io/rules_xcodeproj/issues/new?template=bug.md
             generated,
             transitive = [
                 info.inputs.generated
-                for _, info in transitive_infos
+                for attr, info in transitive_infos
+                if (not attrs_info or
+                    info.target_type in
+                    attrs_info.xcode_targets.get(attr, [None]))
```
cgrindel, line 327: Curious. Why `[None]` vs `[]`?

cgrindel (approved):

xcodeproj/internal/input_files.bzl
brentleyjones: Because it does an `in` check, and `info.target_type` can be `None`, and that needs to be `True`.

PR #443, opened 2022-05-18
Make filelist path check more resilient to various Bazel flags

brentleyjones:
There are flags or circumstances that can cause Bazel's log to print different paths for outputs. By using `--symlink_prefix=/` we cause Bazel to output the full path to its outputs, which we can reliably match against.


jpsim (approved):

erikkerber:
Works for us as well

PR #449, opened 2022-05-18
Fix copying of resource bundles

brentleyjones:
This partially reverts e5ccba4edd16dd13335a8ceaa6457c7648c1173a for bundles. Seems Xcode will use the `BUILT_PRODUCTS_DIR` of the app if this product association isn't setup. We might be able to fix this another way in the future.


erikkerber (approved):
Our app now builds and launches with BwX

PR #454, opened 2022-05-20
Use build variables for bazel output paths

brentleyjones:
This is needed for a future change where we want these paths to vary based on simulator vs device builds.


tools/generator/src/Generator+AddTargets.swift
```diff
@@ -644,11 +632,13 @@ chmod u+w "$header"
         return #"""
 \#(copyGeneratedHeader)\#
 # Copy swiftmodule
+SAVEIFS=$IFS; IFS=$'\n'
+swiftmodule=($BAZEL_OUTPUTS_SWIFTMODULE)
+IFS=$SAVEIFS
```
brentleyjones, line 637: @jpsim do you know if there is there a way to do this another way? Arrays in build settings seem to just be whitespace separated, so if I use those I don't think I can support paths with spaces.

brentleyjones: If I do it this way, I guess this is probably better:
```suggestion
IFS=$'\n' read -r -a swiftmodule <<< "$BAZEL_OUTPUTS_SWIFTMODULE"
```

cgrindel: Your revised version looks right to me.

tools/generator/src/Generator+SetTargetConfigurations.swift
```diff
@@ -201,6 +201,34 @@ $(CONFIGURATION_BUILD_DIR)
...
+                    buildSettings["BAZEL_OUTPUTS_SWIFT_GENERATED_HEADER"] =
+                        try filePathResolver.resolve(
+                            generatedHeader,
+                            useOriginalGeneratedFiles: true
+                        )
+                        .string
+                }
+            }
+
+            if let productOutput = target.outputs.product {
```
jpsim, line 223: Should we throw any errors in the case where some of these aren't set?

Looking at the run scripts, it seems like there might be some environment variables that are always expected to be set?

tools/generator/src/Generator+AddTargets.swift
jpsim: Agreed

jpsim (approved):
Validated our simulator and device builds both still work with this change

tools/generator/src/Generator+SetTargetConfigurations.swift
brentleyjones: What do you mean by when these are unset? They are optional because it will be valid for them to be unset for various types of targets.

Unless I'm misunderstanding what you are asking (probably).

tools/generator/src/Generator+AddTargets.swift
brentleyjones: I couldn't get it to read more than one element this way. I'm just leaving it as is for now.

tools/generator/src/Generator+SetTargetConfigurations.swift
jpsim: What I'm saying is that the scripts that check these variables (e.g .`BAZEL_OUTPUTS_PRODUCT`) do so unconditionally, so are there any cases where the script would try to run but the variable is not set? It's possible that the scripts are only configured to run in the generated projects under the same conditions that would cause these variables to be set, so they're guaranteed to be present when the script is run, but confirming that would require me to look at all the plumbing involved.

Finally, I'm slightly surprised to hear that some targets would produce no `outputs.product`, it seems like a target that produces nothing is unlikely, but I'm probably missing something.

brentleyjones: `outputs.product` is used for bundles and binaries. We don't set it for other outputs, because we don't need to copy the `.a` for example.

And right now, in this change, we only write out the script that uses these variables if they are set. #455 changes the script to check for the existence of the variables.

PR #455, opened 2022-05-20
Extract Copy Bazel Outputs to a script file

brentleyjones:
Similar reasons as #422.


xcodeproj/internal/bazel_integration_files/copy_outputs.sh
```diff
@@ -0,0 +1,87 @@
...
+readonly forced_swift_compile_file="$1"
+readonly product_basename="$2"
+readonly exclude_list="$3"
+
+if [[ "$ACTION" == indexbuild ]]; then
+  # Write to "$BAZEL_BUILD_OUTPUT_GROUPS_FILE" to allow next index to catch up
+  echo "i $BAZEL_TARGET_ID" > "$BAZEL_BUILD_OUTPUT_GROUPS_FILE"
+else
+  # Copy product
+  if [[ -n ${BAZEL_OUTPUTS_PRODUCT+x} ]]; then
```
brentleyjones, line 14: Maybe I shouldn't get so cleaver here and just do this?
```suggestion
  if [[ -n "${BAZEL_OUTPUTS_PRODUCT:-}" ]]; then
```

cgrindel: 👍

cgrindel (approved):

Issue #461, opened 2022-05-23
Add `xcodeproj_test` to test configured `xcodeproj` running the installer and then executing select `xcodebuild` actions against the generated Xcode project


cgrindel:
## Goals
- Provide assurances to `rules_xcodeproj` contributors that changes to the underlying rules work.
- Allow clients of `rules_xcodeproj` to test that the schemes for configured projects build and tests execute.

## Design/Usage
- Provide `xcodeproj_test` that handles the test setup and Xcode project installation.
- Provide a default test runner that executes `xcodebuild` commands against the Xcode project.
- Allow clients to provide their own test runner.

```python
xcodeproj(
    name = "foo_bwx",
    project_name = "Foo (BWX)",
    build_mode = "xcode",
    tags = ["manual"],
    targets = XCODEPROJ_TARGETS,
)

xcodebuild_runner(
    name = "tool_builder",
    command = "build -scheme \"tool\"",
)

xcodeproj_test(
    name = "tool_build_test",
    xcodeproj = ":foo_bwx",
    runner = ":tool_builder",
)
```

brentleyjones:
I'm moving this to Post 1.0, because I don't think it should block the release, not because I don't think it's important. If someone is able to take this on I would love to help get it over the finish line.

PR #462, opened 2022-05-23
Add `--source` and `--destination` options to installer template

cgrindel:
Related to #461.

These options will be exercised in tests once `xcodeproj_test` is complete.


xcodeproj/internal/installer.template.sh
```diff
@@ -2,8 +2,51 @@
...
+  done
+  echo >&2 "${msg}"
+  exit 1
+}
+
+# Process Args
+
+while (("$#")); do
+  case "${1}" in
+    "--source")
```
brentleyjones, line 23: Is source needed as well? Shouldn't it always be from the exec root?

cgrindel: My bad. You're correct. I was looking at the runfiles directory when I put this together. The test's initial PWD is under the `com_github_buildbuddy_io_rules_xcodeproj` directory. I will remove it.

brentleyjones (approved):

Issue #414, opened 2022-05-05
Bug: Non-processable file warning for non-source code files


erikkerber:
## Describe the bug
Building in Xcode can result in warnings for files not recognized by Xcode, included in sources: 

![image](https://user-images.githubusercontent.com/75236/167028155-f1ce09c2-65b8-48f5-8600-df38950c6804.png)

## Reproduction steps

Include an "irregular" file in `srcs`, such as a `.inc`

## Expected behavior

The file is not included in the compile sources phase

## Versions
- Bazel:5.1.0
- rules_xcodeproj: <experimental `build_mode = "bazel"` >
- rules_apple: 
- rules_swift: 

## Additional context

Built using experimental Bazel build model

brentleyjones:
Seems that Bazel uses the headers/whatever in `srcs` as "private" headers, so yeah, we should filter out to known good types.

brentleyjones:
Does this also apply to BwB?

PR #464, opened 2022-05-24
Attempt to filter out non processable files

maxwellE:
Fixes: #414


maxwellE:
@brentleyjones What is the best way to get test coverage on this?

PR #472, opened 2022-05-24
Add `DisambiguatedTargets` wrapper type

brentleyjones:
For now this just has the single `targets` property. Consolidated targets will add more properties to it.

### DIFF
--- a/tools/generator/src/Generator+DisambiguateTargets.swift
+++ b/tools/generator/src/Generator+DisambiguateTargets.swift
@@ -14,11 +14,14 @@ struct DisambiguatedTarget: Equatable {
     let target: Target
 }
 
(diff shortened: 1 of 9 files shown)


tools/generator/src/Generator+DisambiguateTargets.swift
```diff
@@ -14,11 +14,14 @@ struct DisambiguatedTarget: Equatable {
     let target: Target
 }
 
+struct DisambiguatedTargets: Equatable {
+    let targets: [TargetID: DisambiguatedTarget]
+}
```
cgrindel, line 19: For ease of finding the code,  I recommend putting this in its own file.

cgrindel (approved):

PR #476, opened 2022-05-27
Fix `BuildSettingConditional` comparisons against itself

jpsim:
So that a conditional isn't both equal and less than itself at the same time.

I'm not sure if this would ever manifest itself in a real way, but it violates the `Comparable` contract.

### DIFF
--- a/tools/generator/test/BuildSettingConditionalTests.swift
+++ b/tools/generator/test/BuildSettingConditionalTests.swift
@@ -29,7 +29,24 @@ final class BuildSettingConditionalTests: XCTestCase {
         "any": .any,
     ]
 
-    func test_comparable() {
+    func test_comparable() throws {
+        // Arrange
+
+        let conditionalA = try XCTUnwrap(Self.conditionals["A"])
+        let conditionalAny = BuildSettingConditional.any
+
+        // Assert
+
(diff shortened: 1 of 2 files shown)


tools/generator/test/BuildSettingConditionalTests.swift
```diff
@@ -29,7 +29,24 @@ final class BuildSettingConditionalTests: XCTestCase {
         "any": .any,
     ]
 
-    func test_comparable() {
+    func test_comparable() throws {
+        // Arrange
+
+        let conditionalA = try XCTUnwrap(Self.conditionals["A"])
```
brentleyjones, line 35: We control the fixtures, in the same file even, so I would rather not use `XCTUnwrap`.

```suggestion
        let conditionalA = Self.conditionals["A"]!
```

brentleyjones (approved):
Thanks!

tools/generator/test/BuildSettingConditionalTests.swift
jpsim: No problem. I'd argue that as the fixtures change, seeing test failures instead of test crashes is more helpful to debug and iterate, but that's a personal workflow choice.

PR #477, opened 2022-05-27
Reorder build setting conditionals

brentleyjones:
Xcode prefers `sdk` before `arch`.

### DIFF
--- a/tools/generator/src/BuildSettingConditional.swift
+++ b/tools/generator/src/BuildSettingConditional.swift
@@ -14,12 +14,13 @@ extension BuildSettingConditional {
             return key
         }
 
+        // The order here is the order that Xcode likes them (sdk before arch)
         var components = [key]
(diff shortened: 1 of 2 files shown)


tools/generator/src/BuildSettingConditional.swift
```diff
@@ -15,12 +15,12 @@ extension BuildSettingConditional {
         }
 
         var components = [key]
-        if archConditionalAllowed(on: key) {
-            components.append("[arch=\(platform.arch)]")
-        }
         if sdkConditionalAllowed(on: key) {
```
cgrindel, line 18: If the order matters, do you want to add a comment stating that? It might help those of us who are helping out not to mess up the order later.🙂

cgrindel (changes requested):

cgrindel (approved):

PR #485, opened 2022-05-31
Remove spaces from scheme names in fixtures

brentleyjones:
This works around https://github.com/bazelbuild/bazel/issues/4327.


brentleyjones:
Needed for #484.

cgrindel (approved):
This issue is such a pain. Perhaps we should push on getting this fixed in Bazel

cgrindel:
Agreed. That is what I meant to say. 😉

PR #488, opened 2022-06-01
Add `ConsolidatedTarget`

brentleyjones:
`ConsolidatedTarget` will be used in place of `Target` throughout most of the generator (e.g. `DisambiguatedTarget` will have a `ConsolidatedTarget` instead of a `Target`). It represents a single Xcode target, but multiple configurations of the same Bazel target.

### DIFF
--- a/tools/generator/src/Generator+ConsolidateTargets.swift
+++ b/tools/generator/src/Generator+ConsolidateTargets.swift
@@ -1,0 +1,208 @@
+import OrderedCollections
+import XcodeProj
+
+/// Collects multiple Bazel targets (see `Target`) that can be represented by
+/// a single Xcode target (see `PBXNativeTarget`).
+///
+/// In a Bazel build graph there might be multiple configurations of the
+/// same label (e.g. macOS and iOS flavors of the same `swift_library`).
+/// Xcode can represent these various configurations as a single target,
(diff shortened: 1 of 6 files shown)


tools/generator/src/Generator+ConsolidateTargets.swift
```diff
@@ -0,0 +1,208 @@
...
+    let outputs: ConsolidatedTargetOutputs
+
+    /// The `Set` of `FilePath`s that each target references above the baseline.
+    ///
+    /// The baseline is all `FilePath`s that each target references. A reference
+    /// in this case is anything that the `EXCLUDED_SOURCE_FILE_NAMES` and
+    /// `INCLUDED_SOURCE_FILE_NAMES` apply to.
+    let uniqueFiles: [TargetID: Set<FilePath>]
+
+    let targets: [TargetID: Target]
```
cgrindel, line 29: Is it ever important to keep the original unsorted target list? It looks like we mostly just loop over the targets. Perhaps as a simplification, we could consolidate by having `targets` be the sorted list and removing `sortedTargets`.

brentleyjones: Yes. We need the dictionary functionally in some cases later.

Issue #285, opened 2022-04-12
Feature Request: Support building for arm64 iOS


jpsim:
First of all, thanks for creating and sharing this project. I'm looking forward to trying this to integrate with [Envoy Mobile](https://github.com/envoyproxy/envoy-mobile).

One of our main use cases for Xcode integration is to run [our iOS example apps](https://github.com/envoyproxy/envoy-mobile/tree/main/examples/swift) on physical iOS devices.

I'm not sure if this is something that rules_xcodeproj is intended to currently support and I'm using it wrong, or if this is a feature that hasn't been built yet, so I've labeled this as a feature request more than a bug.

Currently, if I run `bazel run //:xcodeproj` from `examples/ios_app` in this repo, open the project in Xcode, update the code signing section for the "Example" iOS app target to the following:

<img width="626" alt="image" src="https://user-images.githubusercontent.com/474794/163023103-837e3d77-b40f-4dfc-8e51-6fd0a677313d.png">

and attempt to build for a device (or build and run) I see the following error message in the Xcode build logs:

> rules_xcodeproj/examples/ios_app/Example/ContentView.swift:2:8: Module 'ExternalFramework' was created for incompatible target arm64-apple-ios12.0-simulator: /var/tmp/_bazel_jsimard/e1a0d3a6038bce3f1a38641ac8938150/external/examples_ios_app_external/ExternalFramework.framework/Modules/ExternalFramework.swiftmodule/arm64.swiftmodule

Possibly because I've already built and run for a local simulator. I tried running `bazel clean --expunge` and cmd-shift-k in Xcode and the error persists.

If this is a bug and iOS device support is intended to work, I'd appreciate if you could share a workaround for the issue (am I cleaning the wrong thing?).

If building for iOS devices isn't currently supported then please consider this to be a feature request for that functionality 😄.

I imagine the fact that I'm on an M1 Mac may be a contributor to this issue, since I've already built the example app for an arm64 simulator and Xcode/bazel/rules_xcodeproj is getting confused between arm64 iOS simulator and arm64 iOS device.

brentleyjones:
Running on device currently isn't supported. Thank you for detailed the feature request though. I'll see if there is a quick work around for the time being (or if I can just unblock it all-together).

jpsim:
[...]

PR #498, opened 2022-06-02
Consolidate targets

brentleyjones:
Part of #285.

| Before | After |
|-|-|
| Targets list showing multiple 'Lib' targets for each platform | Targets list showing a single 'Lib' target |

Sorry for the large PR, I don't think I can break it apart much more. I broke apart some of the changes into refactor commits, to make it slightly easier to review. The large primary commit is mainly just piping `ConsolidatedTarget` through the system, with the behavior changes being in `Generator.disambiguateTargets()` and `Generator.setTargetConfigurations()`. The behavior changes have added tests.

### DIFF
--- a/tools/generator/src/Generator+AddTargets.swift
+++ b/tools/generator/src/Generator+AddTargets.swift
@@ -472,9 +472,15 @@ private extension ConsolidatedTarget {
             return false
         }
 
-        return targets.values.allSatisfy { target in
-            return target.buildSettings["SWIFT_OBJC_INTERFACE_HEADER_NAME"]
-                != .string("")
+        return targets.values.contains { target in
+            guard let headerBuildSetting = target
+                .buildSettings["SWIFT_OBJC_INTERFACE_HEADER_NAME"]
+            else {
+                // Not setting `SWIFT_OBJC_INTERFACE_HEADER_NAME` will cause
+                // the default `ModuleName-Swift.h` to be generated
+                return true
+            }
+            return headerBuildSetting != .string("")
         }
     }
 }
@@ -491,11 +497,7 @@ private extension ConsolidatedTargetLinkerInputs {
 
 private extension ConsolidatedTargetOutputs {
     func forcedBazelCompileFiles(buildMode: BuildMode) -> Set<FilePath> {
-        guard buildMode.usesBazelModeBuildScripts else {
-            return []
-        }
-
-        if hasSwiftOutputs {
+        if buildMode.usesBazelModeBuildScripts && hasSwiftOutputs {
             return [.internal(Generator.bazelForcedSwiftCompilePath)]
         }
 
--- a/tools/generator/src/Generator+CreateProducts.swift
+++ b/tools/generator/src/Generator+CreateProducts.swift
@@ -25,13 +25,15 @@ extension Generator {
                     fileType = target.product.type.fileType
                 }
 
-                let bestTarget = target.sortedTargets.first!
+                // We can only show one product, so show the one for the default
+                // target (i.e macOS, or Simulator, or arm64)
+                let defaultTarget = target.sortedTargets.first!
 
                 // We need to fix the path for deployment location products,
                 // since we override `DEPLOYMENT_LOCATION` and
                 // `BUILT_PRODUCTS_DIR` for them
                 name = target.product.basename
-                path = "bazel-out/\(bestTarget.product.path.path)"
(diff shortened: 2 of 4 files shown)


tools/generator/src/Generator+AddTargets.swift
```diff
@@ -471,7 +466,30 @@ private extension Path {
...
-extension Outputs {
+private extension ConsolidatedTarget {
+    var generatesSwiftHeader: Bool {
+        guard isSwift else {
+            return false
+        }
+
+        return targets.values.allSatisfy { target in
+            return target.buildSettings["SWIFT_OBJC_INTERFACE_HEADER_NAME"]
+                != .string("")
```
jpsim, line 477: Can `target.buildSettings["SWIFT_OBJC_INTERFACE_HEADER_NAME"]` ever be nil? If so, this check should account for that.

tools/generator/src/Generator+CreateProducts.swift
```diff
@@ -25,11 +25,13 @@ extension Generator {
                     fileType = target.product.type.fileType
                 }
 
+                let bestTarget = target.sortedTargets.first!
```
jpsim, line 28: How is this target the best target? Depending on your answer, either renaming this variable or adding a comment could help.

tools/generator/src/Generator+SetTargetConfigurations.swift
```diff
@@ -63,6 +64,72 @@ Target "\(id)" not found in `pbxTargets`
...
+            )
+
+            // Calculate "INCLUDED_SOURCE_FILE_NAMES"
+            guard let uniqueFiles = consolidatedTarget.uniqueFiles[id] else {
+                throw PreconditionError(message: """
+Target with id "\(id)" not found in `consolidatedTarget.uniqueFiles`
+""")
+            }
+            if !uniqueFiles.isEmpty {
+                // This ket needs to not have `-` in it
```
jpsim, line 92:
```suggestion
                // This key needs to not have `-` in it
```

tools/generator/src/Generator+SetTargetConfigurations.swift
```diff
@@ -285,56 +352,88 @@ $(CONFIGURATION_BUILD_DIR)
...
     ) throws {
-        if let testHostID = target.testHost {
-            guard
-                let testHost = disambiguatedTargets.targets[testHostID]?.target
-            else {
-                throw PreconditionError(message: """
-Test host target with id "\(testHostID)" not found in `disambiguatedTargets`
+        let targets = target.targets.values
+
+        guard let aTestHostID = targets.first?.testHost else {
```
jpsim, line 363: Should this find _any_ test host, not just the one set on the first target?

E.g. `targets.lazy.compactMap(\.testHost).first`

tools/generator/src/Generator+SetTargetConfigurations.swift
```diff
@@ -460,6 +559,17 @@ extension Target {
     }
 }
 
+private extension Platform.OS {
+    var sdkRoot: String {
```
jpsim, line 563: Seems like this should be moved to a more central place at some point.

Alternatively, I think you could use XcodeProj for this: https://github.com/tuist/XcodeProj/blob/8c12577a2e07b119c85bbee118e719cc4c9f8af7/Sources/XcodeProj/Utils/BuildSettingsProvider.swift#L46

jpsim (approved):

tools/generator/src/Generator+SetTargetConfigurations.swift
brentleyjones: Yeah, I generally move them once I need to remove the `private`.

tools/generator/src/Generator+AddTargets.swift
brentleyjones: Yes, good catch.

tools/generator/src/Generator+CreateProducts.swift
cgrindel: There is a [comment here](https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/tools/generator/src/Generator%2BConsolidateTargets.swift#L242-L245) that describes the "best" concept.

brentleyjones: :+1:. I also renamed it to `defaultTarget` at this location and added another comment, just for extra clarity.

PR #501, opened 2022-06-02
Apply two different transitions to `xcodeproj`

brentleyjones:
This prevents applying the target specific transition to other attributes, such as `_generator`.


brentleyjones:
It's draft because it has a parent PR. It's just to prevent accidental merging. If I request reviews then it's G2G for review.

cgrindel (approved):

PR #506, opened 2022-06-03
Use `local_provisioning_profile` with multiplatform example

brentleyjones:
This will allow for generation of device targeting projects.


brentleyjones:
I'll have to add a select to make the `provisioning_profile` only apply when building for device, otherwise our Build CI job fails.

brentleyjones:
Okay, good now.

PR #510, opened 2022-06-05
Filter folders from rsynclist

BalestraPatrick:
A project that has a custom rule which generates a `xcassets` with a `AppIcon.appiconset` fails to build due to the fact that `rsynclist` contained the generated `xcassets` but `generated_inputs` didn't. We need to filter out the folders from the generated files list for this to work correctly.

### DIFF
--- a/tools/generator/src/Generator+CreateFilesAndGroups.swift
+++ b/tools/generator/src/Generator+CreateFilesAndGroups.swift
@@ -435,7 +435,11 @@ extension Generator {
                 useOriginalGeneratedFiles: true
             )
         }
-        let rsyncPaths = generatedFiles.map { filePath, _ in filePath.path }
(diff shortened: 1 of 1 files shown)


brentleyjones (approved):

tools/generator/src/Generator+CreateFilesAndGroups.swift
```diff
@@ -435,7 +435,11 @@ extension Generator {
                 useOriginalGeneratedFiles: true
             )
         }
-        let rsyncPaths = generatedFiles.map { filePath, _ in filePath.path }
+        let rsyncPaths = generatedFiles
+            .filter { filePath, environment in
```
brentleyjones, line 439: environment is unused?

PR #512, opened 2022-06-07
Fix `ENABLE_PREVIEWS` setting in `BazelDependencies`

brentleyjones:
We need to set `SWIFT_OPTIMIZATION_LEVEL = -Onone`, otherwise `ENABLE_PREVIEWS` is always `NO`, even when building for previews.


tools/generator/src/Generator+CreateProject.swift
```diff
@@ -50,6 +50,8 @@ $(INDEXING_DEPLOYMENT_LOCATION__)
             "INTERNAL_DIR": """
 $(PROJECT_FILE_PATH)/\(filePathResolver.internalDirectoryName)
 """,
+            // Needed as the default otherwise `ENABLE_PREIVEWS` isn't set
```
BalestraPatrick, line 53: Nit: small typo in the docs

BalestraPatrick (approved):

PR #517, opened 2022-06-08
[Bug] Unable to conume external dep w/XCTest

maxwellE:
I cannot seem to consume an external swift only library that uses XCTest

When running `ExampleTests.__internal__.__test_bundle` I see an error "No such module 'XCTest'"
coming from my external dependencies code


brentleyjones:
And if that does fix it, I think we should fix `rules_swift` to need `test_only` before it adds the search paths for dev only frameworks.

PR #519, opened 2022-06-08
Use the correct version of swiftc for BwB SwiftUI Previews

brentleyjones:
This is the same logic that we use in the ld stub, for the same reason: the `PATH` and `DEVELOPER_DIR` isn't set correctly for us, so we have to infer what `DEVELOPER_DIR` is. We also force the use of the default toolchain, which means this currently doesn't work correctly with custom toolchains.


BalestraPatrick (approved):
@brentleyjones What's exactly the reason why it's not set correctly in this mode?

PR #437, opened 2022-05-12
Add automatic_signing_team_id for automatic code signing

erikkerber:
**NOTE:** Early WIP and PoC

Resolves #436

PR #523, opened 2022-06-09
Add experimental rule to specify Xcode Managed provisioning profiles

### LINKED ISSUE #436: Feature Request: Support "Automatic" Code Signing
Xcode projects by default will set `CODE_SIGN_STYLE=Automatic`. When a developer then tries to debug to a device, Xcode will instruct them to select a `DEVELOPMENT_TEAM` that automatic code signing will use.

Automatic signing can be beneficial for new projects that are getting setup with Bazel and don't have plumbing setup to manage certificates and profiles, or smaller teams that don't care to have fine control over their signing process.

This will affect both BWX and BWB (more on that below)

## Design

Optionally providing an `automatic_signing_team_id` to the `xcodeproj` rule:

```
xcodeproj(
    name = "Acme",
    targets = ["AcmeApp"],
    automatic_signing_development_team = "ABCDEFG",
)
```

If set, the generator will set `DEVELOPMENT_TEAM=ABCDEFG`, and leave unset if not provided.

### BWX Expectations

Xcode will:

- Register the device if not registered
- Regenerate the managed provisioning profile for each bundle ID, if necessary adding the logged in Apple ID's certificate
- When the "Run" action is performed, it will sign the app using the managed profile

### BWB Expectations

Xcode will:
- Register the device if not registered
- Regenerate the managed provisioning profile for each bundle ID, if necessary adding the logged in Apple ID's certificate

With BwX, It is up to the developer to reference that profile using a repository rule such as [local_provisioning_profiles](https://github.com/bazelbuild/rules_apple/pull/1190#discussion_r870663831)

brentleyjones:
Resolves #436.

Closes #437.

### DIFF
--- a/xcodeproj/internal/xcode_provisioning_profile.bzl
+++ b/xcodeproj/internal/xcode_provisioning_profile.bzl
@@ -1,0 +1,61 @@
+"""Implementation of the `xcode_provisioning_profile` rule."""
+
+load("@bazel_skylib//rules:common_settings.bzl", "BuildSettingInfo")
+load(":provisioning_profiles.bzl", "provisioning_profiles")
+
+# Utility
+
+def _requires_team_id(ctx):
+    """Determines whether a team id is needed.
+
+    The team id can come from `ctx.attr.team_id` or
+    `ctx.attr.provisioning_file[AppleProvisioningProfileInfo].team_id`.
+
+    Args:
+        ctx: The aspect context.
+
+    Returns:
+        `True` if a team ide is needed, `False` otherwise.
+    """
+    return ctx.attr._build_mode[BuildSettingInfo].value == "xcode"
+
+# API
+
+def _xcode_provisioning_profile_impl(ctx):
+    target = ctx.attr.provisioning_profile
+
+    info = provisioning_profiles.create_profileinfo(
+        target = target,
+        is_xcode_managed = ctx.attr.managed_by_xcode,
+        profile_name = ctx.attr.profile_name,
+        team_id = ctx.attr.team_id,
+    )
+    if not info.team_id and _requires_team_id(ctx):
+        fail("""\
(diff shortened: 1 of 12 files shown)


brentleyjones:
I'm going to add a little bit more to this to allow non-`local_provisioning_profile` based profiles to work in BwX.

brentleyjones:
Good to go now.

BwB doesn't actually use this values that this provides to Xcode, besides for letting Xcode produce provisioning profiles for Automatic Signing. So if a project doesn't use Automatic Signing, and uses BwB, it doesn't need this new rule.

BwX though needs to use `local_provisioning_profile` and/or this rule for Manual Signing, in order to let Xcode know the required values it needs to code sign. And the profile will also have to exist at `~Library/MobileDevice/Provisioning Profiles`. For Automatic Signing this rule will need to be used (probably with `local_provisioning_profile`).

I'll add stuff to the FAQ about this after this is merged.

BalestraPatrick (approved):
Tested this both in BwX and BwB modes and it worked great!

xcodeproj/internal/xcode_provisioning_profile.bzl
```diff
@@ -0,0 +1,61 @@
...
+
+    return [
+        DefaultInfo(files = target.files),
+        info,
+    ]
+
+xcode_provisioning_profile = rule(
+    implementation = _xcode_provisioning_profile_impl,
+    attrs = {
+        "managed_by_xcode": attr.bool(
```
BalestraPatrick, line 47: I think it would be interesting to document what this attribute means exactly. If set to `False`, what would that affect?

brentleyjones: Yeah, all these rules and attributes need doc 😬.

erikkerber:
Tested as working on our build as well.

Thanks for the legwork on this!

PR #534, opened 2022-06-11
Only build `BazelDependencies` once in a build

brentleyjones:
This applies a different fix that 55142f56245dd411c560862306e77379ac71bbd8 tried to accomplish. Before adding consolidated targets the previous fix worked fine, though it did have potential cache issues with the PATH being different for different platforms.

With consolidated targets there were builds where multiple `BazelDependencies` would be in a single build, particularly during indexing.


BalestraPatrick (approved):

brentleyjones:
Seems this causes issues for non multi-platform builds. Looking for a proper fix.

PR #541, opened 2022-06-13
Properly fix reinstalls of Built with Bazel apps to device

brentleyjones:
This reverts commit 3eb7700934a479344642a7b5f8597c8ba839bdde.

By setting the produced `Info.plist` as an input to the `Copy Bazel Outputs` script, it forces Xcode to process the file before Bazel copies over it.

Thanks @thii for the tip!


brentleyjones:
@BalestraPatrick @jpsim Can onw of you confirm this still fixes the device install issue y'all were seeing? In my testing it did.

jpsim (approved):
👍  Build-install-run cycles continue to work with this change, as they did with the previous workaround.

BalestraPatrick:
Can confirm it fixes reinstalls in my projects as well! 🎉

PR #543, opened 2022-06-13
Work towards removing the need for `InputFileAttributesInfo.excluded`

brentleyjones:
This change causes imported frameworks and libraries to be included in `extraFiles`, even though they are already included in `linkerInputs`. This is already the case for modulemaps as well, they are already included in `modulemaps`. In a future change we can probably prune things from `extraFiles` that we have collected elsewhere.

A refactor of how we handle resource file collection will allow us to fully remove `InputFileAttributesInfo.excluded`, and the `_SUSPECT_GENERATED_EXTENSIONS` check as well.


examples/ios_app/test/fixtures/bwb.xcodeproj/project.pbxproj
```diff
@@ -347,6 +348,7 @@
 				48F0B3D39D176B90AE9F2C8A /* Answer.swift.stencil */,
 				1104D414730D2DC8DC9DD9E9 /* BUILD */,
 				55E65A7323E456CDEBA791E6 /* Greeting.swift.stencil */,
+				6602B63E152B03A28F98567A /* merge.sh */,
```
brentleyjones, line 351: This is a nice side effect 😄.

BalestraPatrick (approved):

PR #546, opened 2022-06-14
Use attribute `entitlements` instead of Bazel final version

brentleyjones:
The Bazel final version has a couple issues:
- It "depends" on the provisioning profile, which can result in circular generation issues
- Whatever changes are made to it, Xcode also makes
- Most people generally don't have generated entitlements, so this removes another generated file


BalestraPatrick (approved):
We have generated entitlements but we don't need them in the project. I tested this and it works.

PR #552, opened 2022-06-14
Fix non-`--define=apple.experimental.tree_artifact_outputs=1` syncing bug

brentleyjones:
Bazel built zips have zeroed out timestamps. So if any unarchived items had the same modified size as before a change, rsync wouldn't copy them over. We now set the timestamps to be the time we unzip.


jpsim (approved):
Confirmed rebuilding for device now work reliably. Thanks!

PR #582, opened 2022-06-24
Add `XcodeProjAutomaticTargetProcessingInfo.should_generate_target`

brentleyjones:
This moves the logic of determining if a target should become an Xcode target out of a hard coded function and into the provider.

It also slightly changes the behavior only generating targets by default for the rules we provide default information for. This causes "unsupported" rules like `swift_proto_library` to no longer become incomplete Xcode targets. A later change will adjust the generator to handle these use cases better.

### DIFF
--- a/xcodeproj/internal/default_input_file_attributes_aspect.bzl
+++ b/xcodeproj/internal/default_input_file_attributes_aspect.bzl
@@ -43,6 +43,9 @@ def _default_input_file_attributes_aspect_impl(target, ctx):
 
     this_target_type = _get_target_type(target = target)
 
+    # Targets that don't produce outputs can't be Xcode targets
+    should_generate_target = target.files != depset()
+
     if CcInfo in target:
         srcs = ("srcs")
     else:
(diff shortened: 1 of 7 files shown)


xcodeproj/internal/default_input_file_attributes_aspect.bzl
```diff
@@ -100,11 +99,16 @@ def _default_input_file_attributes_aspect_impl(target, ctx):
         infoplists = ("infoplists")
     elif AppleFrameworkImportInfo in target:
         xcode_targets = {"deps": [target_type.compile]}
+        should_generate_target = False
```
jpsim, line 102: Doesn't this get overridden on line 107?

```
should_generate_target = executable and not executable.is_source
```

brentleyjones: 107 is in an `else` statement, so no.

jpsim: Ah yes, gotta love semantic white space.

jpsim (approved):

PR #586, opened 2022-06-26
Use `copts_for_current_compilation_mode`

maxwellE:
Uses `copts_for_current_compilation_mode` and `ctx.fragments.objc.copts`
to set compiler flags
for objc and swift compilation

### DIFF
--- a/xcodeproj/internal/opts.bzl
+++ b/xcodeproj/internal/opts.bzl
@@ -109,6 +109,8 @@ def _get_unprocessed_compiler_opts(*, ctx, target):
 
     user_copts = []
     user_swiftcopts = []
+    objc = ctx.fragments.objc
+    objc_copts = []
     if SwiftInfo in target:
         # Rule level swiftcopts are included in action.argv below
         user_swiftcopts = getattr(ctx.rule.attr, "copts", [])
@@ -117,6 +119,7 @@ def _get_unprocessed_compiler_opts(*, ctx, target):
             values = user_swiftcopts,
             attribute_name = "copts",
         )
+        objc_copts = _objc_fragment_copts(objc)
     elif CcInfo in target:
         user_copts = getattr(ctx.rule.attr, "copts", [])
         user_copts = _expand_make_variables(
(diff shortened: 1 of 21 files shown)


xcodeproj/internal/opts.bzl
```diff
@@ -134,7 +138,7 @@ def _get_unprocessed_compiler_opts(*, ctx, target):
     cpp = ctx.fragments.cpp
 
     return (
-        base_copts + cpp.copts + cpp.conlyopts + user_copts,
+        base_copts + cpp.copts + cpp.conlyopts + user_copts + objc_copts,
```
brentleyjones, line 141: This needs to be before user ones, to make sure that user ones can override it. I'm not sure where in the order it is, but can you compare this to what Bazel produces? We might need to remove some of the existing ones as well depending on what you see.

Also, for the swift setting of the flags, I think we need to set a lot less. Looking at rules_swift and what they set, it's just a small subset with some filtered as well. We should mimic that.

BalestraPatrick:
This is great! From my first tests this fixes some issues we've seen in our projects as well.

Judging from the fixtures, I see some duplicate flags (`"-fstack-protector"` and `"-g"` I've noticed). Should we filter duplicates while preserving order?

Moreover, I see that `rules_swift` filters our all `"-g"`. We might need to do the same: https://github.com/bazelbuild/rules_swift/blob/ac77c2da9a4c24cf0696f56a7941aa0def532da1/swift/internal/xcode_swift_toolchain.bzl#L64-L107

xcodeproj/internal/opts.bzl
```diff
@@ -134,12 +138,42 @@ def _get_unprocessed_compiler_opts(*, ctx, target):
     cpp = ctx.fragments.cpp
 
     return (
-        base_copts + cpp.copts + cpp.conlyopts + user_copts,
+        _flattened_unique_copts(base_copts, cpp.copts, cpp.conlyopts, objc_copts, user_copts),
```
brentleyjones, line 141: You can use the `uniq` function instead. It keeps the order and is more performant. M

brentleyjones (approved):
I think there are still some improvements we can make here, but this is a step in the right direction :+1:.

Issue #573, opened 2022-06-22
Feature Request: Provide a means for clients to define custom Xcode schemes


cgrindel:
Today, `rules_xcodeproj` generates an Xcode scheme for each target. It would be useful for clients to be able to define and configure the Xcode schemes as one can do in the Xcode user interface. 

## Suggested Features
- Specify a custom name
- Specify the targets to build in each scheme. 
- Identify the test targets in the build list and use them as the testables.
- Specify the launch target.

## Tasks

- [x] #592
- [x] #612
- [x] #626 
- [x] #630
- [x] #637 
- [x] #646 
- [x] #662 
- [x] #664 
- [x] #656 
- [x] #733
- [x] Generate custom Xcode schemes.
  - [x] #748 
  - [x] #801
  - [x] #807 
  - [x] #803
  - [x] #808 
  - [x] #809
  - [x] #844
  - [x] Remove or adjust top-level target logic: [top-level inference](https://github.com/buildbuddy-io/rules_xcodeproj/issues/573#issuecomment-1197341285), [isTopLevel logic](https://github.com/buildbuddy-io/rules_xcodeproj/issues/573#issuecomment-1197346855)
    - [x] #846
    - [x] #855 
    - [x] #859
    - [x] #865
    - [x] #867 
  - [x] Update "best" host logic selection in `XCSchemeInfo.TargetInfo` when the best host is not a top-level target. (see GH573 comment)
    - [x] #889 
    - [x] #890 
  - [x] Update `buildFor` logic
    - [x] #896 
    - [x] #899
    - [x] #901
    - [x] #905
    - [x] #909 
- [x] Migrate `Hashable` implementation for `XCScheme.BuidableReference` to `tuist/XcodeProj`.
  - [x] tuist/XcodeProj#712
  - [x] #912 
- [x] #913 

## Deferred Items
- Be sure that typos in any of the labels specified in the schemes are checked for validity. [Slack thread](https://ios-dev-at-scale.slack.com/archives/C03BS1YT8JE/p1656675962659949)
  - It turns out that `Label()` does not validate that a label is valid (i.e., exists).
[...]

PR #592, opened 2022-06-26
Design proposal for custom Xcode scheme support

cgrindel:
Related to #573.


brentleyjones:
Btw, it seems you have some undesired fixture changes?

jpsim:
Thanks for drafting this! I'd love if this doc could touch on the strategy for schemes for transitive dependencies.

A big reason we need control over the schemes created is to reduce the "pollution" caused by having a large dependency graph since today, not only do all `targets` passed to `xcodeproj` get a scheme created, but so do all of their transitive dependencies, which can lead to hundreds or even thousands of schemes being created. For Envoy Mobile, we have 2,506 schemes created. This significantly slows down Xcode and adds a lot of clutter.

How do you think schemes of transitive dependencies should be handled? Should there be a `generate_schemes_for_deps = False` flag on either `xcodeproj` or `xcode_scheme`?

cgrindel:
Having gathered some feedback from @brentleyjones and others, I updated the proposal. The new version of the proposal takes a different approach to scheme definitions. Please give it a look and provide any feedback.

@jpsim Good news. The new proposal includes a `scheme_autogeneration_mode` on `xcodeproj`. It allows you to customize if autogenerated schemes should be included.

doc/design/custom_xcode_schemes.md
```diff
@@ -0,0 +1,467 @@
...
+    schemes = _SCHEMES,
+)
+```
+
+The above example adds a second scheme called `My Application`. It is configured with two test targets
+and a single launch target. All three targets are implicitly included in the build list.
+
+### Specifying Launch Arguments and Environment Variables
+
+There are times when it is desirabe to customize the launch action with arguments and environment
```
jpsim, line 141: Another launch configuration option that I use often is to specify a custom working directory. Is this something we could support configuring here?

brentleyjones: I think so, yes.

doc/design/custom_xcode_schemes.md
```diff
@@ -0,0 +1,467 @@
...
+# Assumptions
+#   //Sources/Foo:Foo - swift_library
+#   //Sources/FooTests:FooTests = ios_unit_test
+
+xcodeproj(
+    name = "generate_xcodeproj",
+    project_name = "Command Line",
+    tags = ["manual"],
+    targets = [
+        "//Sources/Foo",
```
brentleyjones, line 42: Unless we want to get into the intricacies of how this will actually create _3 targets_ and _3 schemes_ let's not include the library in `targets`.

```suggestion
#   //Sources/FooTests:FooTests = ios_unit_test
#   //Sources/Foo:Foo - swift_library, a dependency of //Sources/FooTests:FooTests

xcodeproj(
    name = "generate_xcodeproj",
    project_name = "Command Line",
    tags = ["manual"],
    targets = [
```

doc/design/custom_xcode_schemes.md
```diff
@@ -0,0 +1,467 @@
...
+This example is the same as the previous one except arguments and environment variables are
+defined for the launch action.
+
+## Introduction of Scheme Autogeneration Mode
+
+With this proposal, it is now possible to control how Xcode scheme autogeneration works using
+the `scheme_autogeneration_mode` attribute on `_xcodeproj`. 
+
+__Values__
+- `auto`: If no custom schemes are provided, an Xcode scheme will be created for every buildable
```
jpsim, line 200: If there are no custom schemes provided and the mode is `auto`, why wouldn't schemes be created just for the list of `targets` explicitly specified? To get schemes for all buildable transitive deps you'd need to specify `all`.

doc/design/custom_xcode_schemes.md
```diff
@@ -0,0 +1,467 @@
...
+xcodeproj(
+    name = "generate_xcodeproj",
+    project_name = "Command Line",
+    tags = ["manual"],
+    schemes = _SCHEMES,
+)
+```
+
+The above example adds a second scheme called `My Application`. It is configured with two test targets
+and a single launch target. All three targets are implicitly included in the build list.
```
brentleyjones, line 137: >All three targets are implicitly included in the build list.

Is `//Tests/AppUITests` set to build for `Run` or only `Test` (default behavior would only be `Test`). And are we able to force it to build during `Run` by listing it explicitly in `build_action`?

doc/design/custom_xcode_schemes.md
brentleyjones, line 141:
```suggestion
There are times when it is desirable to customize the launch action with arguments and environment
```

doc/design/custom_xcode_schemes.md
```diff
@@ -0,0 +1,467 @@
...
+
+## Introduction of Scheme Autogeneration Mode
+
+With this proposal, it is now possible to control how Xcode scheme autogeneration works using
+the `scheme_autogeneration_mode` attribute on `_xcodeproj`. 
+
+__Values__
+- `auto`: If no custom schemes are provided, an Xcode scheme will be created for every buildable
+  target. If custom schemes are provided, no autogenerated schemes will be created.
+- `none`: No custom schemes are generated.
```
brentleyjones, line 202:
```suggestion
- `none`: No schemes are automatically generated.
```

doc/design/custom_xcode_schemes.md
```diff
@@ -0,0 +1,467 @@
...
+- `all`: Generate a scheme for every buildable target even if custom schemes are provided.
```
brentleyjones, line 203:
```suggestion
- `all`: A scheme is generated for every buildable target even if custom schemes are provided.
```

doc/design/custom_xcode_schemes.md
```diff
@@ -0,0 +1,467 @@
...
+
+The default value for `scheme_autogeneration_mode` is `auto`.
+
+## Implementation Changes
+
+### `xcode_schemes` Module
+
+```python
+
```
brentleyjones, line 212:
```suggestion
```

doc/design/custom_xcode_schemes.md
```diff
@@ -0,0 +1,467 @@
...
+    as_json = True, 
+    build_action = None, 
+    test_action = None, 
+    launch_action = None):
+    """Returns a `struct` or JSON `string` representing an Xcode scheme.
+
+    Args:
+        name: The user-visible name for the scheme as a `string`.
+        as_json: Optional. A `bool` indicating whether the resulting struct 
+            should be returned as JSON.
```
brentleyjones, line 224: What's the use case for this?

doc/design/custom_xcode_schemes.md
cgrindel: Sure. Do you anticipate hardcoding an absolute path in the BUILD file or would you be using `$(location)`?

doc/design/custom_xcode_schemes.md
cgrindel: Sorry. My wording is imprecise. The intention is for `auto` to be work like `all` when no custom schemes exist. I was trying to preserve the current behavior.

doc/design/custom_xcode_schemes.md
```diff
@@ -0,0 +1,467 @@
...
+        **kwargs: Additional arguments to pass to `xcodeproj_rule`.
+    """
+    pass
+```
+
+In addition to what the `xcodeproj` macro does today, if a value is provided for the `schemes`
+parameter, the macro will collect all of the targets listed in the test and launch actions (not the
+build action) and add them to the overall targets list. We only collect targets from the test action
+and launch action because the target list for `xcodeproj` should only contain top-level or
+leaf-nodes (i.e., not depeneded upon by other targets in the Xcode project).
```
brentleyjones, line 329: > We only collect targets from the test action and launch action because the target list for `xcodeproj` should only contain top-level or leaf-nodes (i.e., not depeneded upon by other targets in the Xcode project).

A library _could_ be a top-level target though, and it might not have a test. I think it's fine to force them to re-list that target in `targets`, but it's probably worth mentioning.

Also, we need to make sure to allow duplications (i.e. someone listing out all top level targets in `targets` and specifying custom schemes for each).

doc/design/custom_xcode_schemes.md
jpsim: What would `$(location)` resolve to? When I'm building CLIs I often want to run them in the project's root directory or an absolute directory.

doc/design/custom_xcode_schemes.md
jpsim: That part was actually clear to me, I'm suggesting an alternative which would allow reducing the scheme pollution by default when using `targets` instead of `schemes` while still having a way to generate schemes for all transitive deps using `all`.

brentleyjones: I could see that being another option, where it's `top_level_only`.

jpsim: At Lyft, our generated Xcode project can be configured to go up and/or down N levels following the dependency graph for either transitive or reverse dependencies.

brentleyjones: But that's Focused Projects, not scheme control, right? Focused Projects will only have schemes for targets that are in-focus, so the behavior would be the same then.

So if you had Focused Projects, would you use need this different generation mode?

jpsim: I was under the impression that a "focused project" meant that there would only be schemes only for a subset of the build graph. Are those not the same thing?

brentleyjones: Focused Project means that there will only be targets and their associated files in Xcode. (It's not just schemes that slow down Xcode, the targets themselves do as well.) Schemes can only be for targets that are in Xcode, so if you have a Focused Project it will have fewer schemes. But you can have a non-Focused Project (maybe you want all the targets and files) and still want fewer schemes, which is where I thought `top_level_only` might be useful.

jpsim: ok, so if I'm understanding correctly, limiting the created schemes to a subset of the build graph doesn't mean the project is focused.

Only limiting the source files or targets to a subset of the build graph is referred to as a focused project.

brentleyjones: Correct. We will need a glossary soon 😅.

BalestraPatrick: I really like the addition of `top_level_only`. In our project we have schemes for the top-level buildables (such as apps and extensions mostly). These would cover the majority of the use-cases. Via the separate API it would then be possible to specify additional schemes for other library and test targets.

doc/design/custom_xcode_schemes.md
cgrindel: Digging into `tuist/XcodeProj`, I see [useCustomWorkingDirectory](https://github.com/tuist/XcodeProj/blob/3a93b47a34860a4d7dbcd9cc0ae8e9543c179c61/Sources/XcodeProj/Scheme/XCScheme%2BLaunchAction.swift#L45) that specifies that a custom working directory can be used. However, I don't see [customWorkingDirectory in the data model](https://github.com/tuist/XcodeProj/search?q=customWorkingDirectory). I confirmed that is the value that is written when you specify a custom working directory in Xcode. 

It looks like we will need to put up a PR to tuist to add it.

doc/design/custom_xcode_schemes.md
cgrindel: So, to close this out, do we want the `top_level_only` value? Will you use it day one?

doc/design/custom_xcode_schemes.md
cgrindel: Good question. For reference, here is the [BuildFor enum](https://github.com/tuist/XcodeProj/blob/3a93b47a34860a4d7dbcd9cc0ae8e9543c179c61/Sources/XcodeProj/Scheme/XCScheme%2BBuildAction.swift#L8-L13).

I was thinking that we would just set the `buildFor` to `default`.  For autogenerated schemes, we set the following list:
```swift
                buildFor: [
                    .running,
                    .testing,
                    .profiling,
                    .archiving,
                    .analyzing
                ]
```

I think that we can be as smart of dumb about this as we want. If you think that it is important to be selective, we can add it. I don't have a good sense for whether folks customize this for their real projects.

doc/design/custom_xcode_schemes.md
brentleyjones: We want it, and it seems @BalestraPatrick would use it regardless of Focused Projects support.

doc/design/custom_xcode_schemes.md
cgrindel: This provides some syntactic sugar for the client. In the implementation, I was planning for all of the functions in `xcode_schemes` to return `struct` values. However, when the client uses `xcode_schemes.scheme`, the attribute on `_xcodeproj` rule will accept a `list` of JSON `string` values. I didn't the client to write something ugly like `xcode_schemes.scheme_json`. So, I added the `as_json` parameter. The client won't touch it, but our implementation and tests would.

As I was writing the above, it occurred to me that we can go at this a slightly different way. Today, we wrap calls to `_xcodeproj` in the macro `xcodeproj`. We could have `xcode_schemes.scheme` return a `struct` like the other functions. The `xcodeproj` macro can then serialize them to JSON before passing them to `_xcodeproj`. This approach has another nice feature. It skips a serialization/deserialization roundtrip. 

What do you think?

doc/design/custom_xcode_schemes.md
brentleyjones: I believe Xcode defaults to having test targets `buildFor: [.testing]`.

I say anything we do this:
- Anything `build_action` gets `.testing` in `buildFor`
- Anything in `launch_action` or `build_action` gets `[.running, .profiling, .archiving, .analyzing]` in `buildFor`
- Thus being in both gets all of them

doc/design/custom_xcode_schemes.md
brentleyjones: >The xcodeproj macro can then serialize them to JSON before passing them to _xcodeproj

That is what I was thinking as well. Let's go that route.

doc/design/custom_xcode_schemes.md
jpsim: I'd use `top_level_only` even without custom schemes if it's available before because I do see a perf hit in the scheme selector that is improved when trimming the schemes.

doc/design/custom_xcode_schemes.md
cgrindel: Did you mean `test_action` for the first bullet?

cgrindel: The rest sounds reasonable to me.

doc/design/custom_xcode_schemes.md
cgrindel: I will add it.

doc/design/custom_xcode_schemes.md
```diff
@@ -0,0 +1,501 @@
...
+    ]
+  },
+  "launch_action": {
+    "target": "//Sources/App"
+  }
+}
+```
+
+> TODO: Is selecting targets from the test actions and launch actions sufficient for only incuding
+> top-level targets in the overall targets list?
```
brentleyjones, line 481: What do you mean by this?

doc/design/custom_xcode_schemes.md
cgrindel: I added a section enumerating the build for logic.

doc/design/custom_xcode_schemes.md
```diff
@@ -495,6 +496,17 @@ To simplify the syntax for specifying targets for schemes, we have opted to allo
 specified by their Bazel target only. The logic inside `rules_xcodeproj` will select the correct
 configuration variant based upon the leaf nodes that are included in the scheme.
 
+## Build for Configuration Logic
+
+Xcode schemes support configuraiton that dictates when a target is built. In the `tuist/XcodeProj`
```
brentleyjones, line 501:
```suggestion
Xcode schemes support configuration that dictates when a target is built. In the `tuist/XcodeProj`
```

doc/design/custom_xcode_schemes.md
```diff
@@ -495,6 +496,17 @@ To simplify the syntax for specifying targets for schemes, we have opted to allo
...
+data model, this is modeled under
+[`XCScheme.BuildAction.Entry.BuildFor`](https://github.com/tuist/XcodeProj/blob/3a93b47a34860a4d7dbcd9cc0ae8e9543c179c61/Sources/XcodeProj/Scheme/XCScheme%2BBuildAction.swift#L8-L13).
+The following will be the logic used to set the `BuildFor` for targets listed in custom schemes.
```
brentleyjones, line 504:
```suggestion

The following will be the logic used to set the `BuildFor` for targets listed in custom schemes:
```

doc/design/custom_xcode_schemes.md
cgrindel: In short, are you OK with the technique of populating the the `targets` list for `_xcodeproj` rule using the targets listed under `test_action` and `launch_action`?

brentleyjones: Yes.

doc/design/custom_xcode_schemes.md
cgrindel: I have updated the doc with `working_directory` support for `launch_action`.

brentleyjones (approved):

PR #595, opened 2022-06-27
Apply compiler options the same as Bazel

brentleyjones:
Followup to https://github.com/buildbuddy-io/rules_xcodeproj/pull/586. Part of https://github.com/buildbuddy-io/rules_xcodeproj/issues/415.

- No longer set C related flags for Swift targets, as they aren't used and are instead passed to `OTHER_SWIFT_FLAGS` with `-Xcc` prefixes
- Set Swift PCM flags exactly as rules_swift does
- Correctly retrieve copts for `objc_library` targets by checking for `ObjcProvider` and changing the action that is queried based on that (this check will have to change once `ObjcProvider` is removed)
- Only set objc based flags for objc based actions
- Stop uniquing flags, as Bazel doesn't either

### DIFF
--- a/examples/ios_app/test/fixtures/bwb.xcodeproj/project.pbxproj
+++ b/examples/ios_app/test/fixtures/bwb.xcodeproj/project.pbxproj
@@ -1179,7 +1179,8 @@
 				BAZEL_TARGET_ID = "//ExampleObjcTests:ExampleObjcTests.__internal__.__test_bundle applebin_ios-ios_x86_64-dbg-ST-a0d0e3b8f217";
 				BUILT_PRODUCTS_DIR = "$(CONFIGURATION_BUILD_DIR)";
 				BUNDLE_LOADER = "$(TEST_HOST)";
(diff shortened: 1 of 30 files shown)


BalestraPatrick (approved):
Tested locally in a couple of projects and didn't find regressions. LGTM!

examples/ios_app/test/fixtures/bwb.xcodeproj/project.pbxproj
```diff
@@ -1201,16 +1201,33 @@
...
+					"-Wunused-function",
+					"-Wunused-variable",
+					"-fno-autolink",
+					"-fexceptions",
+					"-fasm-blocks",
+					"-fobjc-abi-version=2",
+					"-fobjc-legacy-dispatch",
+					"-fstack-protector",
 					"-fstack-protector-all",
+					"-g",
```
maxwellE, line 1230: Should this be set?

PR #599, opened 2022-06-27
Stop assuming all `AppleBundleInfo` rules provide `AppleResourceInfo`

brentleyjones:
Fixes issues with `ios_static_framework` and the like.


jpsim:
I guess a test is tricky to craft here since this would need a static framework?

jpsim (approved):

brentleyjones:
It's tricky because we don't fully support `ios_static_framework` yet.

PR #601, opened 2022-06-28
Adjust `hdrs` collection to exclusively use `compilation_context`

### LINKED ISSUE #451: Bug: iOS app with `apple_static_framework_import` dependency fails to generate
## Describe the bug

When adding an iOS app target with an `apple_static_framework_import` dependency as a target to the Xcode project, project generation fails with the following error:

> Framework with file path "FilePath(type: generator.FilePath.PathType.generated, path: ios-sim_arm64-min12.0-applebin_ios-ios_sim_arm64-dbg-ST-aa7db8051c58/bin/Envoy.framework, isFolder: false)" not found in `files`

Defined here: https://github.com/buildbuddy-io/rules_xcodeproj/blob/0.3.0/tools/generator/src/Generator%2BAddTargets.swift#L371-L373

Steps to reproduce:

```
git clone https://github.com/envoyproxy/envoy-mobile.git
cd envoy-mobile
git checkout jp-add-objc-app-to-xcodeproj
./bazelw run //:xcodeproj
```

## Expected behavior

Project should generate successfully.

brentleyjones:
Fixes #451.

This fixes issues where generated files were used with rules like `apple_static_framework_import`.

If a `apple_static_framework_import` doesn't have any headers, but the archive is generated, we will still run into an issue. To resolve that we need to adjust `linker_input_files` to add `static_framework_file`, `library`, and the like to inputs. That's for a later change (probably merging the `input_files` and `linker_inputs` into one module`).

### DIFF
--- a/examples/ios_app/test/fixtures/bwb_spec.json
+++ b/examples/ios_app/test/fixtures/bwb_spec.json
@@ -1168,8 +1168,8 @@
             "info_plist": null,
             "inputs": {
                 "hdrs": [
-                    "Utils/Foo.h",
-                    "Utils/Utils.h"
+                    "Utils/Utils.h",
+                    "Utils/Foo.h"
                 ],
                 "srcs": [
                     "Utils/Foo.m"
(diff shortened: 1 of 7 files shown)


jpsim (approved):
Nice.

Issue #237, opened 2022-04-04
spec.json files should be as minimal as possible


brentleyjones:
The PROJECT_spec.json files are only used to communicate data between the `xcodeproj_aspect` and `generator`. They don't need to be easy to read (though readable in some way is still a goal for debugging). At a minimum we should use custom `Decodable` implementations to reduce the length of keys, which we already do for the `FilePath` DTO. Possibly we can go futher and use something like protobufs.

brentleyjones:
Closing this as done for now. The custom Xcode schemes could be minimized, but I think it could use some more work on the format in general. Also the idea of using protobuf could still be explored, though we have to see if it's slower to generate over our decently optimized current flow.

PR #615, opened 2022-06-30
Expand `_normalized_file_path` to more folder types

brentleyjones:
Part of #237.

Also rename it to `normalized_file_path` and move it to `files.bzl`. This will allow its use outside of `input_files.bzl`.

`.xcdatamodel` isn't included because we need to access the underlying files, similar to localized files.

### DIFF
--- a/xcodeproj/internal/files.bzl
+++ b/xcodeproj/internal/files.bzl
@@ -62,6 +62,37 @@ def parsed_file_path(path):
     else:
         return project_file_path(path)
 
+_FOLDER_TYPE_EXTENSIONS = {
+    ".bundle": None,
+    ".docc": None,
+    ".framework": None,
+    ".scnassets": None,
+    ".xcassets": None,
+}
+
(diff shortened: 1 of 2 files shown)


xcodeproj/internal/files.bzl
```diff
@@ -62,6 +62,37 @@ def parsed_file_path(path):
     else:
         return project_file_path(path)
 
+_FOLDER_TYPE_EXTENSIONS = {
+    ".bundle": None,
+    ".docc": None,
+    ".framework": None,
```
jpsim, line 68: `.xcframework` too?

brentleyjones: No, since the way that bazel handles xcframeworks is to process itself and expose the underlaying frameworks. So in the generated project we need to treat xcframeworks as simple folders. I have a PR soon that will have an example usecase of importing xcframeworks.

jpsim (approved):

PR #637, opened 2022-07-02
Update `xcodeproj` macro to process Xcode schemes

cgrindel:
Related to #573.

- Add `workspace_name_resolvers` module to expose `repository_name()` and `package_name()`. 
  - This was added so that a different implementation could be injected into unit tests.
- Add `bazel_labels` module to normalize label values provided to `xcodeproj` and `xcode_schemes`. 
  - The `make_bazel_labels` function allows for the injection of a custom `workspace_name_resolvers` module. 
  - In a later PR, we will add code to validate that the provided labels exist.
- Update `xcode_schemes` module to include `make_xcode_schemes`. 
  - This allows a custom `bazel_labels` module to be injected.
- Update `xcodeproj` macro to accept and process `schemes`. 
  - This logic combines any targets provided directly to the macro and any top-level targets specified in the schemes.
- Add `schemes_json` attribute to `_xcodeproj` rule. 
  - Note: This is a little different from the proposal. In this PR, I opted to make it a single JSON string instead of a list of string values. It makes it easier to encode the JSON and to decode once we implement that logic.
- Update `//tools/generator:xcodeproj` to accept custom schemes, but still generate the automatic schemes. 
  - To share the xcodeproj target list between the `xcodeproj` declaration and the `xcodeproj_fixtures` declaration, I reworked `//tools/generator:xcodeproj_targets.bzl` to include the schemes definition and provide functions to retrieve them. 
    - This is necessary so that name resolution APIs (`native.repository_name` and `native.package_name`) can be executed without failure. These APIs are only available during the loading phase and must be called from a build file thread.
  - NOTE: These gymnastics are only necessary because we are sharing the target list between two different packages. For regular clients, the syntax as shown in the proposal works.

### DIFF
--- a/tools/generator/xcodeproj_targets.bzl
+++ b/tools/generator/xcodeproj_targets.bzl
@@ -1,6 +1,24 @@
 """Exposes targets used by `xcodeproj` to allow use in fixture tests."""
 
-XCODEPROJ_TARGETS = [
-    "//tools/generator:generator",
-    "//tools/generator/test:tests",
-]
+load("@bazel_skylib//lib:sets.bzl", "sets")
+load("//xcodeproj:xcodeproj.bzl", "xcode_schemes")
+
+def get_xcode_schemes():
+    return [
+        xcode_schemes.scheme(
+            name = "Generator",
+            launch_action = xcode_schemes.launch_action(
+                "//tools/generator:generator",
+            ),
+            test_action = xcode_schemes.test_action([
+                "//tools/generator/test:tests",
+            ]),
+        ),
+    ]
+
+def get_xcodeproj_targets():
+    return sets.to_list(
+        xcode_schemes.collect_top_level_targets(
+            get_xcode_schemes(),
+        ),
+    )
--- a/xcodeproj/internal/bazel_labels.bzl
+++ b/xcodeproj/internal/bazel_labels.bzl
@@ -68,7 +68,7 @@ def make_bazel_labels(name_resolver = native_name_resolver):
     """Creates a `bazel_labels` module using the specified name resolver.
 
     Args:
-        name_resolver: Optional. A name resolver `struct`.
+        name_resolver: Optional. A `name_resolver` module.
 
     Returns:
         A `struct` that can be used as a Bazel labels module.
--- a/xcodeproj/internal/bazel_labels.bzl
+++ b/xcodeproj/internal/bazel_labels.bzl
@@ -1,22 +1,104 @@
 """API for resolving and managing Bazel labels"""
 
-def _absolute(label):
-    if label.startswith("@") or label.startswith("//"):
-        return label
-    repo_name = native.repository_name()
-    if repo_name == "@":
-        repo_name = ""
-    pkg_name = native.package_name()
-    name = label[1:] if label.startswith(":") else label
-    return "{repo_name}//{pkg_name}:{name}".format(
-        repo_name = repo_name,
-        pkg_name = pkg_name,
+_ROOT_SEPARATOR = "//"
+_ROOT_SEPARATOR_LEN = len(_ROOT_SEPARATOR)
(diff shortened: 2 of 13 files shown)


cgrindel:
This PR got a bit bigger than anticipated. If you would prefer, I can break out the module updates into separate PRs. Just let me know.

tools/generator/xcodeproj_targets.bzl
```diff
@@ -1,6 +1,24 @@
 """Exposes targets used by `xcodeproj` to allow use in fixture tests."""
 
-XCODEPROJ_TARGETS = [
-    "//tools/generator:generator",
-    "//tools/generator/test:tests",
-]
+load("@bazel_skylib//lib:sets.bzl", "sets")
+load("//xcodeproj:xcodeproj.bzl", "xcode_schemes")
+
+def get_xcode_schemes():
```
brentleyjones, line 6: Does this have to be a function versus a constant?

tools/generator/xcodeproj_targets.bzl
```diff
@@ -1,6 +1,24 @@
...
+            launch_action = xcode_schemes.launch_action(
+                "//tools/generator:generator",
+            ),
+            test_action = xcode_schemes.test_action([
+                "//tools/generator/test:tests",
+            ]),
+        ),
+    ]
+
+def get_xcodeproj_targets():
```
brentleyjones, line 19: Is this for test use only?

xcodeproj/internal/bazel_labels.bzl
```diff
@@ -0,0 +1,106 @@
+"""API for resolving and managing Bazel labels"""
```
brentleyjones, line 1: Did `Label()` not work as is?

tools/generator/xcodeproj_targets.bzl
cgrindel: Unfortunately, yes. If I do the following in the file,

```python
_SCHEMES = [
    xcode_schemes.scheme(
        name = "Generator",
        launch_action = xcode_schemes.launch_action(
            "//tools/generator:generator",
        ),
        test_action = xcode_schemes.test_action([
            "//tools/generator/test:tests",
        ]),
    ),
]
```

we experience the thread error that I mention in the description:

```
Error in repository_name: The native module can be accessed only from a BUILD thread. Wrap the function in a macro and call it from a BUILD file
```

To preserve the sharing of the targets list, I wrapped them in the functions that you see. They are called on the correct on the correct thread.

tools/generator/xcodeproj_targets.bzl
cgrindel: Yes. It provides the list to `xcodeproj_fixtures`.

xcodeproj/internal/bazel_labels.bzl
cgrindel: The `Label()` only accepts absolute labels. The `bazel_labels.normalize()` does the work of converting an absolute or relative label to a fully-qualified, absolute label. This allows a client to use relative labels. If we were not sharing the schemes/target list for `//tools/generator:xcodeproj`, we could declare the schemes in the `BUILD` file like so:

```python
_SCHEMES = [
    xcode_schemes.scheme(
        name = "Generator",
        launch_action = xcode_schemes.launch_action(":generator"),
        test_action = xcode_schemes.test_action([
            "//tools/generator/test:tests",
        ]),
    ),
]
```

I added a TODO to #573 to add the validity check that we discussed. I plan to add a `bazel_labels.validate()` that will confirm that a label points to something real.

brentleyjones (approved):
I feel we will need some docs in the end to help with the Function vs Constant thing. That will probably be confusing for some.

cgrindel:
Agreed. The good news is that it is really only relevant for examples in this repo. Clients should not need to worry about it.

PR #644, opened 2022-07-05
Unfocus targets with unsupported srcs file types

brentleyjones:
TIL `cc_library` and `objc_library` support pre-compiled sources in the `srcs` field. That doesn't translate well to an Xcode target, so we are excluding them for now. This fixes integration with rules_spm and clang modules.


brentleyjones:
Requires #643.

BalestraPatrick (approved):
😬

PR #656, opened 2022-07-06
Pass the custom Xcode schemes JSON from the `_xcodeproj` rule to the generator

cgrindel:
Related to #573 .

- Add DTO structs representing the custom Xcode scheme information.
- Refactor `xcode_schemes` module moving internal-only APIs to `xcode_schemes_internal`.
- After the custom Xcode schemes structs arrive in the generator, update the Bazel labels in the schemes to be `TargetID` values.


test/fixtures/tvos_app/bwx_spec.json
```diff
@@ -11,6 +11,7 @@
         "VALIDATE_WORKSPACE": false
     },
     "configuration": "darwin_x86_64-fastbuild-ST-d53d69b6b8c1",
+    "custom_xcode_schemes": null,
```
brentleyjones, line 14: Do we have an example of the DTO coming through the spec?

xcodeproj/internal/xcode_schemes_internal.bzl
```diff
@@ -0,0 +1,130 @@
...
+        A `struct` representing a launch action.
+    """
+    return struct(
+        target = target,
+        args = args,
+        env = env,
+        working_directory = working_directory,
+    )
+
+def _replace_labels_with_target_ids(scheme, configuration):
```
brentleyjones, line 83: I don't think this will work as you expect. The configuration on labels can be different. There might be multiple targets for the same configuration. The deps of a target might have a different configuration than their parent.

I believe we need to defer this to `generator`. There we can traverse the deps of a top level targets mentioned (or all if none are mentioned) and find the `TargetID` that best matches a given label.

test/fixtures/tvos_app/bwx_spec.json
cgrindel: My bad. I thought that generator had fixture tests. That is the one spot that we have custom schemes defined. I can add it to one of the examples. Are you OK with me adding custom schemes to `multiplatform`? It seems like we can exercise the target-platform logic, if we add some there.

brentleyjones: The generator does have fixture tests.

brentleyjones: I'm also fine with adding some to `multiplatform`.

xcodeproj/internal/xcode_schemes_internal.bzl
cgrindel: OK. I have a couple of questions.
1. For my edification, which configuration will be resolved [here](https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/xcodeproj/internal/xcodeproj.bzl#L288)? Is this host configuration?
2. Is there a rule-of-thumb for when label-configuration should be resolved in Starlark code vs the generator? The JSON that the Starlark code sends to the generator only seems to send `TargetID` strings.

cgrindel: I agree that we are going to iterate to pick the best one. I misunderstood where we needed to generate the list of all possible combinations. 

I will remove the label to `TargetID` code. Also, I will figure out why we are not seeing generator fixtures being updated.

test/fixtures/tvos_app/bwx_spec.json
cgrindel: The generator fixtures now contain the schemes using the updated `xcodeproj_fixture` macro.

test/fixtures/generator/bwb_spec.json
```diff
@@ -11,6 +11,23 @@
...
+            "launch_action": {
+                "args": null,
+                "env": null,
+                "target": "@//tools/generator:generator",
+                "working_directory": null
+            },
+            "name": "Generator",
+            "test_action": {
+                "targets": [
+                    "@//tools/generator/test:tests"
```
brentleyjones, line 26: We should remove the `@` if it's not an external repo, to match the way Bazel canonicalizes these, and allows for easier matching to `Target.label`.

```suggestion
                    "//tools/generator/test:tests"
```

cgrindel: OK. That is the value that comes back from `native.repository_name()`

cgrindel: Fixed in #664.

tools/generator/src/DTO/Project.swift
```diff
@@ -9,4 +9,5 @@ struct Project: Equatable, Decodable {
     let invalidTargetMerges: [TargetID: Set<TargetID>]
     let extraFiles: Set<FilePath>
     let schemeAutogenerationMode: SchemeAutogenerationMode
+    let customXcodeSchemes: [XcodeScheme]?
```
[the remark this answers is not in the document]
cgrindel: Done

tools/generator/src/DTO/XcodeScheme+LaunchAction.swift
```diff
@@ -0,0 +1,8 @@
+extension XcodeScheme {
+    struct LaunchAction: Equatable, Decodable {
+        let target: String
+        let args: [String]?
```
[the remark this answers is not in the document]
cgrindel: Done

tools/generator/src/DTO/XcodeScheme+TestAction.swift
```diff
@@ -0,0 +1,5 @@
+extension XcodeScheme {
+    struct TestAction: Equatable, Decodable {
+        let targets: Set<String>
+    }
```
brentleyjones, line 4: These are small types, and `XcodeScheme` itself is small. Can we have these nested in the `XcodeScheme` declaration itself? Or do you see each of these files getting larger for each respective type?

xcodeproj/internal/xcode_schemes_internal.bzl
```diff
@@ -0,0 +1,86 @@
+"""Internal API for creating and manipulating Xcode schemes"""
```
brentleyjones, line 1:
```suggestion
"""Internal API for creating and manipulating Xcode schemes."""
```

brentleyjones (approved):

tools/generator/src/DTO/XcodeScheme+TestAction.swift
cgrindel: Either way is fine with me. I assume that the actions will receive additional attributes over time. However, we can cross that bridge when get to it. I will put them in one file for now.

PR #665, opened 2022-07-06
Stop creating schemes for ancillary Watch App targets

brentleyjones:
Xcode doesn't create schemes for the "Watch App Container" or the Watch Kit Extension for a Watch App.


cgrindel (approved):
Sorry that I missed this previously.

brentleyjones:
No problem. It's only after I started digging into it that I noticed Xcode didn't create schemes for these targets.

PR #672, opened 2022-07-07
Add `macos_application` example

brentleyjones:
Also move the `apple_dynamic_framework_import` use case to this, as it can't be used by a command line tool.


examples/command_line/external/BUILD
```diff
@@ -1,16 +1,19 @@
 load(
     "@build_bazel_rules_apple//apple:apple.bzl",
-    "apple_dynamic_framework_import",
+    "apple_static_framework_import",
 )
 load("@rules_cc//cc:defs.bzl", "objc_import", "objc_library")
 
-apple_dynamic_framework_import(
+apple_static_framework_import(
```
brentleyjones, line 7: Seems we still had these flipped.

luispadron (approved):

PR #675, opened 2022-07-07
Add `swiftformat` script and `.swiftformat` configuration file

cgrindel:


brentleyjones:
Is the end goal to expose this as a bazel command similar to our buildifier `.fix` (ideally we apply both with one command)?

cgrindel:
@brentleyjones I use vim + ALE which runs nicklockwood's swiftformat. I continuously run into issues with the defaults not lining up with the code style in this repo. I add `.swiftformat` to contain the settings that seem to match. The `swiftformat` script helps ALE run `swiftformat` with the configuration file.

Are you OK with adding this?

cgrindel:
We can incorporate it into other checks. I am mostly trying to solve a problem that I continuously run into.

brentleyjones:
Yeah, we can start here. Longer term it would be nice to have a command I/others can run, and eventually a CI check as well.

brentleyjones (approved):

PR #676, opened 2022-07-07
Fix `.swiftformat` configuration and add first pass for `swiftlint`

cgrindel:
- `swiftformat`: Do not perform any indenting as it indents multi-line string values. The `--indentstrings false` only ensures that the string is not indented more than the start line.
- `swiftlint`: First pass at config. It excludes test directories and disables some rules that are not relevant for this repository. There are other warnings and errors. However, they may be legitimate. If we enable a CI check, we will want to address those.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,0 +1,7 @@
+disabled_rules:
+  - trailing_comma
+  - todo
+
+excluded:
+  - "**/test/**/*.swift"
+  - "**/*Tests/*.swift"
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -3,5 +3,6 @@ disabled_rules:
   - todo
 
 excluded:
+  # Exclude test files
   - "**/test/**/*.swift"
   - "**/*Tests/*.swift"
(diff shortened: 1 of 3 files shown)


.swiftlint.yml
```diff
@@ -0,0 +1,8 @@
+disabled_rules:
+  - trailing_comma
```
brentleyjones, line 2: This rule enforces trailing commas in dictionaries and lists, right? We actually want that.

cgrindel: Unfortunately, no. It warns when there is a trailing comma.

[trailing_comma](https://realm.github.io/SwiftLint/trailing_comma.html)

brentleyjones (approved):

.swiftlint.yml
rastersize: @cgrindel @brentleyjones You can configure `trialing_comma` to instead enforce that there is a trailing comma. You need to add a config statement for the rule though.

```yaml
trailing_comma:
  mandatory_comma: true
```

cgrindel: Cool. I did not know that. I will give it a try and put up a new PR. Thanks!

PR #677, opened 2022-07-07
Apply `swiftformat` `braces` rule to the codebase

cgrindel:


cgrindel:
FYI. My branch name is wrong. The `allman` option puts the brace on the next line. We do not want that. The desired behavior is the default.

brentleyjones (approved):

PR #681, opened 2022-07-07
Apply `wrapMultilineStatementBraces` for `swiftformat`

cgrindel:
`swiftformat --config .swiftformat --rules braces,wrapMultilineStatementBraces --verbose .`

It turns out #677 was not quite right. I only ran `braces` in that PR. I should have run `braces` along with `wrapMultilineStatementBraces` as the `wrapMultilineStatementBraces` takes precedence over `braces`.


brentleyjones (approved):
Nice. Looks much better.

PR #687, opened 2022-07-08
Expose `NSExtensionPointIdentifier` to `generator`

brentleyjones:
This will be used by the scheme generation code to properly create schemes for WidgetKit extensions.


cgrindel (approved):
It would be great to add some tool-specific tests at some point. It might help other folks who may need to make changes to not break anything. 🙂

brentleyjones:
What do you mean?

In regards to this PR, it's just split from a larger one that uses it, to make them easier to review. That one will be submitted soon.

brentleyjones:
Ahh. I don't have tests as the fixtures would break. I agree we could have more tests, but since I have some coverage, I'm willing to go fast here instead.

PR #698, opened 2022-07-08
Apply `blankLinesBetweenScopes` rule for `swiftformat`

cgrindel:


tools/generator/src/Extensions/Path+Extensions.swift
```diff
@@ -48,6 +48,7 @@ extension Path {
     private var isFramework: Bool {
         return self.extension == "framework"
     }
+
     private var isSceneKitAssets: Bool { self.extension == "scnassets" }
```
brentleyjones, line 52:
```suggestion
    private var isFramework: Bool { self.extension == "framework" }
    private var isSceneKitAssets: Bool { self.extension == "scnassets" }
```

cgrindel: Done

brentleyjones (approved):

PR #700, opened 2022-07-08
Apply `sortedImport` rule for `swiftformat`

cgrindel:


examples/ios_app/ExampleTests/ExampleTests.swift
```diff
@@ -1,7 +1,7 @@
-import Utils
+@testable import Example
```
brentleyjones, line 1: Can we move this to a new section below it instead? I also think there might be a rule for that?

cgrindel: There is an option for `sortedImports` called `--importgrouping`. I will set it to `testable-last`. However, this will not put a blank line. We will need to do that ourselves.

brentleyjones (approved):

Issue #93, opened 2022-03-08
Support `swift_c_module`


brentleyjones:
https://github.com/bazelbuild/rules_swift/blob/master/doc/rules.md#swift_c_module

brentleyjones:
This might be done. We need an example to prove it.

maxwellE:
I can work on adding an example for this

brentleyjones:
Thanks!

PR #711, opened 2022-07-10
Add `swift_c_module` example/fixtures

maxwellE:
Resolves Support `swift_c_module` #93


brentleyjones:
Would it be possible to add this to another example, like `command_line`?

maxwellE:
> Would it be possible to add this to another example, like `command_line`?

Absolutely, did not know how granular we wanted to go with these examples

examples/command_line/swift_c_module/BUILD
```diff
@@ -0,0 +1,27 @@
+load("@build_bazel_rules_swift//swift:swift.bzl", "swift_c_module")
+
+cc_library(
+    name = "lib_impl",
```
brentleyjones, line 4: To keep the project looking nice, can we rename this to `c_lib` and remove the below `alias`? Will prevent the name clashing, and another example already shows the `alias` case.

brentleyjones (approved):

PR #723, opened 2022-07-11
Improve schemes for hosted targets

brentleyjones:


tools/generator/src/DTO/Project.swift
```diff
@@ -6,7 +6,7 @@ struct Project: Equatable, Decodable {
     let buildSettings: [String: BuildSetting]
     var targets: [TargetID: Target]
     let targetMerges: [TargetID: Set<TargetID>]
-    let targetHosts: [TargetID: Set<TargetID>]
+    let targetHosts: [TargetID: [TargetID]]
```
cgrindel, line 9: Curious. Why are these no longer unique?

cgrindel (approved):

tools/generator/src/DTO/Project.swift
brentleyjones: They were always unique by construct. I needed them to be ordered as well.

PR #732, opened 2022-07-12
Add App Clip example

### LINKED ISSUE #79: Support `ios_app_clip`

brentleyjones:
Resolves #79.

### DIFF
--- a/examples/multiplatform/AppClip/Entitlements.entitlements
+++ b/examples/multiplatform/AppClip/Entitlements.entitlements
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>application-identifier</key>
+	<string>TEAMID.APP_CLIP_BUNDLE_ID</string>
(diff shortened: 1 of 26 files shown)


examples/multiplatform/AppClip/Entitlements.entitlements
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>application-identifier</key>
+	<string>TEAMID.APP_CLIP_BUNDLE_ID</string>
```
brentleyjones, line 6: Having to set this for BwB, but not BwX, continues to stump me. Ideally rules_apple would handle that, but then you would have to know info from your parent app, which is tricky. Maybe even just failing to build if it's not defined...

BalestraPatrick: +1 to this. We have a macro that works around this annoying issue.

BalestraPatrick (approved):

examples/multiplatform/AppClip/Entitlements.entitlements
brentleyjones: https://github.com/bazelbuild/rules_apple/issues/1570

PR #733, opened 2022-07-12
Implement Xcode scheme `TargetID` selection logic

cgrindel:
Related to #573 .


tools/generator/test/XcodeScheme+ExtensionsTests.swift
```diff
@@ -0,0 +1,371 @@
...
+    func test_resolveTargetIDs_withIOSAppScheme() throws {
+        // Both the device and simulator TargetID values are available.
+        // Prefer the TargetID values for the simulator.
+        let actual = try iOSAppScheme.resolveTargetIDs(targets: targets)
+        let expected = [
+            libLabel: libiOSx8664TargetID,
+            iOSAppLabel: iOSAppiOSx8664TargetID,
+        ]
+        XCTAssertEqual(expected, actual)
+    }
```
cgrindel, line 58: @brentleyjones I just wanted to confirm that we do want to prefer the simulator platform, if both the simulator and device targets are available. The logic `Platform` seems to suggest that is the case. Is that your understanding?

tools/generator/src/Extensions/XcodeScheme+Extensions.swift
```diff
@@ -0,0 +1,167 @@
...
+
+// MARK: Resolve TargetIDs
+
+extension XcodeScheme {
+    /// Represents a configuration string (Target.configuration).
+    typealias Configuration = String
+
+    /// Determines the mapping of `LabelValue` to the `TargetID` values based upon the scheme's
+    /// configuration.
+    func resolveTargetIDs(targets: [TargetID: Target]) throws -> [LabelValue: TargetID] {
```
cgrindel, line 30: @brentleyjones The primary resolution logic is going to return a single `TargetID` per `LabelValue`. In the case when someone defines a "kitchen sink" scheme (i.e., put all targets for the `multiplatform` example in a single scheme), which of the platforms would you expect to win? Based upon the sort logic in `Platform`, it will be iOS simulator. What would you expect to happen in Xcode?

tools/generator/test/XcodeScheme+ExtensionsTests.swift
brentleyjones: Yes

tools/generator/src/Extensions/XcodeScheme+Extensions.swift
brentleyjones: If many different platform tests are listed as top level targets, we could find each platform version of a listed library, but just finding the "best" one per Platform sorting logic would be an acceptable first version.

cgrindel: The reason that I ask is that I originally had this API return a `[LabelValue: [TargetID]]`. I can go back to that version if that is what we want.

tools/generator/src/Extensions/XcodeScheme+Extensions.swift
```diff
@@ -0,0 +1,167 @@
...
+
+            let targetID: TargetID
+            if schemeLabel.isTopLevel {
+                // If schemeLabel is top-level, then get the Target-TargetID with the "best"
+                // platform.
+                targetID = try targetInfo.best().id
+            } else {
+                // If schemeLabel is not top-level, then collect all of the Target-TargetID for the
+                // top-level configurations and select the first one.
+                let targetIDs = topLevelPlatforms.compactMap { targetInfo.byPlatform[$0]?.id }
```
brentleyjones, line 70: This can select the wrong targets if  there are multiple that differ by the same platform, but different minOS or bazel configuration. I think you need to look at all of the deps of the top level targets and find the first dep that matches the label.

brentleyjones: > I presume that we will use a breadth-first search as to select the `TargetID` that is "closest" to top-level target. Does that sound right?

Yes. And we need to collect these for each top-level target. _Then_ we can select the "best" from _those_.

> Is that an implementation detail or can we rely on it? 

Implementation detail. One day I believe I'll start hashing those values to reduce `spec.json` size. So doing a O(1) lookup of a `Target` from the `TargetID` to get it's `.label` would be the way to go.

brentleyjones: Because of the way `dependencies` work, it might have unrelated platform versions. An example is an iOS app having a dependency on a watch app. I think the search shouldn't match or dig into dependencies that don't have a matching `Platform`.

cgrindel: I have updated the resolution logic to reduce the build tree targets to a set that only includes a top-level platform.

tools/generator/test/XcodeScheme+ExtensionsTests.swift
```diff
@@ -0,0 +1,396 @@
...
+        label: iOSAppLabel,
+        configuration: applebiniOSiOSx8664Configuration,
+        platform: iphoneSimulatorPlatform,
+        product: .init(
+            type: .application,
+            name: "a",
+            path: .generated("z/A.a")
+        ),
+        // This target has a dependency on a watchOS app
+        dependencies: [libiOSx8664TargetID, watchOSAppwatchOSx8664TargetID]
```
cgrindel, line 297: @brentleyjones I updated the test data such that the iOS app includes a dependency on a watchOS app. Does this approximate the scenario that was mentioned in the other thread?

tools/generator/src/Extensions/XcodeScheme+Extensions.swift
```diff
@@ -0,0 +1,170 @@
...
+            resolvedTargetIDs[schemeLabel.label] = targetID
+        }
+
+        // Reduce the number of targets being evaluated. Only include those that have one of the
+        // top-level platforms.
+        let targetsWithTopLevelPlatforms = targets.filterDependencyTree(
+            startingWith: topLevelTargetIDs
+        ) { target in
+            topLevelPlatforms.contains(target.platform)
+        }
```
cgrindel, line 68: @brentleyjones This filters the set of targets being evaluated to those that are in the build tree and has a platform that is one of the top-level platfoms.

tools/generator/test/XcodeScheme+ExtensionsTests.swift
brentleyjones: Yes.

tools/generator/src/Extensions/Dictionary+Extensions.swift
```diff
@@ -0,0 +1,58 @@
...
+        if depTargetIDs.isEmpty { return result }
+
+        // Collect the dependencies for the start targets
+        let deps = try filterDependencyTree(startingWith: depTargetIDs, isIncluded)
+
+        // Merge the results
+        return result.merging(deps) { current, _ in current }
+    }
+
+    // TODO: Remove firstTargetID if not being used
```
brentleyjones, line 30: Still apply?

tools/generator/src/Extensions/Dictionary+Extensions.swift
```diff
@@ -0,0 +1,58 @@
+extension Dictionary where Key == TargetID, Value == Target {
+    // func filterDependencyTree<StartTargets: Sequence>(
+    //     startingWith startTargetIDs: StartTargets
+    // ) -> [Key: Value] where StartTargets.Element == Self.Key {
+    //     return filterDependencyTree(startingWith: startTargetIDs) { _ in true }
+    // }
```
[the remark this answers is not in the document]
cgrindel: Removed.

tools/generator/src/Extensions/Dictionary+Extensions.swift
cgrindel: Removed.

tools/generator/src/Extensions/XcodeScheme+Extensions.swift
cgrindel: Yes. I thought that applying the filter upfront made the code a little easier to understand as it simplifies the label check below. Also, I was using the `multiplatform` example as a use case. Filtering the targets list once does reduce the number of iterations that we potentially need to do when we perform the label search below.

brentleyjones (approved):

PR #741, opened 2022-07-14
Introduce `BazelLabel` type

cgrindel:
I implemented the type such that it parses early. If I implemented exactly like `TargetID`, the client of the type would not know if they had a valid value or not. In other words, I wanted to ensure that a `label.name` would not fail.

Once this is merged, I can update the rest of the code to use it.

### DIFF
--- a/tools/generator/src/DTO/BazelLabel.swift
+++ b/tools/generator/src/DTO/BazelLabel.swift
@@ -1,0 +1,63 @@
+struct BazelLabel: Equatable, Hashable, Decodable {
+    let repository: String
+    let package: String
+    let name: String
+}
+
+// extension BazelLabel {
+//     static func parse(_: String) -> BazelLabel? {
+//         // TODO: IMPLEMENT ME!
+//         return BazelLabel(
+//             repository: "",
+//             package: "",
+//             name: ""
+//         )
+//     }
+// }
+
+extension BazelLabel {
+    init?(_: String) {
+        // TODO: IMPLEMENT ME!
+        self.init(
+            repository: "",
+            package: "",
+            name: ""
+        )
+    }
+}
+
+extension BazelLabel: CustomStringConvertible {
+    var description: String {
+        return "\(repository)//\(package):\(name)"
+    }
+}
+
+extension BazelLabel: RawRepresentable {
+    init?(rawValue: String) {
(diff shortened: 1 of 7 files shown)


tools/generator/src/DTO/BazelLabel.swift
```diff
@@ -0,0 +1,83 @@
...
+        var container = encoder.singleValueContainer()
+        try container.encode(rawValue)
+    }
+}
+
+extension BazelLabel: Decodable {
+    init(from decoder: Decoder) throws {
+        let value = try decoder.singleValueContainer().decode(String.self)
+        // TODO: Switch to throwing an error
+        self.init(value)!
```
brentleyjones, line 81: Since you have all of these parts when you create the label by hand, or we can extract them from `Label`, do we want to have this initialized with a structured type from the spec side?

That can be a followup PR. Mainly thinking about the parsing performance.

tools/generator/src/DTO/BazelLabel.swift
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+extension BazelLabel: RawRepresentable {
+    init?(rawValue: String) {
+        self.init(rawValue)
+    }
+
+    var rawValue: String { "\(self)" }
+}
```
brentleyjones, line 62: Since we have a custom decoder (which I'm not sure we need if we have this), I don't think we need to/should conform to `RawRepresentable`.

brentleyjones (approved):

tools/generator/src/DTO/BazelLabel.swift
```diff
@@ -0,0 +1,83 @@
...
+
+extension BazelLabel: ExpressibleByStringLiteral {
+    init(stringLiteral value: String) {
+        self.init(value)!
```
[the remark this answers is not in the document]
cgrindel: Moved to test library.

tools/generator/src/DTO/BazelLabel.swift
brentleyjones: Let's keep it like it is right now, but I think longer term we will probably want to send it over decomposed. Analysis caching is powerful, so moving more over to that versus generation is ideal.

PR #742, opened 2022-07-14
Use `BazelLabel`

cgrindel:


tools/generator/test/Fixtures.swift
```diff
@@ -1962,9 +1962,6 @@ sed \
         let (files, _) = Fixtures.files(in: pbxProj, parentGroup: mainGroup)
         let products = Fixtures.products(in: pbxProj, parentGroup: mainGroup)
 
-        let internalDirectoryName = "rules_xcp"
-        let workspaceOutputPath: Path = "Project.xcodeproj"
-
```
cgrindel, line 1967: I noticed the warnings for these unused variables. So, I went ahead and removed them.

tools/generator/test/Target+Testing.swift
```diff
@@ -28,7 +28,7 @@ extension Target {
     ) -> Self {
         return Target(
             name: product.name,
-            label: label ?? "//some/package:\(product.name)",
+            label: label ?? .init("//some/package:\(product.name)")!,
```
brentleyjones, line 31: Shouldn't this be covered by the string literal conformance?

cgrindel: Unfortunately, no. It is a `String`, not a `StringLiteral`. The substitution keeps it from being a `StringLiteral`.

tools/generator/src/DTO/BazelLabel.swift
```diff
@@ -11,10 +11,20 @@ extension BazelLabel {
...
+        case missingOrTooManyRootSeparators
+        case missingNameAndPackage
+        case tooManyColons
+    }
+
+    private static func parse(
+        _ value: String
+    ) throws -> (repository: String, package: String, name: String) {
+        // swiftlint:disable:previous large_tuple
+
```
cgrindel, line 24: I updated the parse logic to throw errors to provide better feedback. In addition, I moved the parse logic to its own function so that it is easier to implement the other initializers.

tools/generator/src/Generator/DisambiguateTargets.swift
```diff
@@ -481,7 +481,7 @@ extension ConsolidatedTarget {
     /// The normalized label is used during target disambiguation. It allows the
     /// logic to differentiate targets where the names only differ by case.
     var normalizedLabel: String {
-        return label.lowercased()
+        return "\(label)".lowercased()
     }
 }
```
cgrindel, line 486: Is this `normalizedLabel` a one-off function? I considered adding a normalized property to `BazelLabel`.

brentleyjones: Yeah, it's only needed here.

brentleyjones (approved):

PR #746, opened 2022-07-15
Process linkopts in `generator`

brentleyjones:
And use the correct path for Xcode versus Bazel generated files. This will allow us to stop copying swiftmodules to `$BUILD_DIR`.


tools/generator/src/DTO/Target+LinkerFlags.swift
```diff
@@ -0,0 +1,146 @@
...
+        }
+        .joined(separator: ",")
+}
+
+private func processLinkoptComponent(
+    _ opt: String,
+    xcodeGeneratedFiles: Set<FilePath>,
+    filePathResolver: FilePathResolver
+) throws -> String {
+    /*
```
BalestraPatrick, line 94: What is this referring to?

BalestraPatrick (approved):

tools/generator/src/DTO/Target+LinkerFlags.swift
brentleyjones: I forgot to remove the comment.

PR #761, opened 2022-07-18
Use `$BAZEL_OUT` for modulemaps and header search paths

brentleyjones:
To make this work in BwX mode, a VFS overlay was added to cause the Xcode produced Swift generated headers to take precedence over Bazel ones.


brentleyjones:
@BalestraPatrick (since your setup is the best case for modulemaps, imo)

BalestraPatrick (approved):
Seems to work fine in BwX mode!

PR #765, opened 2022-07-19
Fix duplicate `LinkerInputs.staticLibraries` error

### LINKED ISSUE #764: "Decoded elements aren't unique" error
As of c538247afae41ffef463b6d1d1856453b0b21fd2, I get an error when trying to run the xcodeproj target:

```
At codingPath [targets, Index 55, linkerInputs, staticLibraries]: Decoded elements aren't unique (first duplicate at offset 73)
Please file a bug report at https://github.com/buildbuddy-io/rules_xcodeproj/issues/new?template=bug.md
```

This seems to be a recent regression. I do not get this error with 435cd30e1d4da0977eb9c7d63c98f4976dd856ae.

brentleyjones:
Fixes #764.

It seems that the same library can be in collected multiple times.

### DIFF
--- a/xcodeproj/internal/linker_input_files.bzl
+++ b/xcodeproj/internal/linker_input_files.bzl
@@ -3,7 +3,7 @@
 load("@bazel_skylib//lib:sets.bzl", "sets")
 load("@bazel_skylib//lib:collections.bzl", "collections")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
-load(":collections.bzl", "flatten", "set_if_true")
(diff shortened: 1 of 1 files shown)


jfirebaugh:
It does not fix it. I get the same error as before.

brentleyjones:
Are you able to provide a repro? I can't reason why this would happen.

jfirebaugh:
Here's a reduced test case: https://github.com/jfirebaugh/rules_xcode_reduction/tree/764

brentleyjones:
Thank you for that! This PR should now fix it.

brentleyjones:
You're welcome! ❤️

PR #780, opened 2022-07-20
Stop producing outputs output groups in BwX

brentleyjones:
Besides for doing a little less work, the main reason to not create these output groups is to prevent people from using them external to Xcode when it doesn't make sense. If you want to use these output groups, you need to set the correct `build_mode` on `xcodeproj`.


xcodeproj/internal/output_files.bzl
```diff
@@ -83,7 +88,7 @@ def _create(
         ],
     )
 
-    if direct_outputs:
+    if should_produce_output_groups and direct_outputs:
```
brentleyjones, line 91: @cgrindel here

Issue #779, opened 2022-07-20
Feature Request: Support `rules_ios` static framework bundling


luispadron:
## Summary

Currently `rules_ios` support is limited. There are some assumptions both `rules_xcodeproj` and `rules_ios` make. 

At the moment I'm attempting to generate a project with a single static framework target. This currently fails due to the way `rules_ios` builds the `AppleBundleInfo` provider: [see here](https://github.com/bazel-ios/rules_ios/blob/f8d277c52aef10cdfa7224a72d8414c46e225839/rules/framework.bzl#L770-L792).

From the brief look I've had it looks like `rules_ios` needs to provide:
- [x] `executable_name`: https://github.com/bazel-ios/rules_ios/pull/513
- [x] `minimum_deployment_os_version`: https://github.com/bazel-ios/rules_ios/pull/514
- [x] `archive`
- [x] `archive_root`

I'm opening a feature request in this repository in case we don't need to require these fields for project generation which would make it easier to support other rules. If we do require the fields, we can work to add them to `rules_ios` and close this request once we have supported static frameworks from `rules_ios`.

I'll work on creating a reproducible example (without proprietary code) and follow up in this issue.

luispadron:
After fixing `executable_name` a new errors surfaces:

```
ERROR: /Users/lpadron/Development/cash-ios-bazel/Pods/StringTemplate/BUILD.bazel:11:15: in @com_github_buildbuddy_io_rules_xcodeproj//xcodeproj/internal:default_automatic_target_processing_aspect.bzl%default_automatic_target_processing_aspect,@com_github_buildbuddy_io_rules_xcodeproj//xcodeproj/internal:xcodeproj_aspect.bzl%xcodeproj_aspect aspect on apple_framework_packaging rule //Pods/StringTemplate:StringTemplate: 
Traceback (most recent call last):
        File "/private/var/tmp/_bazel_lpadron/fb0c661113a61963d843063c755533c2/external/com_github_buildbuddy_io_rules_xcodeproj/xcodeproj/internal/xcodeproj_aspect.bzl", line 48, column 33, in _xcodeproj_aspect_impl
                create_xcodeprojinfo(
        File "/private/var/tmp/_bazel_lpadron/fb0c661113a61963d843063c755533c2/external/com_github_buildbuddy_io_rules_xcodeproj/xcodeproj/internal/xcodeprojinfo.bzl", line 365, column 44, in create_xcodeprojinfo
[...]

PR #781, opened 2022-07-20
Default `bundle_info.executable_name` to `bundle_info.bundle_name`

brentleyjones:
Part of #779.

This handles custom rules that don't set `executable_name`.


brentleyjones:
cc: @luispadron

PR #803, opened 2022-07-23
Create custom Xcode schemes

cgrindel:
Related to #573.

- Add initializers with defaults to `XcodeScheme` structs.
- Add `XcodeScheme.ProfileAction`.
- Move `shouldCreateScheme` filter to `createAutogeneratedXCSchemes`. It was in the shared code.
- Add `targetResolver(referencedContainer:)` to `Fixtures`.


tools/generator/src/Extensions/Dictionary+Extensions.swift
```diff
@@ -0,0 +1,16 @@
+extension Dictionary {
+    /// Look up a value using the key. Throw an error if it is not found.
```
brentleyjones, line 2: Can you add this in its own PR, along with changing code to use this new method?

tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -4,11 +4,14 @@ enum XCSchemeConstants {
     // GH399: Derive the defaultLastUpgradeVersion
     static let defaultLastUpgradeVersion = "1320"
     static let lldbInitVersion = "1.7"
-    static let defaultBuildConfigurationName = "Debug"
     static let posixSpawnLauncher = "Xcode.IDEFoundation.Launcher.PosixSpawn"
     static let customLLDBInitFile = "$(BAZEL_LLDB_INIT)"
 }
 
+extension String {
+    static let defaultBuildConfigurationName = "Debug"
```
brentleyjones, line 12: I prefer refactor and behavior changes to be in separate PRs. Can you make this change in its own PR?

cgrindel: #804

tools/generator/src/Extensions/Dictionary+Extensions.swift
cgrindel: I have added the new method in #805. I am happy to help with switching code over to use this new method. However, I don't think that this should hold up this PR.

brentleyjones (dismissed):
If we are creating custom schemes, shouldn't the ones for Generator have changed?

cgrindel:
@brentleyjones You are correct. I will investigate before merging this PR.

cgrindel:
The `//tools/generator:xcodeproj` target is configured with `scheme_autogeneration_mode` set to `all`. Hence, it is generating both the custom and autogenerated scheme. The custom scheme is named `Generator`. One of the autogenerated schemes is called `generator`. The last one in the list wins.

For this PR, I will rename the custom scheme to not conflict. I have added two tasks to #573. 
- Enable `scheme_autogenrate_mode` `auto`.
- Ensure that autogenerated scheme names do not conflict with custom scheme names.

cgrindel:
I verified that the custom scheme is written (Good catch, @brentleyjones). I launched the scheme in Xcode and was able to build and test it successfully.

brentleyjones (approved):
When there is conflicting schemes, can we have the custom ones just override the automatic one? Seems better than an error, since one controls the name of the custom scheme.

brentleyjones:
(Would require changing the order that they are made in, but I think that's for the best.)

cgrindel:
> When there is conflicting schemes, can we have the custom ones just override the automatic one? Seems better than an error, since one controls the name of the custom scheme.

Sure. I was going to add code that checks an autogenerated scheme name and add a prefix (e.g., `(generated)`), if there was a conflict. However, I am happy to just skip the autogenerated one. Whichever you want is fine with me.

brentleyjones:
I like skipping the autogenerated one. Then one can use `top_level` and still define a custom one to "tweak" and existing top level autogenerated scheme.

PR #807, opened 2022-07-23
Keep unique set of TargetInfo values in XCSchemeInfo structs

cgrindel:
Related to #573.

- Implement `Hashable` for `XCScheme.BuildableReference`, `XCSchemeInfo.HostInfo`, and `XCSchemeInfo.TargetInfo`.
- Update `XCSchemeInfo.BuildActionInfo` and `XCSchemeInfo.TestActionInfo` to store their `targetInfos` as `Set<XCSchemeInfo.TargetInfo>`.

### DIFF
--- a/tools/generator/src/Extensions/XCScheme+Extensions.swift
+++ b/tools/generator/src/Extensions/XCScheme+Extensions.swift
@@ -212,3 +212,17 @@ extension XCScheme.ArchiveAction {
         )
     }
 }
+
+extension XCScheme.BuildableReference: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        // This should match the Equatable conformance for XCScheme.BuildableReference.
+        // NOTE: The equality check in XCScheme.BuildableReference checks the `blueprint` property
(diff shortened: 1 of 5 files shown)


tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -212,3 +212,17 @@ extension XCScheme.ArchiveAction {
         )
     }
 }
+
+extension XCScheme.BuildableReference: Hashable {
```
brentleyjones, line 216: Should this be in XcodeProj instead? We can submit a PR, and use a patch until it's accepted.

cgrindel: Potentially, yes. To keep the custom scheme train moving, I would like to do this here for now.

cgrindel: I can track it as a TODO in #573.

brentleyjones (approved):

PR #809, opened 2022-07-23
Suppress autogenerated schemes that conflict with custom scheme names

cgrindel:
Related to #573.


brentleyjones (approved):
Can you rename back the generator custom scheme so the fixtures show this off?

tools/generator/xcodeproj_targets.bzl
```diff
@@ -8,7 +8,7 @@ GENERATOR_SCHEME_AUTOGENERATION_MODE = "all"
 def get_xcode_schemes():
     return [
         xcode_schemes.scheme(
-            name = "Generator App",
+            name = "Generator",
```
brentleyjones, line 11: I know it doesn't exercise the normalization logic, but I really want this scheme to be named after the product, which is lowercase.

```suggestion
            name = "generator",
```

cgrindel: Done

brentleyjones:
Can you update the fixtures as well?

cgrindel:
I think that I have the filename case issue resolved. I am seeing it with the original lowercase name.

PR #811, opened 2022-07-24
Fix sort for `XCScheme.BuildableReference` values

cgrindel:
Related to #573.

This enforces a sort order for `XCScheme.BuildableReference` values. This should minimize unintended diff changes in the fixtures.


cgrindel:
@brentleyjones I would like to merge this PR before #809.

tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -225,3 +225,20 @@ extension XCScheme.BuildableReference: Hashable {
...
+            blueprintName,
+            referencedContainer,
+            buildableIdentifier,
+        ].joined(separator: "|")
+    }
+}
+
+extension Sequence where Element == XCScheme.BuildableReference {
+    var inStableOrder: [XCScheme.BuildableReference] {
+        return sorted { $0.sortValue < $1.sortValue }
```
brentleyjones, line 242: `BlueprintName` is the target name, which is unique within a project (and we aren't doing workspaces), so we can use just that for sorting. We want to use `sortByLocalizedStandard()` for any user visible sorting:

```suggestion
        return sortByLocalizedStandard(\.blueprintName)
```

tools/generator/src/Generator/XCSchemeInfo+TargetInfo.swift
```diff
@@ -161,10 +161,16 @@ extension XCSchemeInfo.TargetInfo {
 
 // MARK: Sequence Extensions
 
+extension Sequence where Element == XCSchemeInfo.TargetInfo {
+    var inStableOrder: [XCSchemeInfo.TargetInfo] {
+        return sorted { $0.pbxTarget.name < $1.pbxTarget.name }
```
brentleyjones, line 166:
```suggestion
        return sortByLocalizedStandard(\.pbxTarget.name)
```

brentleyjones (changes requested):

brentleyjones (approved):

PR #813, opened 2022-07-25
Stop storing `SwiftInfo` in `compilation_providers`

brentleyjones:
We don't need the full providers, and it currently gets in the way of de-stringifying `xcode_target`.


examples/ios_app/test/fixtures/bwb.xcodeproj/project.pbxproj
```diff
@@ -1719,6 +1719,7 @@
 				ENABLE_TESTABILITY = YES;
 				ENABLE_TESTING_SEARCH_PATHS = YES;
 				GCC_OPTIMIZATION_LEVEL = 0;
+				GCC_PREPROCESSOR_DEFINITIONS = AWESOME;
```
brentleyjones, line 1722: It makes sense that these are back, and it was actually a bug that they were missing before.

BalestraPatrick (approved):

PR #821, opened 2022-07-25
Store `File` with `file_path`

brentleyjones:
This will allow us to perform some logic based on the `product` of an `xcode_target`.


brentleyjones:
This changes how they hashed. Was easier to just compare `file_path`s instead.

PR #827, opened 2022-07-26
Set `unfocused_targets` for `generator`

brentleyjones:
This is mainly to have an example of `unfocused_targets` in the fixtures. While we don't have Bazel index importing we want to minimize the use of focused targets, as we don't have indexing for unfocused targets.


brentleyjones:
Draft only because it's a stacked PR. Submitted to be able to show how the parent PR manifests in an example.

PR #831, opened 2022-07-27
Rename `xcodeproj.targets` to `top_level_targets`

brentleyjones:
There has been some confusion on what targets should or need to be set on `xcodeproj.targets`. They should only be top-level targets, and not any of their dependencies. Setting non-top-level targets will result in misconfigured targets appearing in the project.


jpsim (approved):
Also need to update the readme here https://github.com/buildbuddy-io/rules_xcodeproj/blob/c252daab6feab61f960d72ccceb2a9633dcc2957/README.md?plain=1#L134

PR #832, opened 2022-07-27
Fix installer when using `bazel_path`

brentleyjones:


jpsim (approved):
Thanks for the quick fix!

Issue #617, opened 2022-06-30
Bug: Debugging can get broken by bazel disk or remote cache usage


brentleyjones:
## Describe the bug

If the project isn't setting `--features=swift.cacheable_swiftmodules` or `-serialize-debugging-options`, and it's setting a bazel cache, then currently the Index Build bazel-out and the Normal Build bazel-out will upload swiftmodules with different paths serialized into them. Currently both builds will use the same cache keys, causing a sort of cache poisoning. This can result in errors like this in lldb:

```
 SwiftASTContext("MODULE")::GetASTContext() -- failed to initialize ClangImporter: error: /Users/USER/Library/Developer/Xcode/DerivedData/PROJECT_HASH/Index/Build/Intermediates.noindex/bazel_output_base/execroot/REPO/bazel-out/ios-sim_arm64-min14.0-applebin_ios-ios_sim_arm64-dbg-ST-651f601a44de/bin/external/com_google_protobuf/objectivec/objectivec.swift.modulemap:1:8: error: redefinition of module 'objectivec_objectivec'
module "objectivec_objectivec" {
       ^

/private/var/tmp/_bazel_USER/a78081980f52363e79949cde51977ac5/execroot/REPO/bazel-out/ios-sim_arm64-min14.0-applebin_ios-ios_sim_arm64-dbg-ST-651f601a44de/bin/external/com_google_protobuf/objectivec/objectivec.swift.modulemap:1:8: note: previously defined here
module "objectivec_objectivec" {
       ^
```

## Expected behavior

Use `--features=swift.cacheable_swiftmodules` and have an automated solution that doesn't require changing the build graph to add `-serialize-debugging-options` to any local-only targets.

PR #836, opened 2022-07-28
Set `target.swift-*` lldb settings

brentleyjones:
Part of #617.

This allows lldb to work correctly if using the `-noserialize-debugging-options` `swiftc` flag.

### DIFF
--- a/examples/ios_app/test/fixtures/bwb.xcodeproj/project.pbxproj
+++ b/examples/ios_app/test/fixtures/bwb.xcodeproj/project.pbxproj
@@ -12,6 +12,7 @@
 			buildConfigurationList = 914277EC9F57B808A8817CF5 /* Build configuration list for PBXAggregateTarget "BazelDependencies" */;
 			buildPhases = (
 				9A630CF63C380FAE522825A9 /* Bazel Build */,
+				20BEB4AE7798C11A8A904F5E /* Create swift_debug_settings.py */,
 			);
 			dependencies = (
 			);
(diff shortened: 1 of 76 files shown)


brentleyjones:
I need to fix the lldb module name matching, as these two are not equal:
```
/Users/brentley/Library/Developer/CoreSimulator/Devices/7022ED79-E71C-425C-BA0E-634AD3375CD2/data/Containers/Bundle/Application/C23EF714-036C-4EE5-BB91-EF2F66E0AE33/Example.app/Example_ExecutableName
/Users/brentley/Library/Developer/Xcode/DerivedData/bwb-ezjualidhmoodmfdrsrwmwoehsyl/Build/Products/bazel-out/ios-x86_64-min15.0-applebin_ios-ios_x86_64-dbg-ST-a0d0e3b8f217/bin/Example/Example.app/Example_ExecutableName
```

brentleyjones:
Almost there. These two keys don't match. Trying to figure out when we need to add the minimum os to the triple:
```
"arm64-apple-macosx11.0.0 tests.xctest/Contents/MacOS/tests"
"arm64-apple-macosx tests.xctest/Contents/MacOS/tests"
```

brentleyjones:
I swear I was seeing the triples without the os version before. Now I only see the one with os version. Maybe I can add a fallback key lookup just in case...

brentleyjones:
Regardless, latest issue is that these two don't match:
```
lldb: "arm64-apple-macosx11.0.0 tests.xctest/Contents/MacOS/tests"
generated: "arm64-apple-macosx12.0.0 tests.xctest/Contents/MacOS/tests"
```

I have `--macos_minimum_os=12.0` set, so I'm not sure what's up with that yet.

brentleyjones:
And now it works. Hmmmmm 🤔. I'm thinking it was some sort of caching issue, so we just need the one backup now, though maybe I can convert the lldb one to the generic one when doing that. The downside is the small chance of collisions if there are multiple versions of executable paths that only differ on os version, but the backup is better than nothing.

brentleyjones:
Fallback added, I think this is good to go.

@jpsim It would be nice to make sure that this doesn't regress debugging for EM. It doesn't set the flag that removes debug information (`--features=swift.cacheable_swiftmodules`) but EM already sets that, so this should only improve debugging of Swift code for you.

@cgrindel can you test this on your internal project as well?

brentleyjones:
(And I found the 11.0 vs 12.0, and it's related to generator's tests. Digging in.)

brentleyjones:
And now that is fixed! Was looking at yet another wrong place for the triple. So many places to get it in lldb...

jpsim:
Lightweight debugging testing with EM seems to still work. I didn't review the changes here in depth though, there's a lot that was touched.

brentleyjones:
I got a couple offline reports of this working, so merging.

PR #839, opened 2022-07-29
Add `Platform.OSVersion`

brentleyjones:
We will need the `Platform.OSVersion.fullVersion` property in lldb changes.

### DIFF
--- a/tools/generator/src/DTO/Platform.swift
+++ b/tools/generator/src/DTO/Platform.swift
@@ -16,11 +16,35 @@ struct Platform: Equatable, Hashable, Decodable {
         case watchOSSimulator = "watchsimulator"
     }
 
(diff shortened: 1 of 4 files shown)


tools/generator/src/DTO/Platform.swift
```diff
@@ -69,6 +93,13 @@ extension Platform.Variant {
     }
 }
 
+extension Platform.OSVersion: Comparable {
+    static func < (lhs: Platform.OSVersion, rhs: Platform.OSVersion) -> Bool {
+        return lhs.rawValue.compare(rhs.rawValue, options: .numeric)
+            == .orderedAscending
```
brentleyjones, line 99: We could store the components, and do a compare on each... but I'll leave that for a future optimization.

PR #840, opened 2022-07-29
Merge `Target.platform`

brentleyjones:
The library being merged into a top level target might have a different minimum os version. Now that 1ab36b4c6814faa73849c71e9d02a20181c13336 has landed, we need to account for the differences.

### DIFF
--- a/tools/generator/src/DTO/Target.swift
+++ b/tools/generator/src/DTO/Target.swift
@@ -5,7 +5,7 @@ struct Target: Equatable {
     let label: BazelLabel
     let configuration: String
     var packageBinDir: Path
-    let platform: Platform
+    var platform: Platform
     let product: Product
     var isSwift: Bool
     let testHost: TargetID?
--- a/tools/generator/src/Generator/ProcessTargetMerges.swift
+++ b/tools/generator/src/Generator/ProcessTargetMerges.swift
@@ -51,6 +51,9 @@ exist
                 // but that currently seems like too much work.
                 merged.packageBinDir = merging.packageBinDir
 
+                // Update platform
+                merged.platform = merging.platform
+
                 // Update isSwift
                 merged.isSwift = merging.isSwift
 
--- a/tools/generator/test/Fixtures.swift
+++ b/tools/generator/test/Fixtures.swift
@@ -48,6 +48,7 @@ enum Fixtures {
     static let targets: [TargetID: Target] = [
         "A 1": Target.mock(
             packageBinDir: "bazel-out/a1b2c/bin/A 1",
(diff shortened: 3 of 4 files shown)


brentleyjones:
It should pop up here in CI as well, but @cgrindel I don't understand the schemes test failure.

brentleyjones:
Maybe it's not reproducing. I'll fix my fixture failure in a min and see if it passes on CI.

cgrindel:
I am seeing the error on my laptop. Still looking...

brentleyjones:
Now I see it on CI as well. "Good" 😆.

cgrindel:
The issue is that, in the Fixtures, `A 1` is not seen as a dependency of `A 2` when filtering the dependency tree by top-level platform. The `A 1` target is set to `minimumOsVersion` of `10.0` while `A 1` is set to `11.0`. Changing the `A 1` to `11.0` will address the failing tests.

brentleyjones:
But `A 1` needs to be 10.0, to exercise the change of the PR. The fixture is in a valid state right now, it's valid for dependencies to have different minimum os versions. Seems we have a bug in that filtering logic.

cgrindel:
[Here is the filter in question.](https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/tools/generator/src/Extensions/XcodeScheme%2BExtensions.swift#L62-L68)

Perhaps when should implement an `isCompatible()` function on `Platform`. It would do an equality check for `os`, `variant`, and `arch`. It would then do a less-than-or-equal-to check for `minimumOsVersion` and `minimumDeploymentOsVersion`.

brentleyjones:
Yeah, that sounds good.

brentleyjones:
@cgrindel this is good for a review now.

tools/generator/src/DTO/Platform.swift
```diff
@@ -47,6 +47,16 @@ extension Platform {
 \(arch)-apple-\(variant.triplePrefix)\(osVersion)\(variant.tripleSuffix)
 """
     }
+
+    mutating func merge(_ other: Platform) {
+        minimumOsVersion = other.minimumOsVersion
+    }
```
cgrindel, line 53: Do we need a check to ensure that the platforms are mergeable? Can they only merge if they are compatible?

brentleyjones: The Starlark code that is determining target merges is only using direct dependencies, so we are trusting that is right.

brentleyjones: Given that, I think I can just take the platform entirely.

cgrindel (approved):

PR #841, opened 2022-07-29
Don't merge targets with mismatching deployment os versions

brentleyjones:
Also add an example of this.

This is needed, otherwise the code will be compiled under the `minimum_deployment_os_version` instead of `minimum_os_version`.


brentleyjones:
There might be a way to do this correctly, but using the compilation version for the build setting, and adjusting the Info.plist.

brentleyjones:
#843

PR #843, opened 2022-07-29
Correctly set `*_DEPLOYMENT_TARGET`

brentleyjones:
This should be set to the compilation minimum os version. The actual _deployment_ version is set in the `Info.plist`, which is handled by Bazel.


brentleyjones:
@cgrindel more of an FYI for you, given our convo in the other PR.

PR #855, opened 2022-08-01
Update `resolveTargetIDs()` to check for top-level dependency before other logic

cgrindel:
Related to #573, #840.

- Add `isTopLevel` property to `PBXProductType` and `PBXTarget`.
- Update `XcodeScheme.resolveTargetIDs()` to check for top-level targets using the previously mentioned `isTopLevel` properties.
- Update `XcodeScheme.resolveTargetIDs()` to resolve non-top-level targets using the following logic:
  1. Top-level transitive dependency with a matching label
  2. A target with a matching label and uses one of the top-level platforms
  3. First target from a platform-ordered list
- Add `pbxTargetAndKey()` to `TargetResolver`.
- Plumb `xcodeprojBazelLabel` into `resolveTargetIDs()` to provide a user-friendly error message.


tools/generator/src/Extensions/PBXProductType+Extensions.swift
```diff
@@ -292,4 +292,8 @@ extension PBXProductType {
             return true
         }
     }
+
+    var isTopLevel: Bool {
+        return isApplication || isTestBundle
```
brentleyjones, line 297: This doesn't include command line tools. Should it be `isLaunchable || isTestBundle`?

tools/generator/src/Extensions/XcodeScheme+Extensions.swift
```diff
@@ -25,154 +25,159 @@ extension XcodeScheme {
...
-            results[target.label] = targetInfo
-        }
-
-        return results
+extension XcodeScheme.LabelTargetInfo {
+    func firstCompatibleWith<Platforms: Sequence>(
+        anyOf platforms: Platforms
+    ) -> XcodeScheme.TargetWithID? where Platforms.Element == Platform {
+        let uniquePlatforms = Set(platforms)
+        return inPlatformOrder.first { uniquePlatforms.contains($0.target.platform) }
```
brentleyjones, line 161: Just confirming, does this pick the "best" from the topLevelPlatforms? Say we have tvOS and iOS top level targets and a library that is a dependency of both, will this always pick the iOS version of the library?

tools/generator/src/Extensions/PBXProductType+Extensions.swift
cgrindel: Good catch. I will update it.

tools/generator/src/Extensions/XcodeScheme+Extensions.swift
```diff
@@ -25,154 +25,159 @@ extension XcodeScheme {
...
+            let resolvedTargetID: TargetID
+            if let targetID = targets.firstTargetID(
+                under: topLevelTargetIDs,
+                where: { $0.label == label }
+            ) {
+                resolvedTargetID = targetID
+            } else if let targetWithID = labelTargetInfo.firstCompatibleWith(
+                anyOf: topLevelPlatforms
+            ) {
+                resolvedTargetID = targetWithID.id
```
cgrindel, line 94: @brentleyjones This is where we pass in the top-level platforms to `firstCompatibleWith()`.

tools/generator/src/Extensions/XcodeScheme+Extensions.swift
brentleyjones: I now see the `inPlatformOrder` 👍.

brentleyjones (approved):

tools/generator/src/Extensions/XcodeScheme+Extensions.swift
cgrindel: Yes. I added a comment where we pass in the top-level platforms. The `inPlatformOrder` array maintains the list of targets in `Platform` order. Hence, for your example, the first one that we find should be iOS. Also, #856 introduces a better compatibility check to evaluate the minimum OS version as a less-than-or-equal-to.

PR #856, opened 2022-08-02
Add compatibility check for `Platform`

cgrindel:
Related to #573.

- Add `compatibility(with:)` to `Platform`.
- Add `compatibleWith(anyOf:)` to `Platform`.
- Add `SemanticVersion` struct for `minimumOsVersion` comparison.


tools/generator/src/Extensions/Platform+Extensions.swift
```diff
@@ -0,0 +1,62 @@
+// MARK: `Platform.OSVersion.semanticVersion`
+
+extension Platform.OSVersion {
+    var semanticVersion: SemanticVersion? {
```
brentleyjones, line 4: Can we precompute and assign this instead of recomputing it each time? Similar to how I precomputed the `fullVersion`?

cgrindel: Definitely. I wasn't sure if you wanted to add more precomputed properties to the struct.

brentleyjones: I left a comment in the PR that introduced `Platform.OSVersion` that I should probably go the extra distance and extract out the components themselves, but the string based comparison was doing an okay job for now. We can probably even merge this all together into one type and update `Platform.OSVersion`'s `Comparable` conformance.

cgrindel: OK. I will do it in two PRs. I would like to land the `compatibility()` implementation. I will follow it up with a refactor of `Platform.OSVersion`.

cgrindel:
I am closing this PR and will open a new one after rebasing.

tools/generator/src/Extensions/Platform+Extensions.swift
cgrindel: @brentleyjones What do you think about converting `Platform.OSVersion` to an enum? It would have two values `raw(String)` and `semanticVersion(SemanticVersion)`. The `Comparable` implementation would use the existing `<` for `raw(String)` and the `semanticVersion(SemanticVersion)`. If comparing a `raw(String)` and a `semanticVersion(SemanticVersion)`, we would have the `semanticVersion(SemanticVersion)` be less than the raw version.

brentleyjones: It should always be a `SemanticVersion`, and if we fail to parse it as such I think we should throw an error instead.

cgrindel: OK. The enum idea will not work due to [this behavior](https://github.com/buildbuddy-io/rules_xcodeproj/pull/859#discussion_r935856683).

PR #859, opened 2022-08-02
Add compatibility check for `Platform`

cgrindel:
Related to #573.

- Add `compatibility(with:)` to `Platform`.
- Add `compatibleWith(anyOf:)` to `Platform`.
- Add `SemanticVersion` struct for `minimumOsVersion` comparison.
- Remove `fullVersion from `Platform.OSVersion`.


tools/generator/src/DTO/Platform.swift
```diff
@@ -19,14 +19,16 @@ struct Platform: Equatable, Hashable, Decodable {
...
 
         init?(rawValue: String) {
             self.rawValue = rawValue
 
             var components = rawValue.split(separator: ".")
             let componentCount = components.count
             guard componentCount < 3 else {
-                self.fullVersion = rawValue
+                fullVersion = rawValue
+                semanticVersion = nil
```
brentleyjones, line 31: This should still be set. This guard is for when we don't need to adjust `fullVersion`.

tools/generator/src/Generator/SemanticVersion.swift
```diff
@@ -0,0 +1,51 @@
...
+
+// Inspired by https://gist.github.com/mjdescy/a805b5b4c49ed79fb240d3886815d5a2
+struct SemanticVersion: Equatable {
+    public let major: Int
+    public let minor: Int
+    public let patch: Int
+
+    public init?(version: String) {
+        let components = version.split(separator: ".")
+        guard components.count == 3 else { return nil }
```
brentleyjones, line 11: If the count is less than 3, we can assume the other parts are `0`. And if it's more than 3, do we care?

tools/generator/src/DTO/Platform.swift
```diff
@@ -19,14 +19,16 @@ struct Platform: Equatable, Hashable, Decodable {
     struct OSVersion: RawRepresentable, Hashable, Decodable {
         let rawValue: String
         let fullVersion: String
+        let semanticVersion: SemanticVersion?
```
brentleyjones, line 22: Because of how `fullVersion` is constructed, thus should always be non-nil. We should surface an error if it's not.

cgrindel: Then, why do we keep `rawValue` around?

brentleyjones: Because that is the user-facing value set, so we show that in disambiguated target names.

tools/generator/src/Generator/SemanticVersion.swift
cgrindel: I can move the logic that we currently have in `Platform.OSVersion` to here. I was relying on the current `OSVersion` logic to have already executed. Perhaps, I can cover this in the refactor that we discussed [here](https://github.com/buildbuddy-io/rules_xcodeproj/pull/856#discussion_r935855002).

tools/generator/src/DTO/Platform.swift
brentleyjones: Mainly it was because it comes through as `15.0` instead of `15.0.0` and I wanted to show that, but we can have logic to do the same thing with `SemanticVersion` printing.

tools/generator/src/DTO/Platform.swift
cgrindel: Fixed

tools/generator/src/Generator/SemanticVersion.swift
cgrindel: Done

tools/generator/src/DTO/Platform.swift
cgrindel: I have made it required.

brentleyjones (approved):

PR #862, opened 2022-08-02
Limit files generated when using custom schemes

brentleyjones:
Before this change `BazelDependencies` would generate files for all of the targets in a scheme, even when only a subset of targets were being built. Now we leverage the cached data in `XCBuildData` to gain insight into what targets are actually being built, allowing us to only build what is needed.

Sadly this cached data doesn't say which exact target is being built as part of Index Build, so we still have to rely on our off-by-one logic there. Also of note, SwiftUI Previews build every target in a scheme, and this behavior is preserved with this method.


brentleyjones:
@jpsim I've added you for review to make sure this doesn't break normal schemes for you. I did a lot of testing, and everything seems good, but this was a tricky implementation.

xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/python3
+
+import glob
+import json
+import os
+import sys
+import time
+import traceback
+
+def _calculate_build_request_file(objroot):
```
brentleyjones, line 10: I'll admit ahead of time that there should be some unit tests for this file. Maybe before 1.0 we can get to that, but for now I'm relying on the fact that it does the fallback logic in the case of errors, and people will really notice if it's not working.

cgrindel (approved):

xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
```diff
@@ -0,0 +1,147 @@
...
+            if wait_counter > time_to_wait:
+                break
+
+        with open(build_request_file, encoding = "utf-8") as f:
+            # Parse the build-request.json file
+            build_request = json.load(f)
+
+        buildable_target_ids = []
+        for target in build_request["configuredTargets"]:
+            buildable_target_ids.extend(guid_target_ids[target["guid"]])
```
brentleyjones, line 41: These buildRequest and PIFCache files probably shouldn't be depended on. But the same logic can be used to say we should use XCBBuildServiceProxyKit. Since we safely handle API change, or the files just not existing anymore, this is "safe" to rely on, and we can come up with another solution in the future if it ever breaks.

PR #871, opened 2022-08-03
Only list some flags once in `target.swift-extra-clang-flags`

brentleyjones:
This matches how lldb generates the `target.swift-extra-clang-flags` setting.


tools/generator/src/Generator/CreateFilesAndGroups.swift
```diff
@@ -568,11 +568,15 @@ extension Generator {
                     }
                     .uniqued()
 
+                var onceFilePaths: Set<FilePath> = []
+                var onceOtherFlags: Set<String> = []
                 let clangOtherArgs = try lldbContext.clang.map { clang in
```
jpsim, line 573: This map now performs mutations as a side-effect, so it might be clearer to change this to a loop since this is no longer purely functional. I don't feel strongly though.

jpsim (approved):

PR #872, opened 2022-08-03
Use absolute paths in `swift_debug_options.py`

brentleyjones:
lldb translates relative paths to be rooted on `/`, even with `platform settings -w ` set.


brentleyjones:
We only do this for lldb settings, to keep the size of the generated project down.

Issue #202, opened 2022-03-26
Bug: Project can fail to build if the `.xcodeproj` file is renamed


brentleyjones:
## Describe the bug

If the project uses internal rules_xcodeproj files, which will almost always be the be the case (e.g. uses generated files, compile stub, etc.), those paths are hardcoded to include the path to the `.xcodeproj` file. So if the project file is renamed, the build can fail.

## Expected behavior

We can probably get around this by creating a Run Script target that generates these files into Derived Data instead of creating them at project generation, and reference those in the project instead. I'm not 100% sure on that though, as `BUILT_PRODUCTS_DIR` is currently set differently for each target. Maybe we can get that back to a stable location, while still having targets produced into the `bazel-out/`-like directory structure.

If we can't fix this to "just work", we can have the build fail in a predictable way similar to one of the suggested fixes to #201. If `PROJECT_FILE_PATH` is different than expected we can require a regeneration of the project.

erikkerber:
I’m not sure of all the failure modes you’re trying to avoid, but I believe both rules_ios and Tulsi install these files into the Xcode project directory themselves (i.e. `My.xcodeproj/.bazel`). At least things like stubs, and are easily referenced with `PROJECT_DIR`

brentleyjones:
That's where they are now. There wasn't a `PBXSourceTree` that allowed me to easily reference them with `PBXFileReference`. I'll check to see if I missed one though.

brentleyjones:
So this is super interesting. If I use `PBXSourceTree.custom("PROJECT_FILE_PATH")`, the project navigator doesn't work (all the files are always red), but the project still compiles correctly. This is probably because `PROJECT_FILE_PATH` isn't set yet for the navigator, but by the time the build system gets to resolving the `PBXFileReference`, it is set.

Only the compile stub needs to be a `PBXFileReference`. The rest I can just reference via `PROJECT_FILE_PATH` regardless.

If I could find a way to have a `PBXFileReference` without it showing in the navigator, that would be best. Otherwise I think generating the compiler stub to `BUILT_PRODUCTS_DIR`, while leaving the rest in place, but no longer in the project navigator, is the best path forward.

brentleyjones:
I'm thinking the best we will be able to do for this is detect it in `BazelDependencies`, comparing the hardcoded value used in the Project navigator to `PROJECT_FILE_PATH` and bailing early. As it stands, any renamed projects with external dependencies or generated files will fail to build.

[...]

Issue #403, opened 2022-05-04
Feature Request: Support index-while-building when building with Bazel


brentleyjones:
When using the `swift.index_while_building` feature, Bazel will produce index stores while building. We should enable that feature by default in `_xcodeproj_transition` (though once we have multiple configs, we should only enable it for `build`, not `index`) and:

- Add the outputs to the `build` output group
- Use [index-import](https://github.com/MobileNativeFoundation/index-import) to copy the index files into Xcode's index store
    - We will want to use `-import-output-file` to only copy over what's needed for each library
    - Though if using `swift.use_global_index_store`, the per-library index store should already be only what's needed? If so, we could pass that feature as well/instead
    - If we do this based on the need to run `stubs/swiftc.py`, then we can reduce the amount of copying further, as `stubs/swiftc.py` will only run when something has changed

brentleyjones:
With Hybrid BwX, and eventually Focused Projects, the need for this is going up.

brentleyjones:
Non-swift targets will probably need https://github.com/bazelbuild/bazel/issues/15983.

brentleyjones:
Without this there is no indexing data for unfocused targets. Even with that, I'm removing it as a blocker for 0.7.0, as there are performance considerations I want to deeply investigate, including being able to remove index-while-building from the Index Build and SwiftUI Previews builds.

brentleyjones:
https://github.com/buildbuddy-io/rules_xcodeproj/issues/1060

brentleyjones:
#1040 resolved this.

PR #880, opened 2022-08-04
Use actual paths for BAZEL_OUT and BAZEL_EXTERNAL

### LINKED ISSUE #261: Bug: Generated sources show up twice in index
## Describe the bug

CleanShot 2022-04-06 at 19 54 34@2x

The first one is the actual file. The second one is where it is in the Project navigator, which the "Bazel Generated Files" group is rooted on a symlink into Derived Data.

## Expected behavior

Only one instance of the generated files shows up.

## Versions

- rules_xcodeproj: 0.1.0

### LINKED ISSUE #858: Bug: Code completion does not work for external repositories
## Describe the bug

Code completion does not work for external repositories. It works for local sources however.

|Local Files|External Files|
|-|-|
|image|image|

## Reproduction steps

1. Clone https://github.com/jpsim/swiftlint-bazel-example
2. Run `bazel run :swiftlint_xcodeproj && xed .`
3. Build the SwiftLintFramework scheme
4. Type `visit` in the body of the `ForceCastRuleVisitor` located in the `ForceCastRule.swift` file then press escape to trigger the code completion modal

Following the same steps in the [`rules_xcodeproj` branch of SwiftLint](https://github.com/realm/SwiftLint/compare/rules_xcodeproj) works just fine.

## Expected behavior

Code completion should work for both local files and files from external repositories.

## Versions
- Bazel: 5.2.0
- rules_xcodeproj: 7cdff3362ea8fffae42097b9276ff8a6ffbee2fb
- rules_apple: 1.0.1
- rules_swift: 1.0.0

## Additional context

1. I see the same behavior with BwB and BwX
2. Files in a `local_repository` like [`swiftlint_extra_rules/ForbiddenVarRule.swift`](https://github.com/jpsim/swiftlint-bazel-example/blob/main/swiftlint_extra_rules/ForbiddenVarRule.swift) also don't have working code completion
3. _Some_ code completion still works, such as accessing members on an instance, like if I type `node.` like the following screenshot:

image

However, I still see less information than if it fully worked, such as this local files case that also shows documentation and more accurate type information:

image

brentleyjones:
Progress towards #202. Fixes #261. Fixes #858.

This partially reverts 0e6427003f150d9e0f36bc56402f9aea082e9602.

Making this change has numerous benefits:

- Xcode will now index targets that solely consist of generated or external files
- Generated and external sources no longer show up twice in code completion
- There is no longer a difference between "bazel-out" and "gen_dir" in the codebase, greatly simplifying that logic
- Project generation is faster as it doesn't have to invoke `xcodebuild` to determine `$OBJROOT`
- Makes progress towards moveable `.xcodeproj` bundles


shared.bazelrc
```diff
@@ -1,3 +1,5 @@
+startup --output_base=bazel-output-base
+
```
cgrindel, line 2: This will change the output_base for developers as well as CI. If we want this for CI, should we have a CI-specific bazelrc?

brentleyjones: We need it for developers as well, as we need the fixtures to match.

brentleyjones:
Workflows have an issue with `git clean`, so I won't push more changes here until it gets the proper approval so I can shepherd in the changes.

cgrindel (approved):

brentleyjones:
@jpsim Can you confirm this fixes your issue?

jpsim:
> Can you confirm this fixes your issue?

Checking now.

Is it expected to have significantly more stale file warnings in Xcode with this change?

For example with https://github.com/jpsim/swiftlint-bazel-example:

> Stale file '/Users/jsimard/src/swiftlint-bazel-example/SwiftLint.xcodeproj/rules_xcodeproj/links/external/StaticInternalSwiftSyntaxParser/lib_InternalSwiftSyntaxParser.xcframework/macos-arm64_x86_64/lib_InternalSwiftSyntaxParser.framework' is located outside of the allowed root paths.

image

jpsim:
I'm also seeing indexing errors

image

brentleyjones:
The file paths changed, similar to when a new starlark hash is created: https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/doc/faq.md#when-i-build-i-get-warnings-like-stale-file-projectxcodeprojrules_xcodeprojgen_dir-is-located-outside-of-the-allowed-root-paths-how-do-i-fix-this. Should only hit it once.

jpsim:
> Should only hit it once.

ok clean build does indeed make the stale file warnings go away.

I now get full auto-complete in external repositories (e.g. [Source/SwiftLintFramework/Rules/Idiomatic/ForceCastRule.swift](https://github.com/realm/SwiftLint/blob/master/Source/SwiftLintFramework/Rules/Idiomatic/ForceCastRule.swift)), but it's still missing from local repositories (e.g. [swiftlint_extra_rules/ForbiddenVarRule.swift](https://github.com/jpsim/swiftlint-bazel-example/blob/main/swiftlint_extra_rules/ForbiddenVarRule.swift)). So #858 is half fixed for my case.

brentleyjones:
So those indexing errors are because `swiftlint` doesn't compile in BwX mode. It's missing its `SWIFT_INCLUDE_PATHS` (which I will get a fix for in).

As for local repositories, the issue there is symlinks, just those inside of `external`. At this time I'm not sure how to address those, since if we changed their path in the Project navigator, the `target.source-map` won't be correct for them, so breakpoints won't work. I think maybe they will have to be handled with #403.

jpsim:
> So those indexing errors are because `swiftlint` doesn't compile in BwX mode. It's missing its `SWIFT_INCLUDE_PATHS` (which I will get a fix for in).

I'm using BwB though 🤔 

Or are you saying that indexing always uses BwX?

> As for local repositories, the issue there is symlinks, just those inside of `external`.

Thanks for investigating this. I'll go ahead and approve this PR then considering that local repositories will require a different solution.

Would you like me to file a new issue copying #858 but just for local repositories?

jpsim (approved):

brentleyjones:
> Or are you saying that indexing always uses BwX?

The Index Build step uses whichever mode, but the live indexing, which is what you showed a screenshot of, that calls `swiftc` directly, using all of the settings we have set. So if BwX doesn't compile, that will fail as well. Which is why it's so important for BwX to be able to compile everything.

> Would you like me to file a new issue copying https://github.com/buildbuddy-io/rules_xcodeproj/issues/858 but just for local repositories?

Yes please.

.bazelrc
```diff
@@ -1,7 +1,7 @@
 # To update these lines, execute
 # `bazel run @contrib_rules_bazel_integration_test//tools:update_deleted_packages`
-build --deleted_packages=examples/cc/external,examples/command_line/external,examples/ios_app,examples/ios_app/CoreUtilsObjC,examples/ios_app/Example,examples/ios_app/ExampleNestedResources,examples/ios_app/ExampleObjcTests,examples/ios_app/ExampleResources,examples/ios_app/ExampleTests,examples/ios_app/ExampleUITests,examples/ios_app/OnlyStructuredResources,examples/ios_app/TestingUtils,examples/ios_app/Utils,examples/ios_app/external,examples/ios_app/test/fixtures,examples/ios_app/test/fixtures/bwb.xcodeproj/rules_xcodeproj/links,examples/ios_app/test/fixtures/bwx.xcodeproj/rules_xcodeproj/links,App.xcodeproj/rules_xcodeproj/bazel,App.xcodeproj/rules_xcodeproj/links
```
brentleyjones, line 3: @cgrindel Seems like there should be a better way for the glob to ignore `bazel-output-base`. Also, it doesn't seem to work:
```
ERROR: infinite symlink expansion detected
[start of symlink chain]
/Users/administrator/buildbuddy/remote_build/dc1af18c-964f-41ec-b641-f925f65f273a/output-base/execroot/com_github_buildbuddy_io_rules_xcodeproj/bazel-out/darwin-fastbuild/bin/examples/ios_app_test.runfiles/com_github_buildbuddy_io_rules_xcodeproj
/Users/administrator/buildbuddy/remote_build/dc1af18c-964f-41ec-b641-f925f65f273a/output-base/execroot/com_github_buildbuddy_io_rules_xcodeproj/bazel-out/darwin-fastbuild/bin/examples/ios_app_test.runfiles/com_github_buildbuddy_io_rules_xcodeproj/examples/ios_app/bazel-output-base/external/com_github_buildbuddy_io_rules_xcodeproj
[end of symlink chain]
```

brentleyjones: I'll see if I can adjust where I set it to get around this.

brentleyjones:
I'll fix the bugs integration tests soon, probably by exposing a hint for what the workspace directory is.

brentleyjones:
@jpsim A rules_swift version that includes https://github.com/bazelbuild/rules_swift/commit/adcc2f1bc0eab7e8b047a5e2b7eec420c826f18d fixes the `swiftlint` indexing issue.

.bazelrc
brentleyjones: Got around it with some `.bazelignore` entires. It still has loads more entries than needed.

cgrindel: I have run into symlink issues after running a build manually in the child workspace and then running a build in the parent workspace. I have worked around the issue by running `bazel clean --expunge` in the child workspace.

For some background, the reason we use the `--deleted_packages` mechanism instead of `.bazelignore` is [lightly document here](https://github.com/bazel-contrib/rules_bazel_integration_test/blob/bf73bdb085525f5e6059e89af30a6b749f1f3802/README.md#4-configure-the-deleted-packages-for-the-parent-workspace). My recollection is `glob()` will not find files under ignored packages. The following is an excerpt from the doc:

> For glob() in the parent workspace to find files in the child workspaces, we need to use the deleted packages mechanism.

Is the issue discussed in this PR related to using an Xcode project in the child workspace?

brentleyjones: > Is the issue discussed in this PR related to using an Xcode project in the child workspace?

No, it was related to using an in-repo output_base. The only way around that was `.bazelignore`. 

The biggest issue though is that this `--deleted_packages` setting is massive now. I could manually adjust it (I think, to just list the ignore output base and it should cover child packages, right?), but then we lose the use of the tool to update it.

cgrindel: Gotcha. The `--deleted_packages` mechanism is very clunky. Unfortunately, it is not recursive. Each package needs to be listed. 😞

For now, we could write a script that adds the stuff that you have been doing manually. Alternatively, if the `ios_app` example is causing more trouble than it is worth, we could revert it to exist like the other examples.

brentleyjones: Currently Workflows don't support sub-directories, but once they do, I think I'll convert our integration test to just be a Workflow action rooted in the example directory, removing the need for `
rules_bazel_integration_test`.

PR #890, opened 2022-08-05
Implement best host logic when a host is not one of the top-level targets

cgrindel:
Related to #573.

- Implement `compatibleWith()` for a `Sequence` of `Platform` values.
- Move `HostInfo` resolution logic to `HostInfo` file.
- Add unit tests for host resolution logic.

### DIFF
--- a/tools/generator/test/Platform+ExtensionsTests.swift
+++ b/tools/generator/test/Platform+ExtensionsTests.swift
@@ -13,7 +13,7 @@ extension PlatformExtensionsTests {
     func test_compatibility_variantMismatch() throws {
         let platform = Platform(
             os: iOSarm64Platform.os,
-            variant: .tvOSDevice,
+            variant: .iOSSimulator,
             arch: iOSarm64Platform.arch,
             minimumOsVersion: iOSarm64Platform.minimumOsVersion
         )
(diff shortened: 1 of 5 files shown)


tools/generator/test/Platform+ExtensionsTests.swift
```diff
@@ -7,37 +7,36 @@ import XCTest
...
         let platform = Platform.device(os: .tvOS)
-        XCTAssertEqual(platform.compatibility(with: other), .osNotEqual)
+        XCTAssertEqual(platform.compatibility(with: iOSarm64Platform), .osNotEqual)
     }
 
     func test_compatibility_variantMismatch() throws {
         let platform = Platform(
-            os: other.os,
+            os: iOSarm64Platform.os,
             variant: .tvOSDevice,
```
brentleyjones, line 16: The variant doesn't match now.

cgrindel: It never did. The `other` was always `iOS`. For this test, I want `platform` to be identical except for the variant.

brentleyjones: I mean that os has to be a subset of variant. If `os` is something like `.iOS`, then `variant` has to then be `.iOSDevice` or `.iOSSimulator`, otherwise the data is garbage.

cgrindel: I switched it to `.iOSSimulator`.

brentleyjones (approved):

PR #901, opened 2022-08-09
Allow client to customize `buildFor` in custom Xcode schemes

cgrindel:
Related to #573.

- A client now provides `XcodeScheme.BuildTarget` values to `XcodeScheme.BuildAction`. The build target provides a means for specifying individual `buildFor` settings.
- Implemented `XcodeScheme.withDefautls`. This computed property returns a copy of the`XcodeScheme` with the appropriate defaults applied (e.g. default actions, relevant `buildFor` settings).
- Added validation checks to the initializers for `XcodeScheme`, `XcodeScheme.BuildAction`, and `XcodeScheme.TestAction`.


tools/generator/src/DTO/XcodeScheme.swift
```diff
@@ -20,11 +28,112 @@ struct XcodeScheme: Equatable, Decodable {
...
+            func enableBuildForValue(
+                _ label: BazelLabel,
+                _ keyPath: WritableKeyPath<XcodeScheme.BuildFor, XcodeScheme.BuildFor.Value>
+            ) throws {
+                do {
+                    try buildTargets[label, default: .init(label: label, buildFor: .init())]
+                        .buildFor[keyPath: keyPath]
+                        .merge(with: .enabled)
+                } catch XcodeScheme.BuildFor.Value.ValueError.incompatibleMerge {
+                    throw PreconditionError(message: """
```
brentleyjones, line 46: Just confirming this is not a user facing error. User facing errors are thrown a different way and use Starlark based verbiage.

cgrindel: It could be. Can you send a pointer to a user-facing error that you would like me to mimic?

cgrindel: @brentleyjones I should have mentioned you in my last response.

brentleyjones: The main difference is you need to throw a [UsageError](https://github.com/buildbuddy-io/rules_xcodeproj/blob/e233fd39125707cb6a88123271b6c9c53583dfd5/tools/generator/src/Errors.swift#L4). And the message has to be from the point of view of the BUILD file, mentioning what is wrong and how it can be fixed.

cgrindel: I will take a look at some examples in the repo and adjust this one.

cgrindel: I updated it to be a `UsageError`. Let me know what you think.

cgrindel:
@brentleyjones The integration test does not seem to be kicking-off in CI. Any thoughts?

tools/generator/src/DTO/XcodeScheme.swift
```diff
@@ -20,11 +28,112 @@ struct XcodeScheme: Equatable, Decodable {
...
+                _ label: BazelLabel,
+                _ keyPath: WritableKeyPath<XcodeScheme.BuildFor, XcodeScheme.BuildFor.Value>
+            ) throws {
+                do {
+                    try buildTargets[label, default: .init(label: label, buildFor: .init())]
+                        .buildFor[keyPath: keyPath]
+                        .merge(with: .enabled)
+                } catch XcodeScheme.BuildFor.Value.ValueError.incompatibleMerge {
+                    throw UsageError(message: """
+The buildFor value, "\(keyPath.stringValue)", for "\(label)" in the "\(name)" Xcode scheme was \
```
brentleyjones, line 47: We are exposing the value as `build_for` in Starlark, right?
```suggestion
The `build_for` value, "\(keyPath.stringValue)", for "\(label)" in the "\(name)" Xcode scheme was \
```

brentleyjones:
Let's see if it fixes itself after your next push. If not I'll debug.

tools/generator/src/DTO/XcodeScheme.swift
cgrindel: Done

brentleyjones (approved):
Shouldn't this PR have changed the generator scheme to not have `tests` have the `running` `buildFor`?

cgrindel:
> Shouldn't this PR have changed the generator scheme to not have `tests` have the `running` `buildFor`?

Good catch. I ran fixtures update and it did update the settings. Do we need to update the tests to detect changes in scheme files?

cgrindel:
@brentleyjones It looks like Integration Test is not reporting back to GitHub. Thoughts don't how to debug?

Also, I realized that I never update the Starlark to support the new build targets. I am creating a PR for that now.

brentleyjones:
> Do we need to update the tests to detect changes in scheme files?

They should detect changes already, as it simply diffs the whole bundle folder. If they aren't for some reason, yes, we need to fix that.

> It looks like Integration Test is not reporting back to GitHub. Thoughts don't how to debug?

It's a bug on the Workflows side. We are looking into it.

PR #905, opened 2022-08-10
Update Starlark to create `build_target` and `build_for`

cgrindel:
Related to #573.

- Added a custom `build_target` with a `build_for` to `//tools/generator:xcodeproj`. This tests that the value is written by Starlark code and read by the Swift code.

### DIFF
--- a/test/fixtures/generator/bwx_spec.json
+++ b/test/fixtures/generator/bwx_spec.json
@@ -13,7 +13,20 @@
     "configuration": "darwin_x86_64-fastbuild-ST-1b9bd654f600",
     "custom_xcode_schemes": [
         {
-            "build_action": null,
+            "build_action": {
+                "targets": [
+                    {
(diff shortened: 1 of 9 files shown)


test/fixtures/generator/bwx_spec.json
```diff
@@ -13,7 +13,20 @@
...
-            "build_action": null,
+            "build_action": {
+                "targets": [
+                    {
+                        "build_for": {
+                            "analyzing": "unspecified",
+                            "archiving": "enabled",
+                            "profiling": "unspecified",
+                            "running": "unspecified",
+                            "testing": "unspecified"
```
brentleyjones, line 24: Future enhancement: we could not encode when `unspecified` and adjust the Decodeable conformance to default it. Will save some space in these files (which is a goal of #237).

brentleyjones (approved):

PR #909, opened 2022-08-11
Add `archiving` and `analyzing` logic for `buildFor`

cgrindel:
Related to #573 

- Add `isEnabled` and `isDisabled` helpers to `XcodeScheme.BuildFor.Value`.
- Add `enableIfNotIDisabled()` to `XcodeScheme.BuildFor.Value`.
- Add logic to set `analyzing` for all targets.
- Add logic to set `archiving` for all targets with `running` enabled.

### DIFF
--- a/tools/generator/src/DTO/XcodeScheme.swift
+++ b/tools/generator/src/DTO/XcodeScheme.swift
@@ -39,22 +39,6 @@ extension XcodeScheme {
         get throws {
             var buildTargets = [BazelLabel: XcodeScheme.BuildTarget]()
 
-            // func enableBuildForValue(
-            //     _ label: BazelLabel,
-            //     _ keyPath: WritableKeyPath<XcodeScheme.BuildFor, XcodeScheme.BuildFor.Value>
-            // ) throws {
-            //     do {
-            //         try buildTargets[label, default: .init(label: label, buildFor: .init())]
-            //             .buildFor[keyPath: keyPath]
-            //             .merge(with: .enabled)
-            //     } catch XcodeScheme.BuildFor.Value.ValueError.incompatibleMerge {
-            //         throw UsageError(message: """
-// The `build_for` value, "\(keyPath.stringValue)", for "\(label)" in the "\(name)" Xcode scheme was \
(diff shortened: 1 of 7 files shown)


tools/generator/src/DTO/XcodeScheme.swift
```diff
@@ -29,25 +29,36 @@ No actions were provided for the scheme "\(name)".
...
-disabled, but the target is referenced in the scheme's \(keyPath.actionType) action.
-""")
+                var buildTarget = buildTargets[label, default: .init(label: label, buildFor: .init())]
+                switch mode {
+                case .merge:
+                    do {
+                        try buildTarget.buildFor[keyPath: keyPath].merge(with: .enabled)
+                    } catch XcodeScheme.BuildFor.Value.ValueError.incompatibleMerge {
+                        throw UsageError(message: """
+    The `build_for` value, "\(keyPath.stringValue)", for "\(label)" in the "\(name)" Xcode scheme was \
```
brentleyjones, line 54: Can you de-indent this?

brentleyjones (commented):
Should this update the generator scheme?

tools/generator/src/DTO/XcodeScheme.swift
cgrindel: It looks like it is at the right indent. The syntax for properties that throw requires an extra level of indent. I can switch it to a function if you would like.

cgrindel:
> Should this update the generator scheme?

Again, good catch. I rely on the fixture tests to yell at me. For some reason, they are not. I will try to look into this next week.

tools/generator/src/DTO/XcodeScheme.swift
brentleyjones: I mean the string having indent on the left.

brentleyjones:
I have a change that will cause them to be correctly picked up. No need to look into it, it will resolve itself with my next PR.

cgrindel:
@brentleyjones Repost from Slack:

> I am just looking at your post again. Did you want any target marked as running except tests to have `archiving` enabled? I implemented it as any marked as running.

brentleyjones (approved):

PR #915, opened 2022-08-16
Repro issue: Unable to build external `objc_library`

maxwellE:
I am unable to build an external `objc_library` target via the generated
xcodeproj, yet it can build on the
command line

Command line build:

```terminal
INFO: Analyzed target //Example:Example (4 packages loaded, 834 targets configured).
INFO: Found 1 target...
INFO: From MomCompile Example/Example-intermediates/Model.momd:
/Users/maxwellelliott/Development/rules_xcodeproj/examples/ios_app/bazel-output-base/sandbox/darwin-sandbox/53/execroot/__main__/Example/Model.xcdatamodeld/Model2.xcdatamodel:Entity2.parent: warning: Entity2.parent should have an inverse [2]
Target //Example:Example up-to-date:
  bazel-out/applebin_ios-ios_x86_64-fastbuild-ST-d63133e8454d/bin/Example/Example.app
  ```

  Xcode Build:

  ```terminal
  error: 'external/FXPageControl/FXPageControl/FXPageControl.h' file not found
  ```

PR #920, opened 2022-08-16
Add VFS mapping for `$SRCROOT` -> `$exec_root`

brentleyjones:
Closes #915.

This allows `external/` imports in CC code to resolve to `$BAZEL_EXTERNAL`. The reason this works, is that `$exec_root` will have symlinks to the folders and files in the root of workspace, meaning all project files are still found.

Added an example as well to `examples/ios_app`.

Co-authored-by: Maxwell Elliott <elliott.432@buckeyemail.osu.edu>


maxwellE (approved):
Thank you!

brentleyjones:
I'm also going to try a different approach, where I change the `.` header search path to be the execroot, when we have bazel dependencies. That way we would have one fewer search path.

PR #925, opened 2022-08-16
Add `{device,simulator}_only_targets` attributes to `device_and_simulator`

### LINKED ISSUE #924: Feature Request: Ability to omit the need for profiles for some targets in `device_and_simulator`
Currently if a test bundle added to `device_and_simulator` depends on a test host, project generation will fail if that test host (i.e. `ios_application`) does not have a provisioning profile set.

Many times tests are only ever meant to run on simulators, and managing App IDs and profiles for light wrapper applications is overkill.

Being able to pass all targets through `device_and_simulator`, without requiring profiles for all apps, would help ergonomics.

brentleyjones:
Fixes #924.


erikkerber (approved):
A half-way 👍🏼 from me.

This did what it advertised with the Xcode project. I am unable to build due to an unrelated bug I am assuming is on our side. I'll follow up if I resolve that and get these tests building E2E.

PR #900, opened 2022-08-09
Fix indexing of `local_repository` targets

brentleyjones:
Fixes #888.

If an external repository is a symlink (such as the case with `local_repository`) we resolve the symlink for the Project navigator group. We also include additional entries in `bazel.lldbinit` to ensure breakpoints continue work for them.

PR #929, opened 2022-08-17
Resolve `new_local_repository` symlinks as well

brentleyjones:
Similar to #900.


tools/generator/src/Generator/CreateFilesAndGroups.swift
```diff
@@ -214,6 +254,8 @@ extension Generator {
             // e.g. resources/App.strings
             let groupFilePath = localizedContainerFilePath.parent() + fileName
 
+            // TODO: Use `resolveFilePath`?
```
BalestraPatrick, line 257: leftover or a valid TODO?

BalestraPatrick (approved):

tools/generator/src/Generator/CreateFilesAndGroups.swift
brentleyjones: Valid. I couldn't find an easy way, fast, to do it because of the odd structure. Since it's a leaf, and not a source file, and most likely not at the root of a `new_local_repository`, I don't think it will matter to most people. Leaving the TODO to remind a future person about it though, in case they want to fix it.

PR #932, opened 2022-08-17
Ensure application icons are specified correctly in Xcode project

cgrindel:
Related to #949.

Goal: Xcode finds the application icon and uses it for Xcode schemes.

### DIFF
--- a/xcodeproj/internal/default_automatic_target_processing_aspect.bzl
+++ b/xcodeproj/internal/default_automatic_target_processing_aspect.bzl
@@ -13,8 +13,8 @@ load(":providers.bzl", "XcodeProjAutomaticTargetProcessingInfo", "target_type")
 _UNSUPPORTED_SRCS_EXTENSIONS = {
     "a": True,
     "lo": True,
-    "so": True,
     "o": True,
+    "so": True,
 }
 
 def _get_target_type(*, target):
@@ -59,6 +59,7 @@ def _default_automatic_target_processing_aspect_impl(target, ctx):
     else:
         srcs = []
 
+    app_icons = None
     bundle_id = None
     codesignopts = None
     entitlements = None
@@ -104,6 +105,8 @@ def _default_automatic_target_processing_aspect_impl(target, ctx):
             xcode_targets["test_host"] = [target_type.compile]
         if "app_clips" in attrs:
             xcode_targets["app_clips"] = [target_type.compile]
+        if "app_icons" in attrs:
+            app_icons = "app_icons"
         if "codesignopts" in attrs:
             codesignopts = "codesignopts"
         if "entitlements" in attrs:
@@ -174,6 +177,7 @@ def _default_automatic_target_processing_aspect_impl(target, ctx):
             launchdplists = launchdplists,
             entitlements = entitlements,
             bazel_build_mode_error = bazel_build_mode_error,
+            app_icons = app_icons,
         ),
     ]
 
--- a/xcodeproj/internal/default_automatic_target_processing_aspect.bzl
+++ b/xcodeproj/internal/default_automatic_target_processing_aspect.bzl
@@ -98,6 +98,9 @@ def _default_automatic_target_processing_aspect_impl(target, ctx):
         infoplists = ["infoplists"]
         should_generate_target = False
     elif AppleBundleInfo in target:
+        # DEBUG BEGIN
+        print("*** CHUCK AppleBundleInfo target.label: ", target.label)
+        # DEBUG END
         xcode_targets = {
             "deps": [target_type.compile, None],
         }
@@ -106,6 +109,9 @@ def _default_automatic_target_processing_aspect_impl(target, ctx):
         if "app_clips" in attrs:
             xcode_targets["app_clips"] = [target_type.compile]
         if "app_icons" in attrs:
+            # DEBUG BEGIN
+            print("*** CHUCK Has app_icons")
+            # DEBUG END
             app_icons = "app_icons"
         if "codesignopts" in attrs:
             codesignopts = "codesignopts"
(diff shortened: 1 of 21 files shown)


xcodeproj/internal/top_level_targets.bzl
```diff
@@ -222,6 +237,13 @@ def process_top_level_target(
     is_swift = SwiftInfo in target
     swift_info = target[SwiftInfo] if is_swift else None
 
+    # DEBUG BEGIN
+    if AppleResourceInfo in target:
+        apple_resource_info = target[AppleResourceInfo]
+        build_settings["ASSETCATALOG_COMPILER_APPICON_NAME"] = "AppIcon"
```
cgrindel, line 243: @brentleyjones This is the magic setting that sets the `App Icon Source` in Xcode 13.4.1. 

Here are some questions.

### App Icon Name

As you can see, I hardcoded the name, `AppIcon`. I am not sure the best way to identify the icon set to use based upon what is available in the `rules_apple` providers. It looks like I can reach into `AppleResourceInfo.asset_catalogs` and look for entries that have a `.appiconset` suffix. Then, I could just pick the one named `AppIcon` or the first one. This feels pretty fragile to me.

### Xcode Version

I think that this is a recent change to Xcode (i.e., post 12.x). Do we need to consider adding a target Xcode version to `xcodeproj_rule`?

brentleyjones: > It looks like I can reach into `AppleResourceInfo.asset_catalogs` and look for entries that have a `.appiconset` suffix. Then, I could just pick the one named `AppIcon` or the first one. This feels pretty fragile to me.

Instead, let's add `app_icons` to `XcodeProjAutomaticTargetProcessingInfo` (defaulting it to `app_icons` for bundle providing targets) and look at the `app_icons` attribute. That has validation that makes sure it's actually app icons (sticker packs use `.stickersiconset` for example and tvOS apps use `.brandassets`) and it enforces that a single directory is used: https://github.com/bazelbuild/rules_apple/blob/efbb206cacdde114893772d0cef7c2c868546a36/apple/internal/resource_actions/actool.bzl#L126-L131

> I think that this is a recent change to Xcode (i.e., post 12.x). Do we need to consider adding a target Xcode version to `xcodeproj_rule`?

No. `ASSETCATALOG_COMPILER_APPICON_NAME` has been around for a long time, so we are safe just setting it.

xcodeproj/internal/default_automatic_target_processing_aspect.bzl
```diff
@@ -174,6 +177,7 @@ def _default_automatic_target_processing_aspect_impl(target, ctx):
             launchdplists = launchdplists,
             entitlements = entitlements,
             bazel_build_mode_error = bazel_build_mode_error,
+            app_icons = app_icons,
```
brentleyjones, line 180: We should have done it earlier, but can you alphabetize this whole constructor call?

xcodeproj/internal/top_level_targets.bzl
```diff
@@ -436,6 +436,10 @@ def process_top_level_target(
         transitive_infos = deps_infos,
     )
 
+    app_icon_name = _get_app_icon_name(ctx, automatic_target_info)
+    if app_icon_name != None:
+        build_settings["ASSETCATALOG_COMPILER_APPICON_NAME"] = app_icon_name
```
brentleyjones, line 441:
```suggestion
    set_if_true(
        build_settings,
        "ASSETCATALOG_COMPILER_APPICON_NAME",
         _get_app_icon_name(ctx, automatic_target_info),
    )
```

xcodeproj/internal/top_level_targets.bzl
```diff
@@ -475,3 +479,35 @@ def process_top_level_target(
...
+
+    Args:
+        ctx: The aspect context.
+        automatic_target_info: The `XcodeProjAutomaticTargetProcessingInfo` for
+            `target`.
+
+    Returns:
+        The application icon name, if found. Otherwise, None.
+    """
+    if automatic_target_info.app_icons == None:
```
brentleyjones, line 494:
```suggestion
    if not automatic_target_info.app_icons:
```

xcodeproj/internal/top_level_targets.bzl
```diff
@@ -475,3 +479,35 @@ def process_top_level_target(
...
+        return None
+
+    app_icons = getattr(ctx.rule.attr, automatic_target_info.app_icons, None)
+    if app_icons == None:
```
brentleyjones, line 498:
```suggestion
    if not app_icons:
```

xcodeproj/internal/top_level_targets.bzl
```diff
@@ -475,3 +479,35 @@ def process_top_level_target(
...
+        return None
+
+    for target in app_icons:
+        for file in target.files.to_list():
+            suffix_idx = file.short_path.find(".appiconset")
```
brentleyjones, line 503: Do we need to support the other bundle types as well, such as `.brandassets` (tvOS)?

BalestraPatrick:
@cgrindel Is this supposed to be working in BwB? I only see it working in BwX locally. That's likely because BwB doesn't have the "Copy Bundle Resources" step?

brentleyjones:
Setting `CFBundleIconFile` to a specific file, in our Xcode-specific Info.plist, works: https://github.com/buildbuddy-io/rules_xcodeproj/blob/111cd6427447a3ec9f7e1ed9539fb0b416bb141f/xcodeproj/internal/info_plists.bzl#L30

```
	<key>CFBundleIconFile</key>
	<string>Example/Assets.xcassets/AppIcon.appiconset/ios_store_icon.png</string>
```

After the Info.plist is generated, re-opening Xcode shows the scheme.

I would only do this in BwB though, as it's a hack to get that working, and we want the BwX Info.plist to stay pristine. So, I would do this for BwB only _in addition_ to what you've already done here.

cgrindel:
Having just added support for the tvOS assets, all seems to be working for BwX projects. I will look into the BwB projects.

xcodeproj/internal/top_level_targets.bzl
cgrindel: I added `.brandassets` support.

cgrindel:
@BalestraPatrick @brentleyjones I created #949 to track the implementation of the scheme icons. I will fix icons not appearing in BwB projects in a follow up PR.

brentleyjones (approved):

Issue #214, opened 2022-03-28
Provide API documentation


brentleyjones:
We should use [stardoc](https://github.com/bazelbuild/stardoc) to generate API documentation for the rules.

brentleyjones:
@BalestraPatrick @jpsim How valuable do you think it would be to add developer documentation for the generator, using DocC?

jpsim:
> How valuable do you think it would be to add developer documentation for the generator, using DocC?

Probably not hugely beneficial since the generator is only meant to be used by rules_xcodeproj, and not broadly by external consumers.

brentleyjones:
I think this is done. We can continue to improve the documentation we have now, but we do have it being generated.

PR #936, opened 2022-08-18
Add stardoc to generate rules docs

BalestraPatrick:
Addresses #214.

This PR adds the basic stardoc tooling in order to generate documentation for the public rules. This will allow to better understand the API that we provide. Some follow-up PRs might be needed to get more accurate documentation for certain rules that aren't yet properly documented.

Most of the tooling was inspired by the [`rules_apple`](https://github.com/bazelbuild/rules_apple/tree/master/doc/BUILD.bazel) implementation.

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -50,3 +50,14 @@ actions:
           - "*"
     bazel_commands:
       - "run --config=workflows //:buildifier.check"
+
+  - name: Docs Diff Test
+    triggers:
+      push:
+        branches:
+          - "main"
+      pull_request:
+        branches:
+          - "*"
+    bazel_commands:
+      - "test --config=workflows //doc:diff_test"
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -51,7 +51,7 @@ actions:
     bazel_commands:
       - "run --config=workflows //:buildifier.check"
 
-  - name: Docs Diff Test
+  - name: Docs
     triggers:
       push:
         branches:
--- a/xcodeproj/BUILD
+++ b/xcodeproj/BUILD
@@ -1,3 +1,4 @@
+# Required for stardoc integration.
 exports_files([
     "experimental.bzl",
     "xcodeproj.bzl",
--- a/xcodeproj/BUILD
+++ b/xcodeproj/BUILD
@@ -1,0 +1,4 @@
+exports_files([
+    "experimental.bzl",
+    "xcodeproj.bzl",
+])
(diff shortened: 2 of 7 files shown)


xcodeproj/BUILD
```diff
@@ -0,0 +1,4 @@
+exports_files([
```
BalestraPatrick, line 1: This seems needed, otherwise I get:
```
rules_xcodeproj/doc/BUILD.bazel:11:12: no such target '//xcodeproj:experimental.bzl': target 'experimental.bzl' not declared in package 'xcodeproj'; however, a source file of this name exists.  (Perhaps add 'exports_files(["experimental.bzl"])' to xcodeproj/BUILD?) defined by /Users/patrickb/src/rules_xcodeproj/xcodeproj/BUILD and referenced by '//doc:experimental_doc'
```

brentleyjones: Can we add a comment above it saying it's for Stardoc?

xcodeproj/internal/BUILD
```diff
@@ -3,6 +3,31 @@ load(
     "bool_setting",
     "string_setting",
 )
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
```
brentleyjones, line 6: Can you move this above the other load statement to keep it alphabetical (root directory versus a subdirectory)?

xcodeproj/BUILD
BalestraPatrick: Done

brentleyjones:
Can you add a Workflow action to `buildbuddy.yaml` to check that the docs are up-to-date?

xcodeproj/internal/BUILD
BalestraPatrick: Done

BalestraPatrick:
@brentleyjones I added a separate action, but I think this could fit well in the existing `Test` action (since it's a `test_suite`). Thoughts?

brentleyjones:
I like it being separate, so it can fail faster. The other one might be slow because of UI tests.

buildbuddy.yaml
```diff
@@ -50,3 +50,14 @@ actions:
           - "*"
     bazel_commands:
       - "run --config=workflows //:buildifier.check"
+
+  - name: Docs Diff Test
```
brentleyjones, line 54:
```suggestion
  - name: Docs
```

brentleyjones (approved):

brentleyjones:
Thanks for this @BalestraPatrick!

Issue #906, opened 2022-08-11
Feature Request: Set runtime env vars for tests


sebastianv1:
Test targets can use the `env` attribute to supply environment variables for tests at runtime; however, these variables are not propagated when running the test target in BwB.

One potential solution is creating an `env` attribute on custom schemes for the test action similar to what we've done with the launch_action attributes. This would allow us to share the env var dict in the BUILD file with the test target and custom scheme.

## Tasks
- [x] #940 
- [x] #942 
- [x] Fix `Expand Variables Based On` setting when specifying test env vars/args
  - [x] #951 
  - [x] #956 
- [ ] Propagate env variables from `XXX_unit_test`. ([Request](https://github.com/buildbuddy-io/rules_xcodeproj/issues/906#issuecomment-1228751832))

tinder-maxwellelliott:
I am not seeing a `ios_unit_test`s automatically created schemes containing the target's `env` variables. I think we need to fix this for auto creation

cgrindel:
> I am not seeing a `ios_unit_test`s automatically created schemes containing the target's `env` variables. I think we need to fix this for auto creation

@tinder-maxwellelliott Just to confirm. You would like to see values from [this env attribute](https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-ios.md#ios_unit_test-env) propagated to automatic schemes? Propagating to automatic schemes seems very reasonable. 

What do we want to do about custom schemes? Propagating to custom schemes could be useful, but could be confusing, as well. Perhaps, we should add an attribute to `xcode_schemes.test_action()` that dictates whether env variables should be propagated. If we do that, then propagating for automatic schemes would be a no-brainer. We would just need to set the argument.

tinder-maxwellelliott:
> What do we want to do about custom schemes? Propagating to custom schemes could be useful, but could be confusing, as well

I feel that out of the box its intuitive if iOS_unit_test's env is automatically applied, custom schemes would be used to override those default values.

mattdornfeld:
Is there an estimate for when this feature request might be completed? It would be very nice to not have to manually set env vars in XCode to get tests to pass :)

brentleyjones:
Which part are you still needing? Values set via `env` attribute on the rule, passed in with `--test_env`, both, or other?

There is some background work happening on this, but we can probably bump the priority. cc: @maxwellE.

maxwellE:
[...]

PR #942, opened 2022-08-18
Ensure custom command line args and env vars are used in test actions

cgrindel:
Related to #906.

- Fixed logic for all schemes to ensure that `buildMode` specific environment variables (e.g., `bazel`) are included in launch and test actions.
- Add sort and merge operations for `XCScheme.EnvironmentVariable` `Sequence` values. It applies a precedence for overriding environment variables and ensures that the variable lists are consistent for equality and diff purposes.


test/fixtures/generator/bwx.xcodeproj/xcshareddata/xcschemes/generator.xcscheme
```diff
@@ -91,7 +91,7 @@
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
       customLLDBInitFile = "$(BAZEL_LLDB_INIT)"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "NO">
```
brentleyjones, line 94: What is the affect of this? Does this mean the `BAZEL_WORKSPACE_DIRECTORY` env isn't passed anymore? If so, we need that set explicitly for https://github.com/bazelbuild/apple_support/pull/128 to work.

cgrindel: This unchecks the box in the Xcode dialog allowing the custom args/env vars to be used.

I will update the code to ensure that the bazel env vars are included.

cgrindel: If the magic Bazel env vars need to be set in every BwB scheme, then the current logic for custom schemes is not quite right. It only includes them if the a launch action is defined. 

@brentleyjones Can you confirm that we do want the magic Bazel env vars set in the launch action for every BwB scheme?

brentleyjones: We want them set in both the launch and test actions, yes. While Bazel doesn't set `BAZEL_WORKSPACE_DIRECTORY` by default for tests, we still need it for the above feature I liked.

brentleyjones: Looks like this one doesn't have the bazel magic env variables still?

cgrindel: This file is for BwX. I am seeing the magic env vars in the test action in the BwB version of the scheme.

brentleyjones (approved):

PR #944, opened 2022-08-19
Create xcodeproj.doc to import rules instead of macro to improve generated docs with stardoc

BalestraPatrick:
Borrowed this workaround to improve the `xcodeproj` docs from [rules_apple](https://github.com/bazelbuild/rules_apple/blob/d7b8f7d45c9bb0533aee2ed7410d111a0d060f71/apple/ios.doc.bzl#L17-L19).

### DIFF
--- a/doc/BUILD.bazel
+++ b/doc/BUILD.bazel
@@ -3,8 +3,8 @@ load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
 load("@io_bazel_stardoc//stardoc:stardoc.bzl", "stardoc")
 
 _DOC_SRCS = [
-    "xcodeproj",
     "experimental",
+    "xcodeproj.doc",
 ]
 
 [
--- a/xcodeproj/xcodeproj.doc.bzl
+++ b/xcodeproj/xcodeproj.doc.bzl
@@ -1,0 +1,15 @@
+"""Public rules, macros, and libraries."""
+
+# Re-export original rules rather than their wrapper macros
+# so that stardoc documents the rule attributes, not an opaque
+# **kwargs argument
+load(
+    "//xcodeproj/internal:providers.bzl",
+    _XcodeProjAutomaticTargetProcessingInfo = "XcodeProjAutomaticTargetProcessingInfo",
+)
+load("//xcodeproj/internal:xcode_schemes.bzl", _xcode_schemes = "xcode_schemes")
+load("//xcodeproj/internal:xcodeproj_rule.bzl", _xcodeproj = "xcodeproj")
+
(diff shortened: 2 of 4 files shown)


doc/BUILD.bazel
```diff
@@ -3,7 +3,7 @@ load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
 load("@io_bazel_stardoc//stardoc:stardoc.bzl", "stardoc")
 
 _DOC_SRCS = [
-    "xcodeproj",
+    "xcodeproj.doc",
     "experimental",
```
brentleyjones, line 7: Is this ordered this way for a reason?

xcodeproj/xcodeproj.bzl
```diff
@@ -1,5 +1,8 @@
 """Public rules, macros, and libraries."""
 
+# Re-export original rules rather than their wrapper macros
+# so that stardoc documents the rule attributes, not an opaque
+# **kwargs argument
```
brentleyjones, line 5: This should be in the `doc.bzl` file, right?

xcodeproj/xcodeproj.doc.bzl
```diff
@@ -0,0 +1,17 @@
+"""Public rules, macros, and libraries."""
+
+load(
+    "//xcodeproj/internal:providers.bzl",
+    _XcodeProjAutomaticTargetProcessingInfo = "XcodeProjAutomaticTargetProcessingInfo",
+)
+load("//xcodeproj/internal:xcode_schemes.bzl", _xcode_schemes = "xcode_schemes")
+load("//xcodeproj/internal:xcodeproj_rule.bzl", _xcodeproj = "xcodeproj")
+
+# Re-exporting providers
```
brentleyjones, line 10: We can leave all of the comments off in this file, making it one big blob like in rules_apple.

doc/BUILD.bazel
BalestraPatrick: I just thought `xcodeproj.doc` contains the "main" rules, but I can change it to alphabetical instead.

xcodeproj/xcodeproj.bzl
BalestraPatrick: Good catch

doc/BUILD.bazel
brentleyjones: Does it impact the generation in any way? If not then being alphabetical would be best.

BalestraPatrick: Doesn't seem like it. I changed it already in this PR and the diff was the same.

brentleyjones (approved):

PR #950, opened 2022-08-22
Add ability to specify extra Bazel flags

brentleyjones:


brentleyjones:
Draft simply because I want to merge the parent PR first. This is ready for review.

xcodeproj/internal/extra_flags_bazelrc.bzl
```diff
@@ -0,0 +1,82 @@
...
+        content = content,
+        setting = "_extra_common_flags",
+        config_suffix = "",
+    )
+    _process_extra_flags(
+        attr = ctx.attr,
+        content = content,
+        setting = "_extra_generator_flags",
+        config_suffix = "_generator",
+    )
```
brentleyjones, line 28: I forgot that this will cause issues with CI. I'll pull this one out to be in the runner template itself.

shared.bazelrc
```diff
@@ -27,6 +29,7 @@ build:workflows --remote_instance_name=buildbuddy-io/rules_xcodeproj/workflows
 build:workflows --color=yes
 build:workflows --terminal_columns=120
 build:workflows --disk_cache=
+build:workflows --@com_github_buildbuddy_io_rules_xcodeproj//xcodeproj:extra_generator_flags='--config=workflows --bes_backend= --bes_results_url='
```
brentleyjones, line 32: Since this only overrides `extra_generator_flags`, the fixtures will still have `--config=cache` in them.

@cgrindel @BalestraPatrick This means you need to use `--config=cache `when updating the fixtures. I'll update the guide.

xcodeproj/internal/xcodeproj_runner.bzl
```diff
@@ -15,7 +16,11 @@ def _xcodeproj_runner_impl(ctx):
         is_executable = True,
         substitutions = {
             "%bazel_path%": ctx.attr.bazel_path,
-            "%bazelrc%": ctx.file._bazelrc.path,
+            "%bazelrc%": ctx.file._bazelrc.short_path,
```
jpsim, line 19: Is `short_path` right for this one?

brentleyjones: I'll have to check. It was changed as part of something else that I reverted. It's probably wrong though :+1:.

jpsim (approved):
LGTM, one flag/question. Would be good to exercise this in one of the examples.

cgrindel (approved):

brentleyjones:
> Would be good to exercise this in one of the examples.

We already do for the base level config, it adds `--config=cache`. I'll add some to `examples/ios_app` for the other flags.

xcodeproj/internal/xcodeproj_runner.bzl
brentleyjones: I've used `short_path` for all other substitutions, so I'm leaving this for now.

PR #952, opened 2022-08-23
Move Bazel flags into `xcodeproj.bazelrc`

brentleyjones:
Besides for these flags being more discoverable now, it also removes `--features` from the `xcodeproj` transition, preventing the Starlark hash from changing when features are added or removed from the build.


brentleyjones:
Draft only because I don't want to merge #950 yet. Ready for review.

jpsim (approved):

cgrindel (approved):

Issue #959, opened 2022-08-23
Bug: Unable to omit build_for when creating build_target


erikkerber:
## Describe the bug

`build_for` defaults to `None` in `xcode_scheme.build_target`. This `None` value isn't converted into a default `build_for` value, resulting in a Codable error during generation:

```
ERROR: Internal precondition failure:
At codingPath [customXcodeSchemes, Index 0, buildAction, targets, Index 0, buildFor]: Cannot get keyed decoding container -- found null value instead.
```

## Reproduction steps

1. Create a custom scheme with a build target, omitting `build_for`:

```
        build_action = xcode_schemes.build_action([
            xcode_schemes.build_target(
                "//Foo:Bar
            ),
        ]),
```

This passes loading and analysis, but fails at project generation

## Expected behavior
Either:

- `build_for` defaults its values, per model_test.bzl
- A Starlark error is observed

## Versions
- Bazel: 5.3
- rules_xcodeproj: 739b8bb0
- rules_apple:
- rules_swift:

jpsim:
Seeing the same thing with Envoy Mobile if you want an easy repro: `./bazelw run :xcodeproj` from https://github.com/envoyproxy/envoy-mobile

```
ERROR: Internal precondition failure:
At codingPath [customXcodeSchemes, Index 4, buildAction, targets, Index 0]: Expected to decode Dictionary<String, Any> but found a string/data instead.
```

brentleyjones:
This is a slightly different issue, but another that should probably be addressed. I think we should accept a `String` or a `xcode_schemes.build_target` in the array inside of `xcode_schemes.build_action`. Thoughts @cgrindel?

cgrindel:
@erikkerber @jpsim I am going to fix the fact that the default causes a failure. However, I am curious how you ran into this. I would not expect `xcode_schemes.build_target()` to be used unless you were overriding the `build_for`. Are you adding build targets that are not referenced elsewhere in the scheme?

cgrindel:
> This is a slightly different issue, but another that should probably be addressed. I think we should accept a `String` or a `xcode_schemes.build_target` in the array inside of `xcode_schemes.build_action`. Thoughts @cgrindel?

@brentleyjones We can definitely add this ergonomic change. I am surprised that there are so many cases where someone is including build targets that are not referenced elsewhere in the scheme. Granted, I tend to think more about the BwB case.

brentleyjones:
[...]

PR #962, opened 2022-08-23
Ensure that the default for `build_for`, if not specified, is all enabled

cgrindel:
Related to #959.

### DIFF
--- a/test/fixtures/generator/bwb.xcodeproj/xcshareddata/xcschemes/generator.xcscheme
+++ b/test/fixtures/generator/bwb.xcodeproj/xcshareddata/xcschemes/generator.xcscheme
@@ -57,7 +57,7 @@
       </PreActions>
       <BuildActionEntries>
          <BuildActionEntry
-            buildForTesting = "NO"
+            buildForTesting = "YES"
             buildForRunning = "YES"
             buildForProfiling = "YES"
             buildForArchiving = "YES"
--- a/test/fixtures/generator/bwb.xcodeproj/xcshareddata/xcschemes/generator.xcscheme
+++ b/test/fixtures/generator/bwb.xcodeproj/xcshareddata/xcschemes/generator.xcscheme
@@ -57,7 +57,7 @@
       </PreActions>
       <BuildActionEntries>
          <BuildActionEntry
-            buildForTesting = "YES"
+            buildForTesting = "NO"
             buildForRunning = "YES"
             buildForProfiling = "YES"
             buildForArchiving = "YES"
--- a/test/fixtures/generator/bwb_spec.json
+++ b/test/fixtures/generator/bwb_spec.json
@@ -17,11 +17,11 @@
                 "targets": [
                     {
                         "build_for": {
-                            "analyzing": "unspecified",
+                            "analyzing": "enabled",
                             "archiving": "enabled",
-                            "profiling": "unspecified",
-                            "running": "unspecified",
-                            "testing": "unspecified"
+                            "profiling": "enabled",
+                            "running": "enabled",
+                            "testing": "enabled"
                         },
                         "label": "//tools/generator:generator"
                     }
--- a/test/fixtures/generator/bwb_spec.json
+++ b/test/fixtures/generator/bwb_spec.json
@@ -17,11 +17,11 @@
                 "targets": [
                     {
                         "build_for": {
-                            "analyzing": "enabled",
+                            "analyzing": "unspecified",
                             "archiving": "enabled",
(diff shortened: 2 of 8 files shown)


test/fixtures/generator/bwb.xcodeproj/xcshareddata/xcschemes/generator.xcscheme
```diff
@@ -57,7 +57,7 @@
       </PreActions>
       <BuildActionEntries>
          <BuildActionEntry
-            buildForTesting = "NO"
+            buildForTesting = "YES"
```
brentleyjones, line 60: This is wrong. The previous way is the expected default. It should only default to all on for ones that don't have a test action or launch action implying the buildFor.

test/fixtures/generator/bwb_spec.json
```diff
@@ -17,11 +17,11 @@
                 "targets": [
                     {
                         "build_for": {
-                            "analyzing": "unspecified",
```
brentleyjones, line 20: Unspecified is correct.

test/fixtures/generator/bwb.xcodeproj/xcshareddata/xcschemes/generator.xcscheme
cgrindel: I changed the scheme to replicate the missing `build_for` value. Previously, we only specified a value for `archiving`. The rest of the values were unspecified. Hence, the merging resulted in `buildForTesting` being `NO`. The updated scheme does not specify a `build_for` value. So, I set the default to all enabled, based upon this line from our previous conversation:
> buildActionInfo can define which ones (in BUILD files), but defaults to all (like today)

[Comment from FR](https://github.com/buildbuddy-io/rules_xcodeproj/issues/573#issuecomment-1203310073)

I think that the disconnect is that the `tools/generator` scheme does not need to specify any build targets. I added it as an integration test of sorts. Perhaps, it would make more sense to remove it from `tools/generator` and create some schemes in `examples/multiplatform` or `examples/ios_app` that had some interesting combinations.

For now, it might be best for me to put the `tools/generator` scheme back to its previous state. What do you think?

For clients that really want to specify a build target in a scheme, I think the all enabled may still make sense. The only reason that they should add the build target is if the target is not referenced anywhere else in the scheme.

brentleyjones: I think it changed since I made that comment. I can use an empty `build_for()` and all the values will be `unspecified`, which should be the same as not passing a `build_for()` at all (these two are semantically the same to me, and shouldn't differ). And if they are all `unspecified`, I believe we turn them all on in the generator it nothing else ends up implying something different, right?

brentleyjones: The main issue is the semantic difference between none and an empty `build_for()`. I'm fine with it defaulting to all enabled, as it makes sense.

brentleyjones: > My understanding was that if a client specifies a build target, they are trying to customize the build_for or they just want all of the phases enabled.

That does end up being the case most of the time, but I think if they don't specify `build_for`, then we don't know their intent, and if they want to customize that they can.

But maybe that's the wrong way of looking at it. Maybe it's exactly as you say, and then the only change I would request is if someone uses `build_for()` without specifying anything (all are `undefined`) we turn them all to `enabled`. This way both not specifying it and specifying an empty one end up behaving the same.

@erikkerber @jpsim How were you using this? What was the expected outcome? I assume it was listing a target that wasn't listed in a `launch_action` or `test_action`, like grouping a top-level framework target with some unassociated app and test targets (in which case the default would end up building for all actions)?

brentleyjones: One thing I thought of: if someone starts with a `build_target` and later adds a test that depends on it; it would be surprising for the build_for to change.

So, you/I have talked myself back on that part. Let's just make an empty `build_for()` an error to resolve that semantic difference issue.

brentleyjones: @cgrindel Don't make any more changes. I'll take this PR as it stands for now. I have some ideas how to tweak this further, but I can take that on myself as I write some docs for this feature.

brentleyjones (approved):

PR #965, opened 2022-08-24
Add target environment selection to `xcodeproj.top_level_targets`

brentleyjones:
This deprecates the `device_and_simulator` rule.


brentleyjones:
Draft because I need to add more documentation (and resolve the fact that we want actually want to document the macro, not the rule, since we do a lot in it now). But still ready for general review.

cgrindel (approved):

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -698,6 +768,18 @@ targets specified in the `focused_targets` attribute.
...
+        ),
+        "ios_simulator_cpus": attr.string(),
+        "tvos_device_cpus": attr.string(
+            default = "arm64",
+        ),
+        "tvos_simulator_cpus": attr.string(),
+        "watchos_device_cpus": attr.string(
+            default = "arm64_32",
+        ),
+        "watchos_simulator_cpus": attr.string(),
```
brentleyjones, line 782: I need to add these to the macro and add documentation. I will before merging.

Issue #949, opened 2022-08-22
Feature Request: Ensure application icons are specified correctly in Xcode project


cgrindel:
The benefit is that Xcode will use the icons in its user interface (e.g., schemes).

## Tasks
- [x] #932
- [x] #966 
- [x] Fix issues when there are multiple Info.plist for a single target: #985
- [ ] Update the BwB app icon selection to use information in the `Contents.json` for the resource set.

brentleyjones:
I don't believe this should block 1.0.

PR #966, opened 2022-08-24
Ensure app icons for targets in BwB projects appear in schemes user interface

cgrindel:
Related to #949.

- Move application icon functions to `app_icons.bzl`.
- Add `app_icons.get_info()` to retrieve application icon information for a target.
- Implement rudimentary logic for finding a default application icon.

### DIFF
--- a/xcodeproj/internal/app_icons.bzl
+++ b/xcodeproj/internal/app_icons.bzl
@@ -93,5 +93,5 @@ def _create(set_name, set_path, default_icon_path):
 
 app_icons = struct(
     create = _create,
-    get_app_icon_info = _get_app_icon_info,
+    get_info = _get_app_icon_info,
 )
--- a/xcodeproj/internal/app_icons.bzl
+++ b/xcodeproj/internal/app_icons.bzl
@@ -1,0 +1,59 @@
+def _get_resource_set_name(path, suffix):
+    suffix_idx = path.find(suffix)
+    if suffix_idx == -1:
+        return None
+    prev_delimiter_idx = path[:suffix_idx].rfind("/")
+    if prev_delimiter_idx == -1:
+        name_start_idx = 0
+    else:
+        name_start_idx = prev_delimiter_idx + 1
+    return path[name_start_idx:suffix_idx]
+
+_RESOURCE_SET_SUFFIXES = [".appiconset", ".brandassets"]
+
+def _get_app_icon_info(ctx, automatic_target_info):
+    """Attempts to find the applicaiton icon name.
+
+    Args:
+        ctx: The aspect context.
+        automatic_target_info: The `XcodeProjAutomaticTargetProcessingInfo` for
+            `target`.
+
+    Returns:
+        The application icon name, if found. Otherwise, None.
+    """
+    if not automatic_target_info.app_icons:
+        return None
+
+    app_icons = getattr(ctx.rule.attr, automatic_target_info.app_icons, None)
+    if not app_icons:
+        return None
+
+    resource_files = [
+        file
+        for target in app_icons
+        for file in target.files.to_list()
+    ]
+    for file in resource_files:
+        for suffix in _RESOURCE_SET_SUFFIXES:
+            set_name = _get_resource_set_name(file.short_path, suffix)
+            if not set_name:
+                continue
+            return _create(
+                set_name = set_name,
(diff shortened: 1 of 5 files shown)


xcodeproj/internal/app_icons.bzl
```diff
@@ -0,0 +1,111 @@
...
+def _get_app_icon_info(ctx, automatic_target_info):
+    """Attempts to find the applicaiton icon name.
+
+    Args:
+        ctx: The aspect context.
+        automatic_target_info: The `XcodeProjAutomaticTargetProcessingInfo` for
+            `target`.
+
+    Returns:
+        The application icon name, if found. Otherwise, None.
```
brentleyjones, line 59:
```suggestion
        The application icon name, if found. Otherwise, `None`.
```

xcodeproj/internal/app_icons.bzl
```diff
@@ -0,0 +1,111 @@
...
+        default_icon_path = None
+
+    return _create(
+        set_name = set_name,
+        set_path = set_path,
+        default_icon_path = default_icon_path,
+    )
+
+def _create(set_name, set_path, default_icon_path):
+    """Create a `struct` representing informaiton about a target's application \
```
brentleyjones, line 90:
```suggestion
    """Creates a `struct` representing information about a target's application \
```

xcodeproj/internal/app_icons.bzl
```diff
@@ -0,0 +1,111 @@
...
+    icons.
+
+    Args:
+      set_name: The name of the resource set as a `string` value.
```
brentleyjones, line 94:
```suggestion
      set_name: The name of the resource set as a `string.
```

xcodeproj/internal/app_icons.bzl
```diff
@@ -0,0 +1,111 @@
...
+      set_path: The path of the resource set as a `string` value.
```
brentleyjones, line 95:
```suggestion
      set_path: The path of the resource set as a `string`.
```

xcodeproj/internal/app_icons.bzl
```diff
@@ -0,0 +1,111 @@
...
+      default_icon_path: If a default icon should be identified, the path to
+          the icon file will be set as a `string` value.
```
brentleyjones, line 97:
```suggestion
          the icon file will be set as a `string`.
```

brentleyjones (approved):

BalestraPatrick:
@cgrindel Are you aware of any situation where this doesn't work in BwB yet? I can't get it to work in our project. One thing I noticed is that the path to the icon is always relative to the repo root, but our project isn't generated in the root, which causes it not to find the icon. But even after hardcoding the absolute path, it doesn't seem to work. I can see it working in the `ios_app` example locally in BwB too.

brentleyjones:
The Info.plist needs to be generated, then Xcode closed and re-opened. The path being off is probably a bug though.

BalestraPatrick:
Yeah, forgot to mention I always closed and re-opened Xcode but it didn't make a difference (it did in the `ios_app` example). That's why I suspect there's another bug somewhere on top of the path being off.

BalestraPatrick:
Yes it's a reasonable image that is being picked.

cgrindel:
@BalestraPatrick I never tried specifying an absolute path. 

> our project isn't generated in the root,

Can you describe how you are generating the project? Perhaps, we can figure out how to replicate this situation with one of the example projects.

BalestraPatrick:
Basically the `xcodeproj` rule is in a subdirectory of the repo. I think we could replicate that in the `ios_app` by moving the `xcodeproj` to the `BUILD.bazel` file in `Example`.

brentleyjones:
`//examples/Multiplatform` is already in a subdirectory.

cgrindel:
> `//examples/Multiplatform` is already in a subdirectory.

Perfect. We can copy over the icons from the iOS app to the iOS app in Multiplatform.

cgrindel:
I will take a look at this later today.

cgrindel:
I was able to reproduce the issue using `//examples/multiplatform`.  Still investingating.

brentleyjones:
https://github.com/buildbuddy-io/rules_xcodeproj/pull/985 should fix it.

PR #974, opened 2022-08-25
Add more docs to `xcodeproj` macro

brentleyjones:
Also adjust the docs to use the macro again. This adds clarity around `schemes` and is needed for future changes to the macro.


doc/BUILD.bazel
```diff
@@ -4,7 +4,7 @@ load("@io_bazel_stardoc//stardoc:stardoc.bzl", "stardoc")
 
 _DOC_SRCS = [
     "experimental",
-    "xcodeproj.doc",
```
brentleyjones, line 7: I'm leaving the `.doc` logic since I'm changing `experimental` to `experimental.doc` in a followup PR.

PR #976, opened 2022-08-25
Add support for "Compile File" command

### LINKED ISSUE #931: Bug: "Compile File" commands take longer than in vanilla Xcode projects
## Describe the bug

In projects generated via `rules_xcodeproj` -- at least with `build_mode = "bazel"` -- compiling individual files with Product -> Perform Action -> Compile File (or ⌃⌘B) both compiles the file and links the library to which it belongs. In comparison, vanilla Xcode projects never link on ⌃⌘B.  The linking step might take awhile, increasing the length of the feedback loop for certain tasks compared to a vanilla project.

## Reproduction steps

Make a semantically relevant change to a file and then press ⌃⌘B. (Certain changes, such as adding text to an existing comment, seem not trigger the issue. You might have to play around with this.)

## Expected behavior

Only that file is compiled.

## Versions
- Bazel: 5.2.0
- rules_xcodeproj: fef9f9b89f1c6bd1a9ee63ff9dfdf4776bc831e6
- rules_apple: 1.0.1
- rules_swift: as installed by rules_xcodeproj

brentleyjones:
Fixes #931.

This PR changes how the `calculate_output_groups.py` script works, having it use knowledge of the destination platform to select the correct `BAZEL_TARGET_ID`s for all `configuredTargets`, and only using the scheme target ids as a backup.

It also has an unused implementation to pick the library target of a merged target. It's unused because Xcode doesn't update the `buildRequest` reliably for it, and we can get stuck building the wrong output group.

### DIFF
--- a/examples/ios_app/test/fixtures/bwb.xcodeproj/project.pbxproj
+++ b/examples/ios_app/test/fixtures/bwb.xcodeproj/project.pbxproj
@@ -1326,6 +1326,7 @@
 				ARCHS = x86_64;
 				BAZEL_PACKAGE_BIN_DIR = "bazel-out/ios-x86_64-min15.0-applebin_ios-ios_x86_64-dbg-ST-80b7926e333b/bin/TestingUtils";
 				BAZEL_TARGET_ID = "//TestingUtils:TestingUtils ios-x86_64-min15.0-applebin_ios-ios_x86_64-dbg-ST-80b7926e333b";
+				"BAZEL_TARGET_ID[sdk=iphonesimulator*]" = "$(BAZEL_TARGET_ID)";
 				DEBUG_INFORMATION_FORMAT = dwarf;
 				ENABLE_BITCODE = NO;
 				ENABLE_TESTABILITY = YES;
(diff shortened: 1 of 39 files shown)


xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
```diff
@@ -96,49 +131,95 @@ def _calculate_guid_target_ids(base_objroot):
...
+def _select_target_id(target_ids, platform):
+    key = target_ids["key"]
+
+    platforms = {platform: None}
+
+    # We need to try other similar platforms (i.e. other simulator platforms if
+    # `platform`` is for a simulator). This is to support schemes with targets
+    # of multiple platforms in them. Because `dict` is insertion ordered,
+    # `platform` will be checked first.
+    platforms.update(_similar_platforms(platform))
```
brentleyjones, line 176: This is the only thing that is questionable. I'm not sure it matches Xcode's order exactly. In practice it won't break for autogenerated schemes. For custom schemes, there is a possibility that mixed device/simulator targets are put together, but then the default handling below should work. I can't think of ways this breaks, but it's the part that would break if anything.

brentleyjones:
@BalestraPatrick @jpsim @erikkerber Can y'all test this in your projects? This is on the riskier side and I want to have good coverage.

xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
```diff
@@ -36,9 +51,32 @@ def _calculate_output_group_target_ids(
             # Parse the build-request.json file
             build_request = json.load(f)
 
-        buildable_target_ids = []
+        # Xcode gets "suck" in the `buildFiles` or `build` command for top-level
```
jpsim, line 54: Xcode does sometimes suck, but I don't think that's what you meant to say here 🙃 

```suggestion
        # Xcode gets "stuck" in the `buildFiles` or `build` command for top-level
```

xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
```diff
@@ -36,9 +51,32 @@ def _calculate_output_group_target_ids(
...
+        # targets, so we can't reliably change commands here. Leaving the code
+        # in place in case this is fixed in the future, or we want to do
```
jpsim, line 56: Could be good to file a radar on this, and include the number in the comment.

xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
```diff
@@ -96,49 +131,95 @@ def _calculate_guid_target_ids(base_objroot):
...
-
-    try:
-        build_request_file = _calculate_build_request_file(objroot)
-        guid_target_ids = _calculate_guid_target_ids(base_objroot)
-    except Exception:
-        print(
-            f"""\
+    if prefix == "i":
+        # buildRequest for Index Build includes all targets, so we have to
+        # fallback to the scheme target ids (which are actually set by the
```
jpsim, line 200: Nit used as a verb, not a noun.

```suggestion
        # fall back to the scheme target ids (which are actually set by the
```

jpsim (approved):
My local testing worked, running in an iOS simulator, on an iOS device and compiling a single file. I think it's the first time in my 11 years of using Xcode I used that command 😅

BalestraPatrick:
@brentleyjones Seems to work fine for me. Tested in two different projects containing extensions and tests and couldn't spot any regression.

xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
brentleyjones: Could be, but I don't want to draw attention to these files, as they are not public API. I believe when we implement Build with Bazel via Proxy we will get the correct command in the API call, regardless of stale cache files.

brentleyjones:
> I think it's the first time in my 11 years of using Xcode I used that command 😅

Yep, same.

erikkerber:
Late here but our project does work fine with this as well. It does seem to link like you noted in the issue.

Now I need to see if this should fit into my (or others) workflows 😂

brentleyjones:
I think using it to confirm a module you are working on compiles is a nice workflow. This way you don't have to switch your scheme away from the App scheme.

PR #980, opened 2022-08-29
Use non-internal label and name for `*_{ui,unit}_test` targets

### LINKED ISSUE #215: Bug: `*_{unit,ui}_test` targets use the name of the internal bundle
## Describe the bug

`rules_apple` test rules currently have their target names match the internal bundle target instead of the primary target. For example, they are named `ExampleTests.__internal__.__test_bundle` instead of `ExampleTests`.

brentleyjones:
Fixes #215.

This is a breaking change for focused targets and schemes, but one that is easily recovered from.


tinder-maxwellelliott (approved):
Thank you!

jpsim (approved):
Nice ergonomic improvement 👍

PR #990, opened 2022-08-30
Remove `xcodeproj.archived_bundles_allowed`

brentleyjones:
We can now default this to the correct value, and projects can change that in their `.bazelrc` or `xcodeproj.bazelrc` file.


brentleyjones:
@jpsim Tagged you since I know Envoy Mobile had to use `archived_bundles_allowed = True`, so I wanted to make sure this new process works for your use case.

jpsim (approved):

PR #992, opened 2022-08-30
Clarify target names by platform

maxwellE:
This should aid in debugging the generated xcodeproject


examples/multiplatform/xcodeproj_targets.bzl
```diff
@@ -11,19 +11,19 @@ XCODEPROJ_TARGETS = [
         label = "//examples/multiplatform/iOSApp",
         target_environments = ["device", "simulator"],
     ),
-    "//examples/multiplatform/macOSApp/Example:Example",
+    "//examples/multiplatform/macOSApp/Source:macOSApp",
```
brentleyjones, line 14: Not as part of this PR, but we should add unit and unit tests for macOS.

brentleyjones (approved):

examples/multiplatform/xcodeproj_targets.bzl
maxwellE: UI and UnitTests?

brentleyjones: Yeah. We don't have any examples of `macos_ui_test`. We will have `macos_unit_test` when `command_line` is merged, so we don't need to add that.

PR #996, opened 2022-08-30
Split output groups by function

brentleyjones:
We now have these outputs groups:

- `xc`: "Xcode Compiling Inputs", inputs needed to support the compile actions when using BwX
- `xl`: "Xcode Linking Inputs", inputs needed to support the linking and bundling actions when using BwX
- `bc`: "Bazel Compiled Outputs", compiled outputs needed for BwB mode (i.e. swiftmodules)
- `bp`: "Bazel Product Outputs", product outputs needed for BwB mode (i.e. binaries and bundles)

When indexing we only use `xc` and `bc`, otherwise we use `xc,xl` and `bc,bp`.

In the future we will probably add another set of output groups for index stores, to support index-while-building (it will be a different output group to allow us to have a single file to feed to `index-import`).


brentleyjones:
Another risky change, but local tests seem to check out.

brentleyjones:
I've been living on this change for the past 23 hours and it seems to work. Merging and will fix issues if they are reported.

PR #1001, opened 2022-08-31
Skip `apple_universal_binary` targets

brentleyjones:
We will just include the two configurations in the project instead. Later we can hopefully consolidate them into a single target.


brentleyjones:
Herp. Wrong branch.

PR #1009, opened 2022-08-31
Extract `bazel_build.sh`

brentleyjones:
Having this script be in a file makes it easier to edit and review.


xcodeproj/internal/bazel_integration_files/bazel_build.sh
```diff
@@ -0,0 +1,204 @@
...
+  output_path)
+exec_root="${output_path%/*}"
+
+# Create `bazel.lldbinit``
+
+if [[ "$ACTION" != "indexbuild" && "${ENABLE_PREVIEWS:-}" != "YES" ]]; then
+  # shellcheck disable=SC2046
+  "$BAZEL_INTEGRATION_DIR/create_lldbinit.sh" \
+    "$exec_root" \
+    $(xargs -n1 <<< "${RESOLVED_EXTERNAL_REPOSITORIES:-}") \
```
brentleyjones, line 124: Ideally I wouldn't have to invoke another process just to keep quoted substrings as strings, Anyone know any bash-only ways of doing this?

erikkerber: What is a typical value of `RESOLVED_EXTERNAL_REPOSITORIES`? Space separated paths?

erikkerber (approved):
Never had a reason to peruse the old way but this is nicer 👍🏼

PR #1027, opened 2022-09-02
Add support for custom Swift toolchains

brentleyjones:


xcodeproj/internal/bazel_integration_files/bazel_build.sh
```diff
@@ -177,6 +186,7 @@ log=$(mktemp)
   build \
   --config=$config \
   --color=yes \
+  ${toolchain:+--define=SWIFT_CUSTOM_TOOLCHAIN="$toolchain"} \
```
keith, line 189: note that this doesn't really work in a lot of ways last I tried

brentleyjones: What breaks?

brentleyjones: https://github.com/bazelbuild/rules_swift#building-with-custom-toolchains is front and center, so if it doesn't work 😬

keith: linking will probably fail because it doesn't know about the toolchain path

brentleyjones: Worked in my, very minimal, testing: https://app.buildbuddy.io/invocation/ae071594-1485-41bb-9b97-675b3d88598d

thii (approved):

xcodeproj/internal/bazel_integration_files/bazel_build.sh
keith: At the very least you need `--action_env=TOOLCHAINS=toolchain_id` to make sure that wrapped_clang invocations can use the same toolchain. In some situations you'll also need `--host_action_env` with the same

PR #1031, opened 2022-09-04
Stamp build instead of invalidating it with a cache buster

thii:


brentleyjones:
I went this route at the very start, but we can't do it because projects might already use `--embed_label` for something.

thii:
Ahh I see, hadn't thought of that (we only stamp App Store builds).

PR #1053, opened 2022-09-07
Store Index Build output base inside of normal output path

### LINKED ISSUE #1048: Bug: Lots of `Permission denied` errors when trying to delete Xcode's DerivedData
## Describe the bug
```
rm: /Users/thi/Library/Developer/Xcode/DerivedData/Projects-aupyeyirutuqslakjvpamkrvkaiv/Index/Build/Intermediates.noindex/bazel_output_base/execroot/__main__/bazel-out/ios-x86_64-min14.0-applebin_ios-ios_x86_64-dbg-ST-223d7abe647f/bin/external/Nuke/Nuke.indexstore/v5/units: Permission denied
```

## Reproduction steps
Generate a project, build, and run

```
rm -rf ~/Library/Developer/Xcode/DerivedData
```

## Expected behavior
Successfully deleted the DerivedData directory.

## Versions
- rules_xcodeproj: https://github.com/buildbuddy-io/rules_xcodeproj/commit/7ee46ec29bb4365f2c81525200a078b80bf5ca4a

brentleyjones:
Fixes #1048.

We need to have a separate output base for two reasons:

1. To prevent Index Build from blocking normal compiles, by having a separate Bazel lock
2. Because indexing is on a different cycle from building, the outputs might differ slightly (and it might have different flags)

Before this PR we stored the output base in Derived Data. This works, but since Bazel sets read-only permissions on items in that directory, it makes simple commands like `rm -rf ~/Library/Developer/Xcode/DerivedData` fail.

Now we store this new output base under the normal output path (i.e. `$BAZEL_OUT`).


BalestraPatrick (approved):

xcodeproj/internal/bazel_integration_files/bazel_build.sh
```diff
@@ -93,11 +87,21 @@ bazel_cmd=(
...
 # Determine Bazel output_path
 
+if [ "$ACTION" == "indexbuild" ]; then
+  # We use a different output base for Index Build to prevent normal builds and
+  # indexing waiting on bazel locks from the other. We nest it inside of the
+  # normal `bazel-out` directory so that it's cleaned up when running
+  # `bazel clean`.
+
+  bazel_cmd+=(
+    --output_base "$BAZEL_OUT/_rules_xcodeproj/indexbuild_output_base"
```
thii, line 101: Can I request a small change? Can we put the index build output base directly under the main output base instead of under `bazel-out`? I think it's unlikely people want to clean the index store when they run `bazel clean`. Xcode also doesn't clean up the project's `DataStore` directory when you do Cmd + Shift + K. Nesting it in the main output base ensures that a `bazel clean --expunge` from the main build will clean it up.

brentleyjones: Yeah, sounds good, I'll make that change.

PR #1062, opened 2022-09-09
Add support for per-`xcodeproj` Bazel configurations

### LINKED ISSUE #641: Feature Request: Allow per-`xcodeproj` Bazel configuration
In some cases it is useful to specify the configuration only for building from Xcode.
It would be useful to have something like this:
```
xcodeproj(
    name = "xcodeproj",
    project_name = "App",
    bazel_build_config = "rules_xcodeproj"
...
)
```
And if this argument is specified, it will be passed to all calls to bazel.
For example for the build call in `BazelDependencies`: `bazel build --config=rules_xcodeproj  ...`

Note: Now we can achieve the same behavior by setting a custom bazel_path to the wrapper script, which will add the config as a parameter. But this is not really convenient.

brentleyjones:
Fixes #641.


xcodeproj/internal/xcodeproj_macro.bzl
```diff
@@ -57,6 +58,21 @@ def xcodeproj(
             If this is set to `"bazel"`, the project will use Bazel to build
             targets, inside of Xcode. The Xcode build system still unavoidably
             orchestrates some things at a high level.
+        config: Optional. The Bazel configuration to use when generating the
```
jpsim, line 61: Is this actually optional at this level since it defaults to `rules_xcodeproj` if unset?

jpsim: ok.

brentleyjones: I'm just trying to follow convention. I might be totally wrong 🤷‍♂️.

jpsim: I have no idea what the starlark convention is. I know for Swift you'd say this "defaults to `rules_xcodeproj`" rather than say it's optional.

jpsim:
The way this works is a bit surprising to me.

For example I expected that if I have this config:

```
build:my_config --define=abcd=1
```

and I set `config = "my_config",` in my `xcodeproj` invocation that the `--define` would be picked up when building in Xcode, but in fact I have to set this in my config:

```
build:my_config_build --define=abcd=1
```

Previously if I set something on the `rules_xcodeproj` config in bazelrc, it would propagate to all rules_xcodeproj specialized configs (build, info, etc.).

Should this follow the same principle?

brentleyjones:
Just `build:my_config --define=abcd=1` should work, as `build:my_config_build --config=my_config` should be in the `xcodeproj_extra_flags.bazelrc`. I'll see why that's not the case.

examples/integration/test/fixtures/bwb.xcodeproj/rules_xcodeproj/bazel/xcodeproj_extra_flags.bazelrc
```diff
@@ -1,4 +1,14 @@
...
-build:rules_xcodeproj_build --noverbose_failures
-build:rules_xcodeproj_indexbuild --noverbose_failures
-build:rules_xcodeproj_swiftuipreviews --noverbose_failures
+build:rules_xcodeproj_integration_build --config=rules_xcodeproj_build
+build:rules_xcodeproj_integration_generator --config=rules_xcodeproj_generator
+build:rules_xcodeproj_integration_indexbuild --config=rules_xcodeproj_indexbuild
+build:rules_xcodeproj_integration_info --config=rules_xcodeproj_info
+build:rules_xcodeproj_integration_swiftuipreviews --config=rules_xcodeproj_swiftuipreviews
+build:rules_xcodeproj_integration --config=cache
+build:rules_xcodeproj_integration_build --config=rules_xcodeproj_integration
```
brentleyjones, line 7: @jpsim Can you upload your version of this file? This shows how it should work as expected.

jpsim:
```
build:my_config_build --config=rules_xcodeproj_build
build:my_config_generator --config=rules_xcodeproj_generator
build:my_config_indexbuild --config=rules_xcodeproj_indexbuild
build:my_config_info --config=rules_xcodeproj_info
build:my_config_swiftuipreviews --config=rules_xcodeproj_swiftuipreviews
```

brentleyjones: Thanks. I see the issue. I'll add a different integration test for this as well.

brentleyjones: @jpsim Can you give this another spin? I had to go another route to get the ordering and everything correct.

doc/rules-xcodeproj.md
```diff
@@ -292,6 +292,7 @@ argument will pass forward values for globally available attributes (e.g.
...
+| <a id="xcodeproj-config"></a>config |  Optional. The Bazel configuration to use when generating the project or invoking <code>bazel</code> inside of Xcode. This is the basename of multiple configurations. For example, if this is set to <code>"rules_xcodeproj"</code>, then the following configurations will be valid for you to extend in your <code>.bazelrc</code> file:<br><br>- rules_xcodeproj - rules_xcodeproj_build - rules_xcodeproj_generator - rules_xcodeproj_indexbuild - rules_xcodeproj_info - rules_xcodeproj_swiftuipreviews<br><br>See the [baseline <code>xcodeproj.bazelrc</code> file](../xcodeproj/internal/bazel_integration_files/xcodeproj.bazelrc) that is used for more information on these configs.   |  <code>"rules_xcodeproj"</code> |
```
brentleyjones, line 295: This not converting the markdown list to a list is annoying. Looking into it.

examples/integration/test/fixtures/bwb.xcodeproj/rules_xcodeproj/bazel/xcodeproj_extra_flags.bazelrc
jpsim: Works now.

doc/rules-xcodeproj.md
brentleyjones: https://github.com/bazelbuild/stardoc/issues/118

*sigh*

tools/generator/test/Fixtures.swift
```diff
@@ -8,6 +8,7 @@ enum Fixtures {
     static let project = Project(
         name: "Bazel",
         bazelWorkspaceName: "bazel_workspace",
+        bazelConfig: "rules_xcodeproj_fxitures",
```
jpsim, line 11:
```suggestion
        bazelConfig: "rules_xcodeproj_fixtures",
```

jpsim (approved):

PR #1075, opened 2022-09-12
Refine the behavior of `rules_xcodeproj{_*}` configs

brentleyjones:
Having `rules_xcodeproj_build` not be the base config can easily lead to misconfigurations and bugs. Now `rules_xcodeproj` is the effectively the config used for building (though technically it's `_rules_xcodepoj_build` to support project-level configs), and all other configs inherit from it.

Also started the "Usage Guide" with details about the configs.


jpsim (approved):
Quick review and local validation but this is a welcome change.

doc/rules-xcodeproj.md
```diff
@@ -347,7 +347,7 @@ argument will pass forward values for globally available attributes (e.g.
...
+| <a id="xcodeproj-config"></a>config |  Optional. The Bazel config to use when generating the project or invoking <code>bazel</code> inside of Xcode. This is the basename of multiple configs. For example, if this is set to <code>"projectx_xcodeproj"</code>, then the following configs will be available for you to adjust in your <code>.bazelrc</code> file: <code>projectx_xcodeproj</code>, <code>projectx_xcodeproj_generator</code>, <code>rules_xcodeproj_indexbuild</code>, and <code>rules_xcodeproj_swiftuipreviews</code>.<br><br>See the [usage guide](usage.md#bazel-configs) for more information about adjusting Bazel configs.   |  <code>"rules_xcodeproj"</code> |
```
brentleyjones, line 350: I forgot to adjust a couple of the names. Will fix.

doc/usage.md
```diff
@@ -0,0 +1,114 @@
+- [Bazel configs](#bazel-configs)
```
erikkerber, line 1: I imagine this will be expanded to include much more than configs right?

erikkerber (approved):
All looks great, thanks!

doc/usage.md
brentleyjones: Correct. This is just a section that needs to be in this guide, so I thought I would add it alongside this PR.

PR #1076, opened 2022-09-12
Re-enable `working_directory` for `launch_action`

### LINKED ISSUE #933: Feature Request: Add support for custom working directory in launch action for custom Xcode scheme
Related to #573.

It turns out that the [LaunchAction in XcodeProj](https://github.com/tuist/XcodeProj/blob/main/Sources/XcodeProj/Scheme/XCScheme%2BLaunchAction.swift) does not have a `customWorkingDirectory` attribute. (It does have `useCustomWorkingDirectory`.) This attribute will need to be added before custom working directory can be added in `rules_xcodeproj`.

maxwellE:
Resolves #933.

Now that we have updated xcodeproj to contain https://github.com/tuist/XcodeProj/pull/720 we can set the value of the custom working
directory for LaunchActions

### DIFF
--- a/xcodeproj/internal/xcode_schemes.bzl
+++ b/xcodeproj/internal/xcode_schemes.bzl
@@ -258,7 +258,7 @@ def make_xcode_schemes(bazel_labels):
             target = bazel_labels.normalize(target),
             args = args,
             env = env,
-            working_directory = working_directory
+            working_directory = working_directory,
         )
 
     return struct(
--- a/xcodeproj/internal/xcode_schemes.bzl
+++ b/xcodeproj/internal/xcode_schemes.bzl
@@ -249,7 +249,8 @@ def make_xcode_schemes(bazel_labels):
                 environment variables when the target is executed.
             working_directory: Optional. A `string` that will be set as the custom
                  working directory in the Xcode scheme's launch action. Relative
-                 paths will be relative to value of BUILT_PRODUCTS_DIRECTORY.
+                 paths will be relative to the value of `target`'s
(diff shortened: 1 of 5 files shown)


xcodeproj/internal/xcode_schemes.bzl
```diff
@@ -246,6 +247,8 @@ def make_xcode_schemes(bazel_labels):
                 the target when executed.
             env: Optional. A `dict` of `string` values that will be set as
                 environment variables when the target is executed.
+            working_directory: Optional. A `string` that will be set as the custom
```
jpsim, line 250: Is it worth documenting what the relative path behavior is here?

brentleyjones: Yes, especially since off-hand I'm not sure what it is. I believe it's relative to the executable's `BUILT_PRODUCTS_DIRECTORY`, which is usually of little use to anyone.

jpsim (approved):

xcodeproj/internal/xcode_schemes.bzl
```diff
@@ -246,6 +247,9 @@ def make_xcode_schemes(bazel_labels):
...
+                 working directory in the Xcode scheme's launch action. Relative
+                 paths will be relative to value of BUILT_PRODUCTS_DIRECTORY.
```
brentleyjones, line 252:
```suggestion
                 working directory in the Xcode scheme's launch action. Relative
                 paths will be relative to the value of `target`'s
                 `BUILT_PRODUCTS_DIR`, which is unique to it.
```

brentleyjones:
Thanks!

PR #1080, opened 2022-09-12
Add support for adding extra files to generated project

BalestraPatrick:
This PR fixes https://github.com/buildbuddy-io/rules_xcodeproj/issues/1055.


examples/integration/xcodeproj_targets.bzl
```diff
@@ -18,8 +18,9 @@ XCODEPROJ_TARGETS = [
     ),
     "//iMessageApp",
     top_level_target(
-        label = "//iOSApp",
+        label = "//iOSApp/Source:iOSApp",
```
BalestraPatrick, line 21: This was needed because `//iOSApp` is an `alias` and I'm not sure we can get the comparison between labels to work correctly if they're an `alias` (similar to how we enforce non-`alias` labels in the schemes API). Thoughts?

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -189,6 +192,11 @@ targets.
                 additional_linking_files,
             )
 
+        for file, owner_label in owned_extra_files.items():
+            if str(label) == str(owner_label):
+                for f in file.files.to_list():
+                    focused_targets_extra_files.append((None, [file_path(f)]))  # FIXME: should use label instead of None
```
BalestraPatrick, line 198: I'm not sure what the key value should be here. If I leave it `None`, things seems to work fine, but from the initial issue, it seems like it should instead be the label of the target it should be associated with? If I simply pass `label` the file doesn't get added to the project.

brentleyjones: It should be the `id` of the target. So we should create a dictionary that maps labels to every id, then here we would append once for each id in that list for the label.

xcodeproj/internal/xcodeproj_macro.bzl
```diff
@@ -237,9 +244,11 @@ in your `.bazelrc` or `xcodeproj.bazelrc` file.""")
         build_mode = build_mode,
         bazel_path = bazel_path,
         config = config,
+        unowned_extra_files = extra_files,
```
brentleyjones, line 247: Misplaced (ordering)?

examples/integration/xcodeproj_targets.bzl
brentleyjones: Correct, we can't resolve aliases in Starlark. Since we already have that restriction, this seems fine. Let's make sure we have the same `fail()` error when this is an alias.

xcodeproj/internal/xcodeproj_macro.bzl
```diff
@@ -268,11 +277,12 @@ in your `.bazelrc` or `xcodeproj.bazelrc` file.""")
 
 _VALID_TARGET_ENVIRONMENTS = sets.make(["device", "simulator"])
 
-def top_level_target(label, *, target_environments = ["simulator"]):
+def top_level_target(label, extra_files = None, *, target_environments = ["simulator"]):
```
brentleyjones, line 280:
```suggestion
def top_level_target(label, *, extra_files = None, target_environments = ["simulator"]):
```

xcodeproj/internal/xcodeproj_macro.bzl
```diff
@@ -268,11 +277,12 @@ in your `.bazelrc` or `xcodeproj.bazelrc` file.""")
...
     """Constructs a top-level target for use in `xcodeproj.top_level_targets`.
 
     Args:
         label: A `Label` or label-like string for the target.
+        extra_files: Optional. A `list` of extra files to be added to the project.
```
brentleyjones, line 285:
```suggestion
        extra_files: Optional. A `list` of extra files to be added to the project. These files won't be added to the project if this target is unfocused.
```

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -796,6 +818,13 @@ labels must match transitive dependencies of the targets specified in the
 """,
             default = [],
         ),
+        "owned_extra_files": attr.label_keyed_string_dict(
+            allow_files = True,
+            doc = """\
+An optional dictionary of files to be added to the project. The key represents
+the file and the value is the label it should be associated with.
```
brentleyjones, line 825:
```suggestion
the file and the value is the label of the target it should be associated with. These files won't be added to the project if the target is unfocused.
```

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -856,6 +885,13 @@ as two separate but similar Xcode targets.
             aspects = [xcodeproj_aspect],
             providers = [XcodeProjInfo],
         ),
+        "unowned_extra_files": attr.label_list(
```
brentleyjones, line 888: Should be below `unfocused_targets` (ordering).

xcodeproj/internal/xcodeproj_rule.bzl
BalestraPatrick: I noticed one thing when implementing this (which can be seen in the updated fixtures as well). Due to the fact that one target can map to multiple ids, this can cause some interesting things:
```
//Lib:Lib -> //Lib:Lib ios-sim_arm64-min15.0-applebin_ios-ios_sim_arm64-dbg-ST-8c5deb722afc
//Lib:Lib -> //Lib:Lib watchos-arm64-min7.0-applebin_watchos-watchos_arm64-dbg-ST-9b383575a231
//Lib:Lib -> //Lib:Lib tvos-sim_arm64-min15.0-applebin_tvos-tvos_sim_arm64-dbg-ST-88387ad25201
//Lib:Lib -> //Lib:Lib watchos-arm64-min7.0-applebin_watchos-watchos_arm64-dbg-ST-50c41d527b2
```

Due to the logic to merge targets, is this fine, or do we need to find a different way?

brentleyjones: This is expected. And it means that the file will be in the project if any of those targets end up in the file. A file is only added once, regardless of how many targets claim "ownership".

Since we currently focus or unfocus based on label, and not id, we could probably just add ownership to the first id in the list. And we could probably go one step further just make the dictionary hold a single id instead of a list of ids. (All of these optimizations with comments, in case that assumption changes in the future).

xcodeproj/internal/xcodeproj_macro.bzl
brentleyjones: This change seems to have been lost?

xcodeproj/internal/xcodeproj_rule.bzl
brentleyjones: This change seems to have been lost?

BalestraPatrick: Wrong merge conflicts, should be fixed now.

xcodeproj/internal/xcodeproj_macro.bzl
BalestraPatrick: Fixed now.

xcodeproj/internal/xcodeproj_rule.bzl
BalestraPatrick: I can add the optimizations in a follow-up.

brentleyjones:
I just realized something. We don't want the owned_extras to be tied to `top_level_targets`, because then libraries can't have extra files. We want it as a sibling argument to `extra_files` on the macro (called `associated_extra_files`? I'm open to ideas on this). The format can be a dictionary of labels to a list of files, and in the macro we can convert it the same as now, so I only think at the macro level we need a change.

BalestraPatrick:
@brentleyjones I've added your suggestion which makes sense and it seems to work locally when regenerating the project (I can see the `README` in the `Lib` target). What I don't understand is why the fixtures don't reflect this addition after running `./test/update_all_fixtures.sh`.

brentleyjones:
You need to add those properties to the fixture macro and set the values in `xcodeproj_targets.bzl`, so it's set both in the fixture and the manually generated project.

xcodeproj/internal/top_level_target.bzl
```diff
@@ -5,11 +5,12 @@ load("@bazel_skylib//lib:sets.bzl", "sets")
 
 _VALID_TARGET_ENVIRONMENTS = sets.make(["device", "simulator"])
 
-def top_level_target(label, *, target_environments = ["simulator"]):
+def top_level_target(label, *, extra_files = None, target_environments = ["simulator"]):
```
brentleyjones, line 8: We want to remove this now, instead using `associated_extra_files`.

xcodeproj/internal/xcodeproj_macro.bzl
```diff
@@ -57,6 +59,10 @@ def xcodeproj(
             `--define=apple.experimental.tree_artifact_outputs` on
             `build:rules_xcodeproj` in your `.bazelrc` or `xcodeproj.bazelrc`
             file.
+        associated_extra_files: An optional dictionary of files to be added to
+            the project. The key represents the label of the target it should
+            be associated with and the value is a list of files.
+            These files won't be added to the project if the target is unfocused.
```
brentleyjones, line 65:
```suggestion
        associated_extra_files: Optional. A `dict` of files to be added to the
            project. The key is a `string` value representing the label of the
            target the files should be associated with, and the value is a
            `list` of `File`s. These files won't be added to the project if the
            target is unfocused.
```

xcodeproj/internal/xcodeproj_macro.bzl
```diff
@@ -83,6 +89,8 @@ def xcodeproj(
 
             See the [usage guide](usage.md#bazel-configs) for more information
             on adjusting Bazel configs.
+        extra_files: Optional. A `list` of extra files to be added to the project.
+            These files won't be added to the project if this target is unfocused.
```
brentleyjones, line 93:
```suggestion
        extra_files: Optional. A `list` of extra `File`s to be added to the
            project.
```

xcodeproj/internal/xcodeproj_macro.bzl
brentleyjones: ☝️ should be moved to after `unfocused_targets`.

examples/integration/BUILD
```diff
@@ -6,6 +6,10 @@ load(":xcodeproj_targets.bzl", "CONFIG", "XCODEPROJ_TARGETS")
 
 xcodeproj(
     name = "xcodeproj",
+    associated_extra_files = {
+        "//Lib": ["//Lib:README.md"],
+        "//iOSApp/Source:iOSApp": ["//iOSApp:ownership.yaml"],
```
brentleyjones, line 11: 🔡 
```suggestion
        "//iOSApp/Source:iOSApp": ["//iOSApp:ownership.yaml"],
        "//Lib": ["//Lib:README.md"],
```

brentleyjones: (Though, if `buildifier` made it this order, Because Capital < lowercase, then :+1:.)

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -220,6 +224,22 @@ targets.
...
 
+        label_to_id[label] = xcode_target.id
+
+        invalid_extra_files_targets = sets.to_list(
+            sets.difference(sets.make(owned_extra_files.values()), targets_labels),
+        )
+        if invalid_extra_files_targets:
+            fail("""\
+Are you using an `alias`? `top_level_target` requires labels of the actual \
+targets: {}
```
brentleyjones, line 235: Comment is out of date now with what field is being used. Should be `associated_extra_files` now.

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -320,6 +342,16 @@ def _write_json_spec(
     # `extra_files`
     extra_files = inputs.extra_files.to_list()
     extra_files.append((None, [parsed_file_path(ctx.build_file_path)]))
+
+    # Add unowned extra files.
```
brentleyjones, line 346:
```suggestion
    # Add unowned extra files
```

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -320,6 +342,16 @@ def _write_json_spec(
...
+    for file in ctx.attr.unowned_extra_files:
+        for f in file.files.to_list():
+            extra_files.append((None, [file_path(f)]))
+
+    # Add processed owned extra files.
```
brentleyjones, line 351:
```suggestion
    # Add processed owned extra files
```

examples/integration/BUILD
```diff
@@ -6,6 +6,10 @@ load(":xcodeproj_targets.bzl", "CONFIG", "XCODEPROJ_TARGETS")
 
 xcodeproj(
     name = "xcodeproj",
+    associated_extra_files = {
+        "//iOSApp/Source:iOSApp": ["//iOSApp:ownership.yaml"],
+        "//Lib": ["//Lib:README.md"],
+    },
```
brentleyjones, line 12: My comment was lost, but this should be set in `//:xcodeproj_targets.bzl` and reused here and in the fixture target.

examples/integration/xcodeproj_targets.bzl
```diff
@@ -18,13 +18,13 @@ XCODEPROJ_TARGETS = [
...
         target_environments = ["device", "simulator"],
     ),
     "//macOSApp/Source:macOSApp",
     "//macOSApp/Test/UITests:macOSAppUITests",
     top_level_target(
-        label = "//tvOSApp",
+        label = "//tvOSApp/Source:tvOSApp",
```
brentleyjones, line 27: These changes aren't needed anymore, right?

examples/integration/BUILD
BalestraPatrick: `buildifier` doesn't mind this it seems, so I changed to alphabetical.

examples/integration/BUILD
```diff
@@ -2,10 +2,11 @@ load(
     "@com_github_buildbuddy_io_rules_xcodeproj//xcodeproj:xcodeproj.bzl",
     "xcodeproj",
 )
-load(":xcodeproj_targets.bzl", "CONFIG", "XCODEPROJ_TARGETS")
+load(":xcodeproj_targets.bzl", "ASSOCIATED_EXTRA_FILES", "CONFIG", "XCODEPROJ_TARGETS")
 
 xcodeproj(
     name = "xcodeproj",
+    associated_extra_files = ASSOCIATED_EXTRA_FILES,
```
brentleyjones, line 9: We also need an example of `extra_files`, right?

examples/integration/xcodeproj_targets.bzl
BalestraPatrick: Correct, reverted this.

brentleyjones (approved):

PR #1096, opened 2022-09-14
Fix BwX index importing

brentleyjones:
- Include the importer in BwX mode
- Fix object file path

This will only really matter once we support using indexing data from unfocused targets.


brentleyjones:
This actually only works for Xcode 13.3+. Before that Xcode didn't use the flag this is working around. I'll see if I can conditionalize this more.

PR #1100, opened 2022-09-14
Revert runner set `--xcode_version`

brentleyjones:
While `bazel_build.sh` should still have `--xcode_version`, to ensure that Bazel uses the version of Xcode you are building inside of, setting `--xcode_version` based on `xcode-select` can counter-act some people's use of `--xcode_config`. We leave the `--repo_env` though to at least try to fix https://github.com/bazelbuild/bazel/issues/8902.


brentleyjones:
Re-implemented here: https://github.com/buildbuddy-io/rules_xcodeproj/pull/1708

PR #1104, opened 2022-09-15
Make it an error to list non-top-level targets in `top_level_targets`

brentleyjones:
Hopefully we detect all top-level targets as such, and if not, we can shake them out pretty quickly.


xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -61,6 +61,24 @@ def _calculate_unfocused_dependencies(
         ).to_list()
     }
 
+def _process_dep(dep):
+    info = dep[XcodeProjInfo]
+
+    if not info.is_top_level_target:
+        fail("""
+'{label}' is not a top-level target, but was listed in `top_level_targets`. \
+Only list top-level targets (e.g. binaries, apps, or tests) in \
```
brentleyjones, line 70: I should probably mention that distributable frameworks are top-level.

thii (approved):

danielgronlund (approved):

thii:
I feel like this naming kind of conflicts with what Bazel calls top-level targets though. In Bazel terminologies, top-level targets are the targets that you specify in the build command, but here they mean runnable/distributable targets.

brentleyjones:
There are discussions about how library targets shouldn't be buildable on their own (https://groups.google.com/g/bazel-dev/c/QK7CI__ReDM/m/KmSnzudXAAAJ), and in there, they even refer to them as "non-top-level libraries".

I agree that Bazel conflates the two, and I'm always willing to rename things to be more clear.

erikkerber (approved):
No effect on our setup

PR #1117, opened 2022-09-17
Add support for custom pre-build scripts

thii:
Part of https://github.com/buildbuddy-io/rules_xcodeproj/issues/1111


tools/generator/src/Generator/AddBazelDependenciesTarget.swift
```diff
@@ -104,13 +106,41 @@ $(BAZEL_INTEGRATION_DIR)/calculate_output_groups.py
...
+                         mode: .script).string
+            let script = try createBuildScript(
+                in: pbxProj,
+                name: "Pre-build",
+                scriptPath: scriptPath
+            )
+            buildPhases.insert(script, at: 0)
+        }
+
+        if let postBuildScript = postBuildScript {
```
thii, line 127: Calling this post-build is not entirely correct—it's actually a post-bazel build.

examples/simple/BUILD
```diff
@@ -13,9 +13,15 @@ swift_binary(
...
 
+filegroup(
+    name = "pre_build",
+    srcs = ["pre build.sh"],
+)
+
 xcodeproj(
     name = "xcodeproj",
     build_mode = "bazel",
+    pre_build = ":pre_build",
```
BalestraPatrick, line 24: This wouldn't make it possible to have different pre-build scripts for different targets, right? For example, we have one test target where we'd like to fetch some big outputs required for the tests to run, but for other top-level targets we'd like to run SwiftLint for example. My initial idea for this implementation was to have a dictionary `target -> script` like I recently added for `extra_files`, but maybe that's not needed in this case.

tools/generator/src/Generator/AddBazelDependenciesTarget.swift
```diff
@@ -104,13 +106,41 @@ $(BAZEL_INTEGRATION_DIR)/calculate_output_groups.py
                 filePathResolver: filePathResolver
             )
 
+        var buildPhases = [
```
BalestraPatrick, line 109: Doing this here will make it that these run scripts won't apply to BwX. Is that something we want to support as well?

examples/simple/BUILD
erikkerber: While this approach is similar to what we used in our last project generator and solves for some basic things like swiftlint integration, I agree that being able to apply this at the target level is important long term.

How about:

- Making this a label_list instead of a label, for multiple pre/post build scripts
- Each item is a, let's call it `xcodeproj.buildphasescript` struct with a label (for the script) and an optional label_list that calls out the targets that script applies to (similar utility to what @BalestraPatrick mentioned)

That way the easy/default way is to apply any script to all targets, but each script can target specific targets if needed.

thii: Thanks @BalestraPatrick and @erikkerber for the comments!

The suggestions all make sense. I have a question: where (which targets) should we insert the pre-build phases? In BwB, the actual build happens in the `BazelDependencies` target, the build in each library target is just a fake build. If we insert the pre-build phase in the library target, it would only run when the actual build already finishes. But if we insert it in the `BazelDependencies` target, we'd have to always generate a `BazelDependencies` target even in BwX (AFAIK `BazelDependencies` is not generated if everything can be built with Xcode).

> we have one test target where we'd like to fetch some big outputs

erikkerber: That's a good point @thii, and in the Swiftlint specific case this could just as easily live in a build phase for `BazelDependencies` itself (and not on every target, or any combination of other targets).

To that point it's worth flushing the use cases out. I didn't realize `BazelDependencies` would be omitted if not needed, which makes sense. 

Depending on hearing more about @BalestraPatrick's use case, it could be worth always generating `BazelDependencies` for BwX while not always including the Bazel script itself if it's not needed while including any `pre_build` scripts. This would suite the "insert tools that generate warnings and errors into Xcode" use case, but would remove any real `post_build` scripts since things would happen pre-build for most of BwX.

brentleyjones: >I didn't realize BazelDependencies would be omitted if not needed, which makes sense.

But "if needed" depends on various things, and we can extend it to include pre-build scripts in BwX, without forcing it to also include a `bazel_build.sh` call.

thii:
@brentleyjones I see a lot of untracked files after running `./test/update_all_fixtures.sh`. Are these OK to commit? (same on `main`)

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        examples/cc/test/fixtures/bwb.xcodeproj/rules_xcodeproj/bazel/BUILD
        examples/cc/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/BUILD
        examples/cc/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/cc.dia
        examples/cc/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/cc.sh
        examples/cc/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/copy_outputs.sh
        examples/cc/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/ld.sh
        examples/cc/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/libtool.sh
        examples/cc/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/swiftc.py
        examples/integration/test/fixtures/bwb.xcodeproj/rules_xcodeproj/bazel/BUILD
        examples/integration/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/BUILD
        examples/integration/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/cc.dia
        examples/integration/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/cc.sh
        examples/integration/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/copy_outputs.sh
        examples/integration/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/ld.sh
        examples/integration/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/libtool.sh
        examples/integration/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/swiftc.py
        examples/simple/test/fixtures/bwb.xcodeproj/rules_xcodeproj/bazel/BUILD
        examples/simple/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/BUILD
        examples/simple/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/cc.dia
        examples/simple/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/cc.sh
        examples/simple/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/copy_outputs.sh
        examples/simple/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/ld.sh
        examples/simple/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/libtool.sh
        examples/simple/test/fixtures/bwx.xcodeproj/rules_xcodeproj/bazel/swiftc.py
        test/fixtures/generator/bwb.xcodeproj/rules_xcodeproj/bazel/BUILD
        test/fixtures/generator/bwx.xcodeproj/rules_xcodeproj/bazel/BUILD
        test/fixtures/generator/bwx.xcodeproj/rules_xcodeproj/bazel/cc.dia
        test/fixtures/generator/bwx.xcodeproj/rules_xcodeproj/bazel/cc.sh
        test/fixtures/generator/bwx.xcodeproj/rules_xcodeproj/bazel/copy_outputs.sh
        test/fixtures/generator/bwx.xcodeproj/rules_xcodeproj/bazel/ld.sh
        test/fixtures/generator/bwx.xcodeproj/rules_xcodeproj/bazel/libtool.sh
        test/fixtures/generator/bwx.xcodeproj/rules_xcodeproj/bazel/swiftc.py

nothing added to commit but untracked files present (use "git add" to track)
```

brentleyjones:
Hmm, those shouldn't be changed as part of this. I always add what that script changes though. If they look weird in the PR I'll comment on it.

brentleyjones:
@thii Can you add all of the swiftlint changes in its own PR that we can merge before this?

examples/simple/test/fixtures/bwb_spec.json
```diff
@@ -24,6 +24,7 @@
     },
     "label": "//test/fixtures:xcodeproj_bwb.generator",
     "name": "bwb",
+    "pre_build_script": null,
```
brentleyjones, line 27: Doesn't seem to be working?

examples/simple/BUILD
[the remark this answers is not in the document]
thii: Yes, either is fine, I think.

examples/simple/test/fixtures/bwb_spec.json
thii: I'm not sure why running `test/update_all_fixtures.sh` on my machine doesn't update this, but produces a lot of untracked files. I'm seeing the pre-build phase generated correctly in the project by running:

```
cd examples/simple
bazel run xcodeproj
```

thii: This file has the correct contents:

```
examples/simple/bazel-output-base/execroot/__main__/bazel-out/darwin-dbg/bin/xcodeproj.generator_spec.json
```

```
{"bazel_config":"rules_xcodeproj","bazel_workspace_name":"__main__","build_settings":{"ALWAYS_SEARCH_USER_PATHS":false,"BAZEL_PATH":"bazel","CLANG_ENABLE_OBJC_ARC":true,"CLANG_MODULES_AUTOLINK":false,"COPY_PHASE_STRIP":false,"ONLY_ACTIVE_ARCH":true,"USE_HEADERMAP":false,"VALIDATE_WORKSPACE":false},"configuration":"darwin-dbg","custom_xcode_schemes":[],"extra_files":["BUILD"],"force_bazel_dependencies":false,"index_import":{"t": "g", "_": "darwin-opt-exec-2B5CBBC6/bin/external/rules_xcodeproj_index_import/index-import"},"label":"//:xcodeproj.generator","name":"Simple","pre_build_script":"pre build.sh","replacement_labels":[],"scheme_autogeneration_mode":"auto","target_hosts":[],"target_merges":[],"targets":["//:SwiftBin darwin-dbg-ST-5d7f514c2b25",{"build_settings":{"DEBUG_INFORMATION_FORMAT":"dwarf","ENABLE_BITCODE":false,"ENABLE_TESTABILITY":true,"GCC_OPTIMIZATION_LEVEL":"0","OTHER_SWIFT_FLAGS":"-Xcc -O0 -Xcc -DDEBUG=1 -Xcc -fstack-protector -Xcc -fstack-protector-all","PRODUCT_MODULE_NAME":"SwiftBinModuleName","SWIFT_ACTIVE_COMPILATION_CONDITIONS":"DEBUG HI","SWIFT_OBJC_INTERFACE_HEADER_NAME":"","SWIFT_OPTIMIZATION_LEVEL":"-Onone","SWIFT_VERSION":"5"},"configuration":"darwin-dbg-ST-5d7f514c2b25","inputs":{"srcs":["main.swift"]},"label":"//:SwiftBin","linker_inputs":{"linkopts":["-headerpad_max_install_names","-no-canonical-prefixes","-lc++"]},"lldb_context":{"c":[{"o":["-O0","-DDEBUG=1","-fstack-protector","-fstack-protector-all"],"q":[".",{"_":"darwin-dbg-ST-5d7f514c2b25/bin","t":"g"}]}]},"name":"SwiftBin","outputs":{"p":{"_":"darwin-dbg-ST-5d7f514c2b25/bin/rules_xcodeproj/SwiftBin/SwiftBin","t":"g"}},"package_bin_dir":"bazel-out/darwin-dbg-ST-5d7f514c2b25/bin","platform":{"arch":"x86_64","minimum_os_version":"12.3","os":"macos","variant":"macosx"},"product":{"executable_name":"SwiftBin","name":"SwiftBin","path":{"_":"darwin-dbg-ST-5d7f514c2b25/bin/rules_xcodeproj/SwiftBin/SwiftBin","t":"g"},"type":"com.apple.product-type.tool"},"search_paths":{"quote_includes":[".",{"_":"darwin-dbg-ST-5d7f514c2b25/bin","t":"g"}]}}]}
```

thii: I'll take a look again tomorrow.

brentleyjones: You need to update the fixture definition. Set the value in` xcode_targets.bzl` and reuse that in both places.

tools/generator/BUILD
```diff
@@ -62,6 +62,7 @@ apple_universal_binary(
 
 xcodeproj(
     name = "xcodeproj",
+    pre_build = "//:swiftlint",
```
brentleyjones, line 65: This also needs to be set in the fixture, similar to the `simple` example.

brentleyjones (approved):

ra1028:
In native Xcode, build phases are usually set on a per-target basis, so it may be better to implement it as a parameter of `top_level_target` instead of `xcodeproj` and apply it transitively to downstream modules.
having `pre-build` in `xcodeproj` as a convenient shortcut is still reasonable, though.

thii:
@ra1028 It was discussed [here](https://github.com/buildbuddy-io/rules_xcodeproj/pull/1117#discussion_r973618592) and [here](https://github.com/buildbuddy-io/rules_xcodeproj/issues/1111#issuecomment-1250983104). This PR only addresses the use case of integrating linting tools.

ra1028:
@thii 
I missed that! Makes sense.

PR #1121, opened 2022-09-20
Set forceAbsoluteProjectPath for other -fmodule-map-file set location

### LINKED ISSUE #1120: Bug: Relative path passed to -fmodule-map-file in BwX
## Describe the bug

If using a local (not external or generated) Swift framework, BwX builds will result in several errors regarding a path not being absolute:

CleanShot 2022-09-19 at 22 23 20@2x

CleanShot 2022-09-19 at 23 07 26@2x

## Reproduction steps

1. Make a project depending on a local `apple_dynamic_(xc)framework`
2. Build with BwX

## Versions
- Bazel: 5.3.1
- rules_xcodeproj: `main` post-0.8.0
- rules_apple: 1.1.2
- rules_swift: 1.2

erikkerber:
Resolves #1120 

I'm ~78% sure this was just a typo/omission, and should match what is set in **CreateFilesAndGroups.swift**

Tested and this removes the error shown in the issue.


brentleyjones:
It wasn't an omission, it just wasn't encountered before. We only want to use that flag when needed, to keep the command-line length down. Thanks for the fix!

brentleyjones (approved):

PR #1122, opened 2022-09-20
Disable all SwiftLint rules that cause errors for now

thii:
Tested with latest SwiftLint.

### DIFF
--- a/tools/generator/src/DTO/Target+LinkerFlags.swift
+++ b/tools/generator/src/DTO/Target+LinkerFlags.swift
@@ -123,6 +123,7 @@ private func processLinkoptComponent(
             if let `extension` = filePath.path.extension,
                `extension` == "swiftmodule"
             {
+                // swiftlint:disable:next shorthand_operator
                 filePath = filePath + "\(swiftTriple).swiftmodule"
             }
         }
--- a/tools/generator/src/Generator/AddBazelDependenciesTarget.swift
+++ b/tools/generator/src/Generator/AddBazelDependenciesTarget.swift
@@ -18,6 +18,7 @@ extension Generator {
                 files.containsGeneratedFiles)
     }
 
+    // swiftlint:disable:next function_parameter_count
     static func addBazelDependenciesTarget(
         in pbxProj: PBXProj,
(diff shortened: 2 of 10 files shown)


tools/generator/src/DTO/Target+LinkerFlags.swift
```diff
@@ -123,6 +123,7 @@ private func processLinkoptComponent(
             if let `extension` = filePath.path.extension,
                `extension` == "swiftmodule"
             {
+                // swiftlint:disable:next shorthand_operator
                 filePath = filePath + "\(swiftTriple).swiftmodule"
```
brentleyjones, line 127: Can we resolve this instead?

tools/generator/src/Generator/CreateFilesAndGroups.swift
```diff
@@ -454,6 +456,7 @@ extension Generator {
             var coreDataContainer: XCVersionGroup?
             let components = fullFilePath.path.components
             for (offset, component) in components.enumerated() {
+                // swiftlint:disable:next shorthand_operator
                 filePath = filePath + component
```
brentleyjones, line 460: Can we resolve this instead?

tools/generator/src/Generator/AddBazelDependenciesTarget.swift
```diff
@@ -18,6 +18,7 @@ extension Generator {
                 files.containsGeneratedFiles)
     }
 
+    // swiftlint:disable:next function_parameter_count
```
brentleyjones, line 21: You are now globally disabling this.
```suggestion
```

brentleyjones (approved):

tools/generator/src/DTO/Target+LinkerFlags.swift
thii: It doesn't build

brentleyjones: I wonder if it needs `String()` on the rhs so it knows what the type is, or if we need to define another operator. Anyway, comment is fine for now.

PR #1124, opened 2022-09-20
Disable codesigning for bundles in BwX

### LINKED ISSUE #1123: Bug: Codesign errors building to device with BwX
## Describe the bug

Building to device with BwX, in a project that has resource bundles, fails with codesigning errors requiring these bundles to have a development team set

CleanShot 2022-09-20 at 08 46 02@2x

## Reproduction steps

Build to device with BwX with a project with resource bundles.

Xcode 14 may be a requrirement

## Expected behavior

Builds

## Versions
- Bazel: 5.3.1
- rules_xcodeproj: HEAD (0.8+)
- rules_apple: 1.1.2
- rules_swift: 1.2.0

erikkerber:
Fixes #1123.

It appears the default `CODE_SIGNING_ALLOWED` changed in Xcode 14

https://developer.apple.com/forums/thread/708659

### DIFF
--- a/tools/generator/src/Generator/SetTargetConfigurations.swift
+++ b/tools/generator/src/Generator/SetTargetConfigurations.swift
@@ -358,11 +358,8 @@ $(CONFIGURATION_BUILD_DIR)
             )
         }
         
-        if buildMode == .xcode && target.platform.os == .iOS {
-            // Xcode 14 defaults this value to YES, and iOS bundles do not need to be signed
-            if target.product.type == .bundle {
-                buildSettings["CODE_SIGNING_ALLOWED"] = false
-            }
+        if buildMode == .xcode && target.shouldDisableCodeSigning {
+           buildSettings["CODE_SIGNING_ALLOWED"] = false
         }
         
         // Set VFS overlays
@@ -631,6 +628,11 @@ extension Target {
 
     func linkParamsFilePath() throws -> FilePath {
         return try .internal(internalTargetFilesPath() + "\(name).link.params")
+    }
+    
+    // Used to work around CODE_SIGNING_ENABLED = YES in Xcode 14
+    var shouldDisableCodeSigning: Bool {
+        return platform.os == .iOS && product.type == .bundle
     }
 }
 
--- a/tools/generator/src/Generator/SetTargetConfigurations.swift
+++ b/tools/generator/src/Generator/SetTargetConfigurations.swift
@@ -357,6 +357,11 @@ $(CONFIGURATION_BUILD_DIR)
                 to: ldRunpathSearchPaths
             )
         }
+        
+        // Xcode 14 defaults this value to YES, and iOS bundles do not need to be signed
+        if target.product.type == .bundle {
+            buildSettings["CODE_SIGNING_ALLOWED"] = false
+        }
 
         // Set VFS overlays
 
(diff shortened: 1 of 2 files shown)


tools/generator/src/Generator/SetTargetConfigurations.swift
```diff
@@ -357,6 +357,13 @@ $(CONFIGURATION_BUILD_DIR)
                 to: ldRunpathSearchPaths
             )
         }
+        
+        if target.product.type == .bundle {
+            buildSettings.set(
+                "CODE_SIGNING_ALLOWED",
+                to: false
+            )
```
brentleyjones, line 365: You only need to use `.set` if the type can't be inferred. This will work:
```suggestion
            buildSettings["CODE_SIGNING_ALLOWED"] = false
```

tools/generator/src/Generator/SetTargetConfigurations.swift
```diff
@@ -357,6 +357,11 @@ $(CONFIGURATION_BUILD_DIR)
                 to: ldRunpathSearchPaths
             )
         }
+        
+        // Xcode 14 defaults this value to YES, and iOS bundles do not need to be signed
```
brentleyjones, line 361: Please limit this to BwX only.

tools/generator/src/Generator/SetTargetConfigurations.swift
```diff
@@ -357,7 +357,14 @@ $(CONFIGURATION_BUILD_DIR)
                 to: ldRunpathSearchPaths
             )
         }
-
+        
+        if buildMode == .xcode && target.platform.os == .iOS {
```
erikkerber, line 361: @brentleyjones This feels a little gross, but assume limiting to iOS is the right thing here?

Am I also reading the fixtures right in that there isn't a current iOS resource fixture?

brentleyjones: There is a generated project that has it, which is enough coverage in my mind.

And yes, we should only set this for `buildMode == .xcode && target.product.type == .bundle && target.platform.os != .macOS`. You can extract that into an extension on `target`: `var shouldCodeSign: Bool` and then the check is `buildMode == .xcode && !target.shouldCodeSign`.

tools/generator/src/Generator/SetTargetConfigurations.swift
```diff
@@ -625,6 +629,11 @@ extension Target {
     func linkParamsFilePath() throws -> FilePath {
         return try .internal(internalTargetFilesPath() + "\(name).link.params")
     }
+    
+    // Used to work around CODE_SIGNING_ENABLED = YES in Xcode 14
+    var shouldDisableCodeSigning: Bool {
+        return platform.os == .iOS && product.type == .bundle
+    }
```
erikkerber, line 636: I used the inverse of your suggestion, since I don't think we can decide in a simple context if it should be _allowed_, but we know exactly when the default should be reversed.

tools/generator/src/Generator/SetTargetConfigurations.swift
erikkerber: https://github.com/buildbuddy-io/rules_xcodeproj/pull/1124#discussion_r975491405

PR #1127, opened 2022-09-20
Add Thread Sanitizer support

chiragramani:
This PR adds Thread Sanitizer support for BwB mode.

**Repro steps:**

1. Check out the integration example directory and change the build mode from “xcode” to “bazel”.
2. ```bazel run :xcodeproj && xed .```
3. Build the AppClip target with Thread sanitizer enabled.
4. The build fails with the error message: 
```Could not determine version of clang to find its Thread Sanitizer library: /Users/chirag.ramani/Uber/rules/rules_xcodeproj/examples/integration/Integration.xcodeproj/rules_xcodeproj/bazel/cc.sh```.

**Investigation:**
- The initial idea here was to modify the existing ```cc.sh``` to also add support for the version(as that's what the above error talks about). But just making that change doesn't make it work.
- Renaming the file from ```cc.sh``` to ```clang.sh``` fixes the above issue. But then, it gives a different error:

      ```Unable to copy Thread Sanitizer library: Could not determine where it lives.```

- Looks like, Xcode needs the same relative structure as present in the default clang to copy the required tsan dylibs. 

```DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/14.0.0/lib/darwin/libclang_rt.tsan_iossim_dynamic.dylib```.

- In order to provide this relative structure, the PR symlinks the ```DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/lib/``` directory to ```$(BAZEL_INTEGRATION_DIR)/../lib```. This way, Xcode can use the custom clang as well as copy the required dylibs. Hence, updating the .gitignore.

- Now, though the above dylib was correctly referenced by Xcode but it was being removed in the “Copying Bazel outputs phase” resulting tsan not really applying. Hence, the exclude list was updated to add tsan.dylib as an exception. It's a wildcard because clang has different dylib per target os and target architecture. This also requires passing `--features=tsan` to make it fully work and hence the bazelrc was modified and a new config `build:rules_xcodeproj_tsan` was added to address this.

- After applying the above changes, Thread sanitizer works fine and shows the same LLDB and IDE UI output that pure Xcode builds show.

Screen Shot 2022-09-19 at 3 38 02 PM

**TODOs:** 

1. Planning to add a new Sanitizer app(inside example/integration) where tsan, asan examples are there - so that people who are exploring can run this and see how it works with BwB
2. Though other sanitizers will share the same code, but it needs a bit more work(```ubsan``` and ```asan```) and their support will be addressed in a different PR.

### CODE REVIEW

File: xcodeproj/internal/bazel_integration_files/clang.sh [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+    ;;
+  --serialize-diagnostics)
+    shift
+    cp "${BASH_SOURCE%/*}/cc.dia" "$1"
+    ;;
+  *.o)
+    break
+    ;;
+  -v)
+    clang=$(xcrun --find clang)
```
chiragramani, line 20: TODO: Check if the developer directory is in the environment so that we can avoid this call.

brentleyjones:
Great work @chiragramani!

Can we do most of those various changes as their own smaller PRs? For example, renaming cc.sh to clang.sh can be one PR. I would like to review each step that way.

.gitignore
```diff
@@ -3,5 +3,5 @@
 **/*.xcodeproj/xcuserdata/*
 **/*.xcodeproj/project.xcworkspace/xcuserdata/*
 **/*.xcodeproj/*
-
+**/rules_xcodeproj/lib
```
brentleyjones, line 6:
```suggestion

# Symlink to `$DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/lib`
**/*.xcodeproj/rules_xcodeproj/lib

```

examples/integration/BUILD
```diff
@@ -15,7 +15,7 @@ exports_files(["README.md"])
 xcodeproj(
     name = "xcodeproj",
     associated_extra_files = ASSOCIATED_EXTRA_FILES,
-    build_mode = "xcode",
+    build_mode = "bazel",
```
brentleyjones, line 18:
```suggestion
    build_mode = "xcode",
```

.gitignore
```diff
@@ -4,4 +4,7 @@
 **/*.xcodeproj/project.xcworkspace/xcuserdata/*
 **/*.xcodeproj/*
 
-user.bazelrc
+# Symlink to `$DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/lib`
+**/*.xcodeproj/rules_xcodeproj/lib
+
+user.bazelrc
```
brentleyjones, line 10: Nit: newline at end of file please.

tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -143,6 +155,23 @@ echo "$BAZEL_LABEL,$BAZEL_TARGET_ID" >> "$SCHEME_TARGET_IDS_FILE"
...
+        buildableReference: XCScheme.BuildableReference
+    ) -> XCScheme.ExecutionAction {
+        return .init(
+            scriptText: #"""
+if [ "${ENABLE_THREAD_SANITIZER:-}" == "YES" ]; then
+    dest="$BAZEL_INTEGRATION_DIR/lib"
+    DEV_LIB_DIR="$SYSTEM_DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/lib"
+    ln -sF "$DEV_LIB_DIR" "$dest"
+fi
+"""#,
```
brentleyjones, line 170: My other comment seems to have gone away: but I think this should be in `bazel-build.sh` instead of a scheme script. Also, we can check `BAZEL_BUILD_MODE` and only do this when `!= "xcode"`. Also also, since we have the info available we can support the custom toolchain correctly here (but shouldn't since it wouldn't match the stub currently, so add a comment to that effect).

tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -143,6 +155,23 @@ echo "$BAZEL_LABEL,$BAZEL_TARGET_ID" >> "$SCHEME_TARGET_IDS_FILE"
...
+    
+    /// Symlinks `$DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/lib` to `$(BAZEL_INTEGRATION_DIR)/lib` so that Xcode can copy sanitizers' dylibs.
+    static func symlinkDefaultToolchainUsrLibDirectory(
+        buildableReference: XCScheme.BuildableReference
+    ) -> XCScheme.ExecutionAction {
+        return .init(
+            scriptText: #"""
+if [ "${ENABLE_THREAD_SANITIZER:-}" == "YES" ]; then
+    dest="$BAZEL_INTEGRATION_DIR/lib"
+    DEV_LIB_DIR="$SYSTEM_DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/lib"
```
brentleyjones, line 167: Why the use of `SYSTEM_DEVELOPER_DIR` versus `DEVELOPER_DIR`?

xcodeproj/internal/installer.template.sh
```diff
@@ -167,4 +167,9 @@ if [[ -f "$dest/rules_xcodeproj/generated.xcfilelist" ]]; then
   done
 fi
 
+# TODO: Get this approach validated and find the right location to add this, if this approach is approved.
+touch "$dest/rules_xcodeproj/lib"
+DEV_LIB_DIR="$(xcode-select -p)/Toolchains/XcodeDefault.xctoolchain/usr/lib"
+ln -sF "$DEV_LIB_DIR" "$dest/rules_xcodeproj/lib"
```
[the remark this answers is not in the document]
chiragramani: Hey! Yes, after your suggestion of having it in `bazel_build.sh`, I discovered that Xcode expects the symlink to be available before all the build phases are evaluated. So, the options were either to have it in the installer template or in the Scheme's pre-actions.

Exploring the scheme's pre-actions because:

1. We get a lot more control here. We can only apply this only for launchable targets instead of all and also listen to the thread sanitizer env variable.

Please let me know your thoughts on this and we can discuss about how to address this more elegantly.

tools/generator/src/Extensions/XCScheme+Extensions.swift
chiragramani: I found that the Scheme script env had `SYSTEM_DEVELOPER_DIR` that was pointing to $(xcode-select -p) and looks like `DEVELOPER_DIR ` was not there. I will double-check on this and will let you know.

chiragramani: `DEVELOPER_DIR` is present - my bad. I will update the PR. Thanks!

xcodeproj/internal/installer.template.sh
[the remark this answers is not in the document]
brentleyjones: Ahh, I didn't know Xcode checked the symlink before `Bazel Build` ran. Okay, a scheme pre-action works!

tools/generator/src/Extensions/XCScheme+Extensions.swift
chiragramani: Updated!

tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -143,6 +155,23 @@ echo "$BAZEL_LABEL,$BAZEL_TARGET_ID" >> "$SCHEME_TARGET_IDS_FILE"
...
+    /// Symlinks `$DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/lib` to `$(BAZEL_INTEGRATION_DIR)/lib` so that Xcode can copy sanitizers' dylibs.
+    static func symlinkDefaultToolchainUsrLibDirectory(
+        buildableReference: XCScheme.BuildableReference
+    ) -> XCScheme.ExecutionAction {
+        return .init(
+            scriptText: #"""
+if [ "${ENABLE_THREAD_SANITIZER:-}" == "YES" ]; then
+    dest="$BAZEL_INTEGRATION_DIR/../lib"
+    DEV_LIB_DIR="$DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/lib"
+    ln -sF "$DEV_LIB_DIR" "$dest"
```
brentleyjones, line 168:
```suggestion
    # TODO: Support custom toolchains once clang.sh supports them 
    src="$DEVELOPER_DIR/Toolchains/XcodeDefault.xctoolchain/usr/lib"
    dest="$BAZEL_INTEGRATION_DIR/../lib"
    ln -sF "$src" "$dest"
```

tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -143,6 +155,23 @@ echo "$BAZEL_LABEL,$BAZEL_TARGET_ID" >> "$SCHEME_TARGET_IDS_FILE"
...
+fi
+"""#,
+            title: "Symlink Default Toolchain /usr/lib directory",
```
brentleyjones, line 171:
```suggestion
            title: "Symlink Toolchain /usr/lib directory",
```

tools/generator/test/CreateAutogeneratedXCSchemesTests.swift
```diff
@@ -172,6 +171,13 @@ class CreateAutogeneratedXCSchemesTests: XCTestCase {
                 hostIndex: hostIndex
             ),
         ]
+        if expectedTargetInfo.productType.isLaunchable {
```
brentleyjones, line 174: We should only do this for `buildMode != .xcode`, right? There shouldn't be a need for Xcode built apps.

tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -83,11 +93,13 @@ extension XCScheme {
...
 extension XCScheme.BuildAction {
     convenience init(
-        buildActionInfo: XCSchemeInfo.BuildActionInfo
+        buildActionInfo: XCSchemeInfo.BuildActionInfo,
+        otherPreActions: [XCScheme.ExecutionAction] = []
     ) throws {
         self.init(
             buildActionEntries: try buildActionInfo.targets.buildActionEntries,
-            preActions: try buildActionInfo.preActions.map(\.executionAction) + buildActionInfo.targets.map(\.targetInfo).buildPreActions(),
+            preActions:  try buildActionInfo.preActions.map(\.executionAction) + buildActionInfo.targets.map(\.targetInfo).buildPreActions() +
```
brentleyjones, line 101: nit: 80 chars

tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -26,7 +26,17 @@ extension XCScheme {
...
+            var otherPreActions = [XCScheme.ExecutionAction]()
+            if let launchableTarget = buildActionInfo.launchableTargets.first {
+                otherPreActions.append(
+                    .symlinkDefaultToolchainUsrLibDirectory(
+                        buildableReference: launchableTarget.targetInfo.buildableReference
+                    )
+                )
+            }
+            buildAction = try .init(buildActionInfo: buildActionInfo,
+                                    otherPreActions: otherPreActions)
```
brentleyjones, line 38: nit: wrapping
```suggestion
            buildAction = try .init(
                buildActionInfo: buildActionInfo,
                otherPreActions: otherPreActions
            )
```

tools/generator/test/CreateAutogeneratedXCSchemesTests.swift
chiragramani: Good point! Yeah, since we are not using a custom clang for BwX mode, we don't need it. But if there is a CC override in BwX mode, we will need a similar setup there as well. 

I updated the PR to only add this support for BwB mode since all this setup needs to have the same relative structure that Xcode expects.

tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -26,7 +26,20 @@ extension XCScheme {
     ) throws {
         let buildAction: XCScheme.BuildAction?
         if let buildActionInfo = schemeInfo.buildActionInfo {
-            buildAction = try .init(buildActionInfo: buildActionInfo)
+            var otherPreActions = [XCScheme.ExecutionAction]()
```
brentleyjones, line 29: nit:
```suggestion
            var otherPreActions: [XCScheme.ExecutionAction] = []
```

brentleyjones (approved):

brentleyjones:
This still needs a test added, right?

chiragramani:
> This still needs a test added, right?
> 
> > Once this approach is approved, I plan to move the code from AppClip to a new sanitizer test suite.

Yes, I am taking it in the next PR. 👍🏼 Planning to add a new Sanitizer app(inside example/integration) where tsan, asan examples are there - so that people who are exploring can run this and see how it works with BwB. Thoughts?

brentleyjones:
That works.

PR #1160, opened 2022-09-26
Use `additionalPaths` to better handle target merging

brentleyjones:
Now that we have the `additionalPaths` technology, we don't have to change the `packageBinDir` of a top-level target. Instead we add the paths of the merging target to the merged target's `additionalPaths`, which cause them to be found with `xcodeGeneratedFiles`.

To ensure that `.swiftmodule` lookup continues to work, we adjust the assignment to `xcodeGeneratedFiles` to use the merged target's `packageBinDir`. Also, if it's a framework target, we adjust the path to be inside of the `Modules` directory of the produced framework.


brentleyjones:
I think this causes SwiftUI Previews and Indexing to have issues because the implicit `-I` that Xcode adds will now find empty swiftmodules. I'm trying to see if there is a way around that, or if part of this has to be reverted.

brentleyjones:
I found `ENABLE_DEFAULT_SEARCH_PATHS`. I'll make a PR to use that before updating this PR.

PR #1161, opened 2022-09-26
Stop Xcode from setting default search paths

brentleyjones:
This properly fixes SwiftUI Previews and Indexing jobs, removing the need for part of our `bazel-out-overlay.yaml` VFS overlay.


tools/generator/test/Fixtures.swift
```diff
@@ -2144,19 +2164,24 @@ $(INTERNAL_DIR)/targets/a1b2c/A 2/A.link.params
...
+                "OTHER_SWIFT_FLAGS": #"""
+-vfsoverlay $(OBJROOT)/bazel-out-overlay.yaml
+"""#,
+                "PREVIEWS_SWIFT_INCLUDE_PATHS__": """
+$(PREVIEWS_SWIFT_INCLUDE_PATHS__NO)
+""",
+                "PREVIEWS_SWIFT_INCLUDE_PATHS__NO": "$(BUILD_DIR)/bazel-out/x",
+                "PREVIEWS_SWIFT_INCLUDE_PATHS__YES": """
+$(BUILD_DIR)/bazel-out/x $(PREVIEWS_SWIFT_INCLUDE_PATHS__NO)
+""",
```
brentleyjones, line 2177: This is setting the same path twice. I should make sure the path isn't already in the base ones before doing this.

PR #1172, opened 2022-09-27
Revert "Ignore `-enable-bare-slash-regex` (#1073)"

brentleyjones:
This reverts commit 05a797078fe81af491496dcc4b7f305fc2d4b9d5. This change was causing issues with indexing. I don't have time to determine why, and having this appear twice in the build command is NBD.


brentleyjones:
I might re-revert this, it's probably something else.

Issue #96, opened 2022-03-08
Support `{ios,macos,tvos,watchos}_application`


brentleyjones:
- [x] `ios_application`
- [x] `macos_application`
- [x] `tvos_application`
- [x] `watchos_application`

For each:
- Check if it's already supported, and if not, make required changes
- Add an example
- Add a fixture test

For `watchos_application`:
- [x] Also support "WatchApp for iOS App" with "Embed Watch Content" `PBXCopyFilesBuildPhase `: #718 

Non-minimal fields that need support:
- [x] Fully support `app_icons`: #730
- [x] Support `alternate_icons`: #1125
- [x] Support `codesignopts`: #593
- [x] Support `entitlements`: #367
- [x] Support `executable_name`: #721
- [x] Support `frameworks`: #1163
- [x] Support `infoplists`: #225
- [x] Support `launch_images`: #713
- [x] Support `launch_storyboard`: #712
- [x] Support `provisioning_profile`
- [ ] https://github.com/buildbuddy-io/rules_xcodeproj/issues/1653
- [x] Support `strings`: #1176

maxwellE:
@brentleyjones For the "Non-minimal fields that need support", are we saying that those are not supported at all right now?

brentleyjones:
They might be supported, but we don't have examples of them if they aren't, so we can't check them off yet.

maxwellE:
Ill start on `alternate_icons` today

brentleyjones:
`settings_bundle` needs some special `AppleResourceInfo` handling. It should work for BwB though as is.

brentleyjones:
Closed as complete because the only remaining thing has it's own issue, and it's BwX only.

Issue #78, opened 2022-03-08
Support `{ios,tvos,watchos}_framework`


brentleyjones:
- [x] Support `codesignopts`: #593 
- [X] Support `executable_name`: #721
- [x] Support `exported_symbols_lists`: #649 
- [x] Support `extension_safe`: #1165
- [ ] Support `hdrs`: #1178
- [x] Support `infoplists`: #225
- [x] Support `provisioning_profile`
- [x] Support `resources`: #208
- [x] Support `strings`: #1176
- [x] Support `version`: #74

Issue #72, opened 2022-03-08
Support `{ios_*,macos,tvos,watchos}_extension`


brentleyjones:
- [x] Bundled correctly: #689 
- [x] Correct schemes: #723
- [ ] Support `app_icons`: #730
- [x] Support `codesignopts`: #593 
- [x] Support `entitlements`: #367
- [X] Support `executable_name`: #721
- [x] Support `exported_symbols_lists`: #649 
- [ ] Support `frameworks`: #78 
- [x] Support `infoplists`: #225
- [x] Support `provisioning_profile`
- [x] Support `resources`: #208 
- [x] Support `sdk_frameworks`
- [ ] Support `strings`
- [x] Support `version`: #74

PR #1176, opened 2022-09-27
Add `strings` example

maxwellE:
Part of #72, #78, and #96. 

Adds a usage of `strings` with an example

### DIFF
--- a/xcodeproj/internal/top_level_targets.bzl
+++ b/xcodeproj/internal/top_level_targets.bzl
@@ -235,6 +235,15 @@ def process_top_level_target(
             ),
         )
 
+    if automatic_target_info.strings:
+        additional_files.extend(
+            getattr(
+                ctx.rule.files,
+                automatic_target_info.strings,
+                [],
+            ),
+        )
+
     infoplist = info_plists.adjust_for_xcode(
         info_plists.get_file(target),
         app_icon_info.default_icon_path if app_icon_info else None,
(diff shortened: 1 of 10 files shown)


xcodeproj/internal/top_level_targets.bzl
```diff
@@ -235,6 +235,15 @@ def process_top_level_target(
...
 
+    if automatic_target_info.strings:
+        additional_files.extend(
+            getattr(
+                ctx.rule.files,
+                automatic_target_info.strings,
+                [],
+            ),
+        )
+
```
brentleyjones, line 246: I believe strings are handled by the resource management. I think all of the handling can be removed, and we only need the example. You can confirm in the BwX example at compile and runtime?

maxwellE: Screen Shot 2022-09-27 at 4 59 19 PM

Seeing this after I remove all the handling logic and just leave in the edits to the example

brentleyjones: Yep, that is just files in the `extra_files`, you should be able to see them still in the `resources` sections.

maxwellE: Screen Shot 2022-09-27 at 5 18 05 PM

You are right! Missed those at first glance

brentleyjones (approved):

Issue #402, opened 2022-05-04
Replay Swift serialized diagnostics


brentleyjones:
Once rules_swift has the ability to generate serialized diagnostics, when building with Bazel, we should:

- Supply the needed `feature` in the `_xcodeproj_transition` (if it uses a feature)
- Add the `dia` and parseable output files to a targets output group in `output_files.bzl` for both build and index
- Copy over these files as we do other outputs, to the location expected by Xcode
- `cat` the parsable output in `stubs/swiftc.py`

brentleyjones:
rules_swift will need an updated version of this: https://github.com/bazelbuild/rules_swift/pull/304

Issue #400, opened 2022-05-04
Replay clang serialized diagnostics


brentleyjones:
Once https://github.com/bazelbuild/bazel/issues/15191 is available, when building with Bazel, we should:

- Supply the needed `feature` in the `_xcodeproj_transition`
- Add the `dia` files to a targets output group in `output_files.bzl` for both build and index
- Copy over these files as we do other outputs, to the location expected by Xcode
- Adjust `stubs/cc.sh` to only copy over `stubs/cc.dia` if a `.dia` file doesn't exist (previous Bazel versions or the flag isn't set)

PR #1177, opened 2022-09-28
Run Bazel build and stub builds in parallel for more performant builds

thii:
Instead of waiting for Bazel build to finish, we bring Bazel build to the background, and only check back for its status before copying outputs in the build of the requesting target.

### DIFF
--- a/tools/generator/src/Generator/AddBazelDependenciesTarget.swift
+++ b/tools/generator/src/Generator/AddBazelDependenciesTarget.swift
@@ -90,6 +90,20 @@ $(BAZEL_INTEGRATION_DIR)/calculate_output_groups.py
         )
         pbxProj.add(object: configurationList)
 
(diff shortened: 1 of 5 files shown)


thii:
There is a blocker with this approach: unable to disable the `ProcessInfoPlistFile` action, so targets with an Info.plist will fail.

jpsim:
Will this cause the bazel build logs to no longer be available in Xcode though?

brentleyjones:
Yeah, that's an issue as well. I think with Xcode 14 I'll be able to have the copy outputs scripts run in parallel to this job, without losing the output. When addressing #400 and #402, the goal is that only bazel-specific output is output here, and outputs for various compiles and linking are shown in the targets themselves.

thii:
The bazel build logs are printed out at the end, so there will be a delay with warnings and errors.

PR #937, opened 2022-08-18
Remove single merge restriction on target merging

brentleyjones:
This can result in multiple targets building the same code in BwX mode, but if that isn't desired a minimal target can be inserted to mimic the previous behavior.

This change makes working with SwiftUI Previews nicer, as you can create a demo application for each `swift_library`.

PR #1184, opened 2022-09-28
Revert "Revert "Remove single merge restriction on target merging (#937)" (#1017)"

brentleyjones:
This reverts commit 19c7708bc0e4db49a432cf2f92022261cb5478a6.

With 7215544cc88e9dc5c471470c3472937f01e47090, we no longer have duplicate output errors in Xcode.


brentleyjones:
My recent framework support for finding swiftmodules makes this non-deterministic. So looking into that.

brentleyjones:
Because of frameworks, I don't think this is possible currently.

PR #1188, opened 2022-09-30
Improve error handling of XCBuildData caches

brentleyjones:


xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
```diff
@@ -109,19 +127,27 @@ def _calculate_label_and_target_ids(
...
-        wait_counter += 1
-        if wait_counter > time_to_wait:
-            break
+    # The PIF cache will only be created before the `SetSessionUserInfo`
+    # command, which normally happens when a project is opened. If Derived Data
+    # is cleared  while the project is open
+    if not (os.path.exists(project_cache) and os.path.exists(target_cache)):
+        print(
+            f"""\
+error: PIFCache ({pif_cache}) doesn't exist. If you manually cleared Derived \
```
thii, line 139: I got this error now. If it's not critical, can we make this a warning instead?

brentleyjones: It is critical, as you will build all the targets in the scheme, which could make the build take a lot longer (as it could build a bunch of tests). This is a quick fail, and an easy fix.

PR #1193, opened 2022-10-02
Add Undefined Behavior Sanitizer Support

chiragramani:
This PR adds support for the undefined behaviour sanitizer.

The following code when executed in BwB mode with ubsan enabled, should error out.

```
    int x = INT_MAX;
    x += 1; // Integer overflow here
    printf("x = %i\n", x);
```
Screenshot 2022-10-01 at 7 21 11 PM

- The example application will be added in the following PR.

### DIFF
--- a/tools/generator/src/Extensions/XCScheme+Extensions.swift
+++ b/tools/generator/src/Extensions/XCScheme+Extensions.swift
@@ -171,7 +171,8 @@ echo "$BAZEL_LABEL,$BAZEL_TARGET_ID" >> "$SCHEME_TARGET_IDS_FILE"
     ) -> XCScheme.ExecutionAction {
         return .init(
             scriptText: #"""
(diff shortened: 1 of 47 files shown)


tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -171,7 +171,8 @@ echo "$BAZEL_LABEL,$BAZEL_TARGET_ID" >> "$SCHEME_TARGET_IDS_FILE"
     ) -> XCScheme.ExecutionAction {
         return .init(
             scriptText: #"""
-if [ "${ENABLE_ADDRESS_SANITIZER:-}" == "YES" ] || [ "${ENABLE_THREAD_SANITIZER:-}" == "YES" ]; then
+
```
brentleyjones, line 174:
```suggestion
```

brentleyjones (approved):

PR #1196, opened 2022-10-02
Add Scheme diagnostics

chiragramani:
This PR adds support for custom schemes specifying the diagnostics that the scheme should enable/disable by default.

This was a todo in the previous PRs. 

Currently, I find that the generator is using the custom scheme and the same can be modified to test this PR:
```
  diagnostics = xcode_schemes.diagnostics(
                    sanitizers = xcode_schemes.sanitizers(
                        thread = True,
                        undefined_behavior = True,
                    ),
                ),
```

**TODO:**
1. Once this is landed, there would be a different PR to add custom scheme support for `examples/sanitizers`.


tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -240,6 +240,9 @@ extension XCScheme.LaunchAction {
             askForAppToLaunch: launchActionInfo.askForAppToLaunch ? true : nil,
             customWorkingDirectory: launchActionInfo.workingDirectory,
             useCustomWorkingDirectory: launchActionInfo.workingDirectory != nil,
+            enableAddressSanitizer: launchActionInfo.diagnostics?.enableAddressSanitizer ?? false,
```
brentleyjones, line 243: Even though this API seems to indicate you can have multiple enabled at once, it's really a radio button, right? If so, I think it should be more like `sanitizer = "address"` in the BUILD API and translate that (optional) enum into bools here?

chiragramani: It's a checkbox but some of the functionality behaves like a radio button. (For ex - asan cannot be enabled with tsan). Eventually the diagnostics will be extended by other options such as: "Malloc Scribble", "Zombie Objects" etc, so a couple of them can be enabled at a time. 

I am thinking about this point - `sanitizer = "address" in the BUILD API `...

brentleyjones: So usan can be enabled with asan? If so, this makes sense, and we just need to error when they can't be enabled together.

I still like a `sanitizers` argument, with simple `address = True, thread = False` arguments under it. That way it's grouped like the UI in Xcode.

chiragramani: Yes, asan/tsan can be enabled with ubsan.

>  I still like a sanitizers argument, with simple address = True, thread = False arguments under it. That way it's grouped like the UI in Xcode.

Yes, I agree, that would be a better representation. I will update the PR, thanks.

tools/generator/test/XCScheme+ExtensionsTests.swift
```diff
@@ -494,3 +494,58 @@ class XCSchemeExtensionsTests: XCTestCase {
         topLevelTargetInfos: []
     ).orThrow()
 }
+
```
brentleyjones, line 497:
```suggestion
```

xcodeproj/internal/xcode_schemes.bzl
```diff
@@ -276,11 +276,49 @@ def make_xcode_schemes(bazel_labels):
...
+            sanitizers = sanitizers,
+        )
+
     def _launch_action(
             target,
             args = None,
             env = None,
-            working_directory = None):
+            working_directory = None,
+            diagnostics = None):
```
brentleyjones, line 321: Can you put this above `env`?

xcodeproj/internal/xcode_schemes.bzl
```diff
@@ -293,6 +331,7 @@ def make_xcode_schemes(bazel_labels):
                 custom working directory in the Xcode scheme's launch action.
                 Relative paths will be relative to the value of `target`'s
                 `BUILT_PRODUCTS_DIR`, which is unique to it.
+            diagnostics: Optional. A `struct` representing Xcode scheme's diagnostics.
```
brentleyjones, line 334: Can you put this above `env`?

xcodeproj/internal/xcode_schemes.bzl
```diff
@@ -355,6 +395,8 @@ def make_xcode_schemes(bazel_labels):
         build_for_values = xcode_schemes_internal.build_for_values,
         launch_action = _launch_action,
         test_action = _test_action,
+        sanitizers = _sanitizers,
+        diagnostics = _diagnostics,
```
brentleyjones, line 399: Can you put this above `sanitizers`?

xcodeproj/internal/xcode_schemes_internal.bzl
```diff
@@ -163,7 +163,8 @@ def _launch_action(
         build_configuration_name,
         args = None,
         env = None,
-        working_directory = None):
+        working_directory = None,
+        diagnostics = None):
```
brentleyjones, line 167: Above env?

xcodeproj/internal/xcode_schemes_internal.bzl
```diff
@@ -176,6 +177,7 @@ def _launch_action(
             environment variables when the target is executed.
         working_directory: Optional. A `string` that will be set as the custom
             working directory in the Xcode scheme's launch action.
+        diagnostics: Optional. A `struct` representing Xcode scheme's diagnostics.
```
brentleyjones, line 180: Above env?

xcodeproj/internal/xcode_schemes.bzl
```diff
@@ -303,6 +342,7 @@ def make_xcode_schemes(bazel_labels):
             args = args,
             env = env,
             working_directory = working_directory,
+            diagnostics = diagnostics,
```
brentleyjones, line 345: Above env?

tools/generator/src/Generator/XCSchemeInfo+LaunchActionInfo.swift
```diff
@@ -48,7 +51,8 @@ extension XCSchemeInfo.LaunchActionInfo {
             ),
             args: original.args,
             env: original.env,
-            workingDirectory: original.workingDirectory
+            workingDirectory: original.workingDirectory,
+            diagnostics: original.diagnostics
```
brentleyjones, line 55: Above env?

tools/generator/src/Generator/XCSchemeInfo+LaunchActionInfo.swift
```diff
@@ -7,13 +7,15 @@ extension XCSchemeInfo {
...
+        let diagnostics: DiagnosticsInfo?
 
         init(
             buildConfigurationName: String,
             targetInfo: XCSchemeInfo.TargetInfo,
             args: [String] = [],
             env: [String: String] = [:],
-            workingDirectory: String? = nil
+            workingDirectory: String? = nil,
+            diagnostics: DiagnosticsInfo? = nil
```
brentleyjones, line 18: Above env?

xcodeproj/internal/xcode_schemes.bzl
```diff
@@ -276,11 +276,49 @@ def make_xcode_schemes(bazel_labels):
...
+
+        Args:
+            sanitizers: Optional. A `struct` value as created by
+                `xcode_schemes.sanitizers`.
+
+        Returns:
+            A `struct` representing scheme's diagnostics.
+        """
+        return struct(
+            sanitizers = sanitizers,
```
brentleyjones, line 313: How does the `Decodable` handle the `None` here?

xcodeproj/internal/xcode_schemes.bzl
brentleyjones, line 321: How does the `Decodable` handle the `None` here?

xcodeproj/internal/xcode_schemes.bzl
chiragramani: Thanks for pointing this out. My earlier implementation would have errored out if sanitizers was `None`. I made sanitizers optional so now the project generation works fine.

xcodeproj/internal/xcode_schemes.bzl
chiragramani: We have kept `diagnostics` as an optional property in the generator so if diagnostics is `None`, diagnostics will be `nil` when decoded and the sanitizers default state(in the generated scheme) will be `False`.

PR #1219, opened 2022-10-04
Move swiftc stub to //tools

erikkerber:
~Pedantry in this context, but this did trip our Swiftlint rules for `print` within the generated project.~

~Low confidence in the choice for subsystem and category fields.~

This now simply moves the swiftc stub out of bazel_integration_files and makes it a tool, so it won't be bundled into the generated project


xcodeproj/internal/bazel_integration_files/main.swift
```diff
@@ -114,7 +116,7 @@ for arg in args {
         let matches = try NSRegularExpression(pattern: #".*?/Contents/Developer)/.*"#)
             .matches(in: sdkPath, range: fullRange)
         guard let developerDir = matches.first else {
-            print("warning: Failed to parse DEVELOPER_DIR from '-sdk'")
+            logger.log("warning: Failed to parse DEVELOPER_DIR from '-sdk'")
```
brentleyjones, line 119: This needs to be a print though.

This file shouldn't get bundled in the project, so that's the main issue, right?

xcodeproj/internal/bazel_integration_files/BUILD
```diff
@@ -37,45 +26,6 @@ filegroup(
     visibility = ["//visibility:public"],
 )
```
brentleyjones, line 27: Can you remove the `main.swift` line mentioned above this?

brentleyjones (approved):

brentleyjones (approved):

PR #1220, opened 2022-10-04
Apply the same `env -i` command in the runner

brentleyjones:
This is to ensure the generator run and the in-Xcode build are as similar as possible.

### DIFF
--- a/xcodeproj/internal/runner.template.sh
+++ b/xcodeproj/internal/runner.template.sh
@@ -43,6 +43,13 @@ done
 
 cd "$BUILD_WORKSPACE_DIRECTORY"
 
+# Resolve path to bazel before changing the env variable. This allows bazelisk
+# downloaded bazel to be found. This won't call `tools/bazel` if it exists,
+# which we should find a fix for (e.g. by finding a way to resolve to bazelisk
+# instead of the downloaded bazel).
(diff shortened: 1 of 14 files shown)


tinder-maxwellelliott (approved):

maxwellE (approved):

xcodeproj/internal/runner.template.sh
```diff
@@ -68,12 +68,21 @@ touch /tmp/rules_xcodeproj/WORKSPACE
 echo 'exports_files(["top_level_cache_buster"])' > /tmp/rules_xcodeproj/BUILD
 date +%s > "/tmp/rules_xcodeproj/top_level_cache_buster"
 
+bazel_cmd=(
+  env -i
+  HOME="$HOME"
+  PATH="/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
```
keith, line 74:
```suggestion
  PATH="/usr/bin:/bin:/usr/sbin:/sbin:/opt/homebrew/bin:/usr/local/bin"
```

I would recommend this so system binaries get preferred over brew

brentleyjones: :+1: I'll do that in the next PR.

PR #1222, opened 2022-10-05
Apply environment variables only if they exists

thii:
The runner is run via bazel, which may have limited environment
variables exposed if we use a bazel wrapper.

This fixes the following error for us:

```
bazel-output-base/execroot/__main__/bazel-out/darwin_arm64-dbg/bin/xcodeproj-bazel-sim_arm64-Dev-runner.sh:
line 83: USER: unbound variable
```


thii:
@brentleyjones Any ideas why the doc diff test failed? It passed locally for me.

xcodeproj/internal/runner.template.sh
```diff
@@ -74,10 +74,10 @@ unset BAZELISK_SKIP_WRAPPER
 
 readonly bazel_cmd=(
   env -i
-  HOME="$HOME"
+  HOME="${HOME:-}"
   PATH="/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin"
-  TERM="$TERM"
-  USER="$USER"
+  TERM="${TERM:-}"
+  USER="${USER:-}"
```
brentleyjones, line 80: Instead let's create an `env_values` array with the always set ones, then check for values and add them to the array, and finally expand the array.

brentleyjones:
I'm not sure on the docs job. Seems to be a CI flake. I'll merge it when needed.

brentleyjones (approved):

PR #1223, opened 2022-10-05
Add a new output group for generated source files

thii:
This ensures that generated source files are downloaded when buiding
with "Builds without the Bytes".

### DIFF
--- a/xcodeproj/internal/bazel_integration_files/bazel_build.sh
+++ b/xcodeproj/internal/bazel_integration_files/bazel_build.sh
@@ -21,8 +21,8 @@ if [ "$ACTION" == "indexbuild" ]; then
     # Inputs for compiling
     readonly output_group_prefixes="xc"
   else
-    # Compiled outputs (i.e. swiftmodules)
-    readonly output_group_prefixes="bc"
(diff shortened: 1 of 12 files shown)


xcodeproj/internal/bazel_integration_files/bazel_build.sh
```diff
@@ -29,9 +29,9 @@ else
     # Inputs for compiling, inputs for linking, and index store data
     readonly output_group_prefixes="xc,xl,xi"
   else
-    # Compiled outputs (i.e. swiftmodules), products (i.e. bundles), and index
-    # store data
-    readonly output_group_prefixes="bc,bp,bi"
+    # Compiled outputs (i.e. swiftmodules), products (i.e. bundles), generated
+    # inputs, and index store data
+    readonly output_group_prefixes="bc,bp,bg,bi"
```
brentleyjones, line 34: We need to add this to the `indexbuild` section above as well, to ensure that when Index Build is done source files exist for indexing jobs.

brentleyjones (approved):
I'm surprised I missed this, as this can lead to some indexing issues. Thanks!

PR #1225, opened 2022-10-05
Fix swiftc stub for incremental compilation

brentleyjones:
We need to parse the JSON in order to determine the exact paths to dependency files, since Xcode adds a hash to the end of partials.

Also apply some code cleanup.

### DIFF
--- a/tools/swiftc_stub/main.swift
+++ b/tools/swiftc_stub/main.swift
@@ -39,26 +39,28 @@ func touchDepsFiles(_ args: OrderedSet<String>) throws {
     else { return }
 
     if isWMO(args: args) {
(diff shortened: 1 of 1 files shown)


tools/swiftc_stub/main.swift
```diff
@@ -39,26 +39,29 @@ func touchDepsFiles(_ args: OrderedSet<String>) throws {
...
+        let data = try Data(contentsOf: outputFileMapPath)
+        let outputFileMapRaw = try JSONSerialization.jsonObject(
+            with: data,
+            options: []
+        )
+        guard let outputFileMap = outputFileMapRaw as? [String: [String: Any]]
+        else {
+            return
         }
+        print(outputFileMap)
```
brentleyjones, line 58: Will remove

thii (approved):

PR #1232, opened 2022-10-05
Apply sanitizer features to both normal and SwiftUI Previews builds

brentleyjones:
Also apply ubsan when alongside other ones if applicable.


brentleyjones:
FYI @chiragramani.

PR #1238, opened 2022-10-07
Remove `Copy Bazel Outputs` scripts from non-product targets

brentleyjones:
The idea about keeping them around was that we would use them for copying diagnostics and then modifying a file to force a "recompile". There are other ways to achieve that though, so these scripts are just plain overhead right now.


brentleyjones:
This regressed Index build, since we no longer get `$SCHEME_TARGET_IDS_FILE` updated.

PR #1260, opened 2022-10-10
Set `PROJECT_DIR` to Bazel's execution root

brentleyjones:
By anchoring the project in Bazel's execution root, we can reduce the amount of path manipulation we need to perform on Bazel provided compiler and linker flags. This will also allow us to support multiple output bases for various modes (e.g. SwiftUI Previews or Runtime Sanitizers), as the `.sourceRoot` (really `PROJECT_DIR`) relative `PBXBuildFile`'s will adjust as we change `PROJECT_DIR` (which we can change based on things like `ENABLE_PREVIEWS`).


brentleyjones:
I think this broke "simple projects" (BwX without `BazelDependencies`). I'll get a fix in soon.

brentleyjones:
I was wrong, they are fine.

PR #1262, opened 2022-10-11
Revert "Remove `Copy Bazel Outputs` scripts from non-product targets (#1238)"

brentleyjones:
This reverts commit 5f8c7df80a07970289ac9a7fd638ab37c2363e27.

The change broke "Off-by-one Index Build". We will have to figure out how to do this while still preserving that functionality.


brentleyjones:
FYI @thii.

PR #1264, opened 2022-10-11
Move nested output base down into `execroot`

brentleyjones:
This way it's still cleaned with `bazel clean`, but there is no chance of recursive symlinks.

The name has an underscore, so it shouldn't conflict with another external repo when using `--experimental_sibling_repository_layout`.


brentleyjones:
I like this change regardless if it fixes the issue, so I'm going to merge it. Let me know @erikkerber if it's still broken after a `bazel clean` and I have another trick up my sleeve to try.

PR #1276, opened 2022-10-12
Propagate sanitizer related flags to build command

chiragramani:
On the latest main, though the application builds but the feature flags are not propagated to the build command. This PR updates the bazel_build.sh to apply the `build_pre_config_flags`.

It also, fixes the reference to "INTERNAL_DIR" in the scheme's pre-action.


brentleyjones (approved):
Thanks! Sorry about that.

PR #1291, opened 2022-10-13
Really really fix symlink issue

brentleyjones:
I don't know how to make a single invocation work for this thing, so going old-school and manually removing it beforehand.


thii:
Does this work? `ln -shfF`

brentleyjones:
Ohh, probably. I'll try that locally and if it stays fixed for a while I'll commit it.

PR #1294, opened 2022-10-13
Convert VariableExpansionContextInfo types to Optional<T>

### LINKED ISSUE #1118: Cleanup: Remove use of `.none` enum values in favor of optional values
> Assuming you mean 'Optional<XcodeScheme.VariableExpansionContext>.none'; did you mean 'XcodeScheme.VariableExpansionContext.none' instead?

`XcodeScheme.VariableExpansionContext` and `XcodeScheme.VariableExpansionContextInfo` should be removed in favor of `Optional<BazelLabel>` and `Optional<TargetInfo>`.

erikkerber:
Resolves #1118 

image


tools/generator/src/Generator/XCSchemeInfo+PrePostActionInfo.swift
```diff
@@ -55,11 +53,11 @@ extension Sequence where Element == XCSchemeInfo.PrePostActionInfo {
         topLevelTargetInfos: TargetInfos
     ) throws -> [XCSchemeInfo.PrePostActionInfo]
     where TargetInfos.Element == XCSchemeInfo.TargetInfo {
-        try map { action in
+        map { action in
             XCSchemeInfo.PrePostActionInfo(
                 name: action.name,
-                expandVariablesBasedOn: try .init(
-                    resolveHostsFor: action.expandVariablesBasedOn,
+                expandVariablesBasedOn: try? .init(
```
brentleyjones, line 59: I don't think we want `try?` here.

tools/generator/src/Generator/XCSchemeInfo+TestActionInfo.swift
```diff
@@ -65,8 +64,8 @@ extension XCSchemeInfo.TestActionInfo {
             args: original.args,
             diagnostics: original.diagnostics,
             env: original.env,
-            expandVariablesBasedOn: .init(
-                resolveHostsFor: original.expandVariablesBasedOn,
+            expandVariablesBasedOn: try? .init(
```
brentleyjones, line 67: Same

tools/generator/src/Generator/XCSchemeInfo+PrePostActionInfo.swift
erikkerber: This was the one line I was debating about. The `try?` is just paired with the `orThrow()` to propagate a nil value while keeping this map a single expression.

brentleyjones: But the `.init` could throw for other reasons, and we want those to not be swallowed.

erikkerber: `TargetInfo.init` is not a throwing initializer. If you mean it could be turned into one in the future — I agree this would make it hard to detect.

I could change this to be a `guard let ... else { return action }` and it would do the (I believe) desired behavior of not mapping that value to anything new iff there isn't a `expandVariablesBasedOn`/`TargetInfo` value.

erikkerber: Let me know what you think of 4400cda214c019a6ec3bf1d0b37a116a60f01cea

brentleyjones (approved):

PR #1303, opened 2022-10-14
WIP: Parsing env variables for launch actions

maxwellE:
Starts the work of extracting env values from *binary targets for launch action schemes


examples/integration/.bazelrc
```diff
@@ -11,6 +11,7 @@ build:rules_xcodeproj --@com_github_buildbuddy_io_rules_xcodeproj//xcodeproj:ext
 build:rules_xcodeproj_integration --define=foo=bar
 
 build --test_env=INTEGRATION_TEST_ENV_VAR=TRUE
+build --action_env=INTEGRATION_ACTION_ENV_VAR=TRUE
```
brentleyjones, line 14: I don't believe this is used the same way as `test_env`, and is only exposed to build actions. Let's remove it.

PR #1307, opened 2022-10-17
Remove use of output group maps

brentleyjones:
Output group maps were a hack used to ensure that all intermediate files we needed were downloaded when using Build without the Bytes. With Bazel 5.3, we now have `--experimental_remote_download_regex`, which allows us to download intermediate files matching a given regex. That, plus adjustment to our output groups, ensure that we continue to download the same files.

Why do we want to make this change now though? There are a couple downsides to using output group maps:

- All downloads happen at the end of the build instead of in the middle
- Bazel can run into memory or thread issues with the way that the output maps download everything at once
- They require the use of the "top level cache buster", which is a file we modified every build. This required some nasty hacks to handle in the general case, but more importantly, it caused re-evaluation of the `xcodeproj` implementation function each build, which could take many seconds for larger projects


xcodeproj/internal/xcodeproj.template.bazelrc
```diff
@@ -15,6 +15,9 @@ build:rules_xcodeproj --experimental_action_cache_store_output_metadata
 # Xcode builds shouldn't mess with the workspace's symlinks
 build:rules_xcodeproj --experimental_convenience_symlinks=ignore
 
+# Ensure all our intermediate files are downloaded with BwtB
+build:rules_xcodeproj --experimental_remote_download_regex=.*\\.indexstore/.*|.*\\.swiftdoc$|.*\\.swiftmodule$|.*\\.swiftsourceinfo$
```
thii, line 19: Is there a way to filter this for just the target platform? I think this would also force downloading of Swift tools' swiftmodules etc. (though I don't think it would make any significant difference)

brentleyjones: I'll look into that :+1:

xcodeproj/internal/bazel_integration_files/bazel_build.sh
```diff
@@ -152,31 +146,28 @@ touch "$build_marker"
...
-    echo "error: If your bazel version is less than 5.2, you may need to" \
-"\`bazel clean\` and/or \`bazel shutdown\` to work around a bug in project" \
-"generation." >&2
-    echo "error: If you are still getting this error after all of that," \
-"please file a bug report here:" \
+  if [[ ! -f "$filelist" ]]; then
+    echo "error: Bazel didn't create the indexstore filelist (\"$filelist\")." \
+"Please regenerate the project to fix this." >&2
+    echo "error: If you are still getting this error after regenerating your" \
+"project, please file a bug report here:" \
```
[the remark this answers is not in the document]
brentleyjones: I'll make it a single error.

xcodeproj/internal/filelists.bzl
```diff
@@ -0,0 +1,46 @@
...
+        command = """
+if [[ $OSTYPE == darwin* ]]; then
+  cp -c \"$1\" \"$2\"
+else
+  cp \"$1\" \"$2\"
+fi
+""",
+        arguments = [
+            args,
+            output.path,
```
thii, line 24: Probably overkill but we could also put `output` in another `Args` object.

brentleyjones: I don't see the benefit. The existing `Args` object is used in order to get the `.params` file from it, and to prevent expanding the `files` `Depset`. `output` on the other-hand doesn't have that issue, and putting it in an `Args` object just to get unwrapped will be unnecessary overhead, right?

brentleyjones: Ahh, so it reduces memory usage, for when this action isn't run. I didn't think of that. I'll make that change.

thii (approved):

BalestraPatrick (approved):
Works in our testing!

brentleyjones:
If we notice any BwtB issues after merging, we still have time before the 0.10.0 release to fix them.

PR #1288, opened 2022-10-12
Add fewer target dependencies in BwB mode

brentleyjones:
This improves the performance in BwB mode by preventing Xcode from running scripts and stubs that aren't needed. It is also a prerequisite to fix #1054. This change will also allow target merging happen for more targets in BwB mode, which makes the SwiftUI Previews experience better.

PR #1312, opened 2022-10-18
Repro: Unable to use swiftUI previews

maxwellE:
This reproducer highlights a potential issue when
using swiftUI for a target that shares a scheme
with a unit test target for the same underlying
target.

Removing the test target from the scheme fixes the issue and allows swiftuI to work:

xcode log:

```terminal
UnsupportedProductTypeError: Previews not supported for SomeModule

Previews can be defined in applications, frameworks, Swift packages, or dynamic libraries, but not in static libraries

productType = com.apple.product-type.library.static
buildable = SomeModule
```

### DIFF
--- a/examples/integration/.bazelrc
+++ b/examples/integration/.bazelrc
@@ -12,5 +12,9 @@ build:rules_xcodeproj_integration --define=foo=bar
 
 build --test_env=INTEGRATION_TEST_ENV_VAR=TRUE
 
+build --ios_minimum_os=14.0
+build --features apple.arm64_simulator_use_device_deps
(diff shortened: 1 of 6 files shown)


brentleyjones:
The issue here is that `SomeModule` can't merge into `SomeModuleFramework.iOS`, because it's a transitive dependency of `SomeModuleTests`. Adding `frameworks = [":SomeModuleFramework.iOS"]` to `SomeModuleTests` fixes this.

After #1288, we will be able to adjust the BwB target merging logic, and this will work without changes.

brentleyjones:
Closing since I believe the comment above addresses this.

PR #1335, opened 2022-10-20
Use `ZippyJSONDecoder` instead of `JSONDecoder`

brentleyjones:
Currently results in about a 2.5x speedup. For a specific large real world application this saves over 5 seconds of generation time.

| Before | After |
| --- | --- |
| JSONDecoder taking 9.27s in Instruments |  ZippyJSONDecoder taking 3.6s in Instruments |


BalestraPatrick (approved):
⚡️

PR #1344, opened 2022-10-21
Add cache warming xcodeproj example

thii:
This adds an xcodeproj cache warming example, with instructions on how
to set up a workspace to make sure cache warming builds can run on any
macOS platform to produce cache for any platforms.


examples/integration/BUILD
```diff
@@ -15,16 +15,51 @@ load(
...
+#
+#    bazel run //:xcodeproj-x86_64 \
+#      --@com_github_buildbuddy_io_rules_xcodeproj//xcodeproj:extra_common_flags=--remote_download_minimal \
+#      -- \
+#      --build_output_groups all_targets
+#
+#    bazel run //:xcodeproj-sim_arm64 \
+#      --@com_github_buildbuddy_io_rules_xcodeproj//xcodeproj:extra_common_flags=--remote_download_minimal \
+#      -- \
+#      --build_output_groups all_targets
```
brentleyjones, line 42: Since this is about to change, and I'll be adding it to `usage.md`, I'm going to wait to merge this PR.

brentleyjones (commented):
@thii Do you know why the Starlark hash changes with this?

thii:
(I ran`bazel cquery //:xcodeproj-sim_arm64` without and with that flag to get the config)

examples/integration/BUILD
```diff
@@ -15,16 +15,51 @@ load(
...
-    scheme_autogeneration_mode = SCHEME_AUTOGENERATION_MODE,
-    schemes = get_xcode_schemes(),
-    tags = ["manual"],
-    top_level_targets = XCODEPROJ_TARGETS,
-    unfocused_targets = UNFOCUSED_TARGETS,
-)
+# Example of xcodeproj cache warming
+#
+# 1. Register a single macOS execution platform in WORKSPACE
+#    register_execution_platforms("@build_bazel_apple_support//platforms:macos_x86_64")
```
brentleyjones, line 21: This makes the path the same for both (`x86_64`), does it actually force anything to be non-fat `x86_64` (after accounting for `--@build_bazel_rules_swift//swift:universal_tools`)? Like `wrapped_clang` or any other host tools?

brentleyjones:
I pushed changes to the PR to match the new API, along with some small tweaks. lmk if it still looks good to you @thii.

examples/integration/BUILD
```diff
@@ -15,16 +15,54 @@ load(
...
-    top_level_targets = XCODEPROJ_TARGETS,
-    unfocused_targets = UNFOCUSED_TARGETS,
-)
+# Example of xcodeproj cache warming
+#
+# 1. Register a single macOS execution platform in `WORKSPACE`:
+#
+#      register_execution_platforms("@build_bazel_apple_support//platforms:macos_x86_64")
+#
+# 2. Create a `platform_mappings` file.
```
erikkerber, line 24: I was unfamiliar with this [until Googling it](https://bazel.build/concepts/platforms#platform-mappings).

Is this something we could do for users via `--platform_mappings`? I admit as someone who <read above>, I don't know if I'll know if/when I need to update those mappings.

examples/integration/BUILD
```diff
@@ -15,16 +15,54 @@ load(
...
-    schemes = get_xcode_schemes(),
-    tags = ["manual"],
-    top_level_targets = XCODEPROJ_TARGETS,
-    unfocused_targets = UNFOCUSED_TARGETS,
-)
+# Example of xcodeproj cache warming
+#
+# 1. Register a single macOS execution platform in `WORKSPACE`:
+#
+#      register_execution_platforms("@build_bazel_apple_support//platforms:macos_x86_64")
```
erikkerber, line 22: Would somebody need to register two execution platforms if they had Apple Silicon on CI?

examples/integration/BUILD
```diff
@@ -15,16 +15,54 @@ load(
...
+# 4. Define an `xcodeproj` target for each macOS platform, e.g.
+#    `xcodeproj-x86_64` and `xcodeproj-sim_arm64`. Even if you don't need to
+#    support any of iOS, tvOS, or watchOS platforms, you still need to set all
+#    of the `ios_simulator_cpus`, `tvos_simulator_cpus`, and
+#    `watchos_simulator_cpus` to avoid Starlark transition effect. See
+#    https://github.com/buildbuddy-io/rules_xcodeproj/issues/1094.
+# 5. Run each of the following commands to produce a cache for each macOS host
+#    platform. This can be run from an Intel or Apple silicon Mac.
+#
+#      bazel run --config=cache //:xcodeproj-x86_64 \
```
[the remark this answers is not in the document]
brentleyjones: It maps to the config that is in this repo.

examples/integration/BUILD
thii: This forces any non-fat host tools to be compiled as `x86_64`, so at least they still continue to run if you have Rosetta installed on your (remote or local) executor.

Since the purpose is to ensure a "macos_universal" platform, you can also use `register_execution_platforms("@build_bazel_apple_support//platforms:macos_arm64")` instead, or don't install Rosetta at all, and the build will error out if you have any non-fat tools.

examples/integration/BUILD
thii: Hmm. I expect users to be aware of this if they want to implement cache sharing between Intel and Apple silicon. I don’t think `rules_xcodeproj` should pass this on behalf of users because platform mappings is only a temporary API (until `rules_apple` fully supports building with platforms), and users may also use it for other kinds of builds. The `platform_mappings` file here is just an example, users may have their own definitions of `platform` targets and `platform_mappings` file, for instance, because of different `exec_properties` requirements from their RBE provider.

examples/integration/BUILD
thii: No. AFAIK currently you can only have one execution platform for Apple builds. Even if you define more than one execution platform, only the first one will be used. If you don't define anything, the host platform (`@local_config_platform//:host`) will be used, which is host-dependent. The purpose of this is to use a host-independent platform to build.

brentleyjones (approved):

PR #1350, opened 2022-10-21
Add official command-line API

brentleyjones:


BalestraPatrick (approved):

docs/usage.md
```diff
@@ -112,3 +122,202 @@ flags were used during project generation, then those adjustments are made in
...
+
+### `clean`
+
+When you run `bazel clean` normally (i.e. not using this API), it won't affect
+the rules_xcodeproj output base the way you expect. `bazel clean --expunge`
+will though, as it will blow away both environments. To clean the
+rules_xcodeproj output base use the API instead:
+
+```
+bazel run //:xcodeproj -- clean
```
thii, line 220: 👍

docs/usage.md
```diff
@@ -112,3 +122,202 @@ flags were used during project generation, then those adjustments are made in
...
+
+Depending on how you have your [rules_xcodeproj Bazel configs](#bazel-configs)
+set up, you might be able to run `bazel query` without using the API. I
+recommend using the API instead though, to prevent fetching external
+dependencies in the primary output base. The other queries, `cquery` and
+`aquery`, should always be performed through the API, to ensure the targets
+are properly configured:
+
+```
+bazel run //:xcodeproj -- aquery 'set(//some:target)'
```
thii, line 233: This is nice, but it changes the semantic way of using bazel. I don't think we (me) would want to train people doing the-rules_xcodeproj-way-of-using-bazel.

thii (approved):

docs/usage.md
brentleyjones: I'm not sure of a way around it. If I don't mention this the error you get is hard to understand, so I have to say _something_.

docs/usage.md
```diff
@@ -112,3 +122,202 @@ flags were used during project generation, then those adjustments are made in
...
+
+```
+bazel run //:xcodeproj -- [option ...] command [command_flag ...]
+```
+
+For example, this will call `bazel info output_base` in the rules_xcodeproj
+environment:
+
+```
+bazel run //:xcodeproj -- info output_path
```
jszumski, line 151: Should this match the sentence above?
```suggestion
bazel run //:xcodeproj -- info output_base
```

docs/usage.md
```diff
@@ -112,3 +122,202 @@ flags were used during project generation, then those adjustments are made in
...
+this API, because the `xcodeproj` rule applies a
+[configuration transition](https://bazel.build/extending/config#user-defined-transitions)
+to targets. That means that building targets by specifying their labels will
+build potentially different versions of those targets, and minimally versions
+that have different cache keys.
+
+rules_xcodeproj uses
+[output groups](https://bazel.build/extending/rules#requesting_output_files)
+to "address" these correctly configured targets. It uses a set of private
+output groups, but it also exposes the some [public ones](#output-groups).
```
jszumski, line 184:
```suggestion
output groups, but it also exposes some [public ones](#output-groups).
```

docs/usage.md
brentleyjones: Thinking more about this, I could have the command portion quoted, similar to when one does `sh -c`. Then it would be something like `bazel run //:xcodeproj -- 'query //some:target'` which bazel wouldn't mess with.

PR #1357, opened 2022-10-25
Make bazel integration files copying more general

brentleyjones:
This allows us to include generated files, like the swiftc stub, without special casing.


xcodeproj/internal/bazel_integration_files/BUILD
```diff
@@ -26,6 +26,19 @@ filegroup(
     visibility = ["//visibility:public"],
 )
 
+genrule(
+    name = "swiftc_stub",
+    srcs = ["//tools/swiftc_stub:universal_swiftc_stub"],
+    outs = ["swiftc"],
+    # Make the swiftc_stub have the right name
+    cmd = "cp -c $< $@",
+    message = "Renaming swiftc stub",
```
thii, line 35: Where does this requirement come from? I see that Tulsi still uses the name [`swift_stub.py`](https://github.com/bazelbuild/tulsi/blob/4ca229461f41488292f2138c7675dcc530d3bee3/src/TulsiGenerator/Scripts/swiftc_stub.py#L1) and has no problem with it.

thii (approved):

xcodeproj/internal/bazel_integration_files/BUILD
brentleyjones: Xcode 14 requires the name to be `swiftc` in some setups. We sidestep it a bit right now because we use `SWIFT_USE_INTEGRATED_DRIVER=NO`, but without that, `SWIFT_EXEC` needs to be named this.

PR #1360, opened 2022-10-26
Fix horrible BwX `FilePathResolver` performance

### LINKED ISSUE #1359: Bug: Very slow BwX generation times
## Describe the bug

Generating with BwX/xcode is an order of magnitude slower than BwB on our project. ~1 minute vs ~10 minutes

## Reproduction steps

Build a project with BwB then BwX and observe the generation time differences. I am working on a more specific repro if this does not happen on all projects.

## Expected behavior

🐇 

## Versions
- Bazel: 6.0.0rc1
- rules_xcodeproj: ff51b809b50ae96cdabbe02925f4391087dd0a91
- rules_apple: 1.1.2
- rules_swift: 1.2.0

brentleyjones:
Fixes #1359.

Regressed with 5febe757f6c771e32b82bf640f3db5b2f65d34ab.

We were hashing `FilePathResolver` as part of `MemoizationKey`, to ensure that if there were multiple versions of `FilePathResolver` then the values wouldn't conflict. That itself isn't much of an issue, except in BwX mode that would include a large `xcodeGeneratedFiles` dictionary. Changing `FilePathResolver` to a class and embedding the memoization into the instance side-steps this issue. In the future `FilePathResolver` will be in Starlark, removing all of this anyway...

### DIFF
--- a/tools/generator/src/FilePathResolver.swift
+++ b/tools/generator/src/FilePathResolver.swift
@@ -1,24 +1,13 @@
 import PathKit
 
-// TODO: Make thread safe if we ever go concurrent
-var memoizedPaths: [MemoizationKey: Path] = [:]
-
-struct MemoizationKey: Equatable, Hashable {
-    let filePath: FilePath
-    let transformedFilePath: FilePath
-    let useBazelOut: Bool?
-    let forceFullBuildSettingPath: Bool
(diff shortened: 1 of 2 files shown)


brentleyjones:
Tests expose that _there_ it's not a singleton. Going to convert to a class and embed the memoization.

PR #1361, opened 2022-10-26
Fix `local_repository` regression

brentleyjones:
Regressed with 5aec77f2831a70164406d461811bedab7a46e6aa.

`.srcRoot` now means Bazel's execution root. So just like `mainGroup`, we need this to be an absolute path all the time now.

### DIFF
--- a/tools/generator/test/GeneratorTests.swift
+++ b/tools/generator/test/GeneratorTests.swift
@@ -340,6 +340,7 @@ final class GeneratorTests: XCTestCase {
         func createFilesAndGroups(
             in pbxProj: PBXProj,
             buildMode: BuildMode,
+            _forFixtures: Bool,
             forceBazelDependencies: Bool,
             targets: [TargetID: Target],
             extraFiles: Set<FilePath>,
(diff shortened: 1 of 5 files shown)


tools/generator/test/GeneratorTests.swift
```diff
@@ -340,6 +340,7 @@ final class GeneratorTests: XCTestCase {
         func createFilesAndGroups(
             in pbxProj: PBXProj,
             buildMode: BuildMode,
+            _forFixtures: Bool,
```
jpsim, line 343: Where is this used? I'd expect a change at the call site in this PR, but don't see any. Maybe there's some magic happening that's eliding the parameter label at the call site for some reason?

jpsim (approved):
Fixes the issues I was with indexing earlier today

tools/generator/test/GeneratorTests.swift
brentleyjones: This `createFilesAndGroups()` is called inside of `generate()`, where it is passed the constant `false` down below. I didn't feel it was worth tracking, so here and the previous spot I just add the parameter as needed, but I don't track it in the `XYZCalled` structs.

PR #1385, opened 2022-11-02
Set `PROJECT_DIR` to `$SRCROOT` for `BazelDependencies`

brentleyjones:
Also:

- Create `$BAZEL_EXECUTION_ROOT` during Index Build, to allow `Copy Bazel Outputs / Generate Bazel Dependencies (Index Build)` script to work even after `bazel clean`
- Use `$BAZEL_EXECUTION_ROOT` instead of `$PROJECT_DIR` where that is what we meant


thii:
It still breaks before that, because we have a pre-build build phase.

thii:
But only the first two index builds, after that all worked.

thii:
Looks like it's because with index build, Xcode still runs the next build phase (Bazel Build) even though the previous one (Pre-build) fails.

brentleyjones:
Hmm, it shouldn't break in the pre-build script, since `BazelDependencies` sets `PROJECT_DIR` to `$(SRCROOT)`, so the pre-build run script shouldn't have an issue with `$BAZEL_EXECUTION_ROOT` (previously `$PROJECT_DIR`) not existing.

thii:
`PROJECT_DIR` is set to `$(SRCROOT)`, but what I see in the index build log is:

```
PhaseScriptExecution Pre-build\ Run\ Script /Users/user/Library/Developer/Xcode/DerivedData/Projects-fypotnookqgqmegzxfveuwcvjqgb/Index/Build/Intermediates.noindex/Projects.build/rules_xcodeproj/Script-1FE315C2E97C99DF62676716.sh (in target 'BazelDependencies' from project 'Projects')
    cd /Users/user/dev/ios-app/bazel-output-base/rules_xcodeproj/build_output_base/execroot/__main__
    /bin/sh -c /Users/user/Library/Developer/Xcode/DerivedData/Projects-fypotnookqgqmegzxfveuwcvjqgb/Index/Build/Intermediates.noindex/Projects.build/rules_xcodeproj/Script-1FE315C2E97C99DF62676716.sh
```

brentleyjones:
Crap, I see that too now. So this change doesn't do what I expected. Maybe I can keep parts of it, but it doesn't solve the main issue 😕.

brentleyjones:
Closing for now.

PR #1387, opened 2022-11-04
Fix debugging of mixed-lanugage module

brentleyjones:
We don't want the default modulemap for the `objc_library` to be generated, as it results in duplicated modulemaps in the the `target.swift-extra-clang-flags` setting.


brentleyjones:
FYI @thii @maxwellE.

tinder-maxwellelliott:
> FYI @thii @maxwellE.

Thanks

PR #1480, opened 2022-12-16
Improve parallelism of `Target` parsing

brentleyjones:


brentleyjones:
Still slow for two reasons:
- We are waiting for dealloc of the project for some reason
- Merging takes a long time, and we really just need an append

brentleyjones:
Still slight improvements here, so merging.

PR #1488, opened 2022-12-19
Include Robinhood as consumers

sebastianv1:
Thank you for the great ruleset!

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@ rules.
 ## Projects using rules_xcodeproj
 
 - [Envoy Mobile](https://github.com/envoyproxy/envoy-mobile)
+- Robinhood
 - Spotify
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +54,7 @@ rules.
 - [Envoy Mobile](https://github.com/envoyproxy/envoy-mobile)
 - Spotify
 - [SwiftLint](https://github.com/realm/SwiftLint)
+- [Robinhood](https://github.com/robinhoodmarkets)
```
brentleyjones, line 57: Can you put this before the `S` eateries? Thanks!

Also, the links are to show public implementations or proof of use (like blog posts or talks), so can you remove the link as well?

brentleyjones (approved):

README.md
sebastianv1: Yeah makes sense

PR #1502, opened 2022-12-22
Add support for bzlmod

BalestraPatrick:
This PR adds initial support for bzlmod. I still have to test it with an example project, but the basic infrastructure should be ready.

bzlmod is only really ready to be used starting in Bazel 6.0. For this reason, the `MODULE.bazel` depends on rules_apple 2.0.0, which is also the release that requires Bazel 6. I will need to first check the newest version of rules_apple and rules_swift into the Bazel Central Registry, and then we can proceed testing the rules_xcodeproj integration.

### DIFF
--- a/examples/cc/MODULE.bazel
+++ b/examples/cc/MODULE.bazel
@@ -1,0 +1,15 @@
+bazel_dep(name = "rules_apple", repo_name = "build_bazel_rules_apple", version = "2.0.0")
+bazel_dep(name = "bazel_skylib", version = "1.3.0")
+bazel_dep(name = "rules_xcodeproj", repo_name = "com_github_buildbuddy_io_rules_xcodeproj", version = "0.0.0")
+bazel_dep(name = "rules_swift", repo_name = "build_bazel_rules_swift", version = "1.5.1")
+bazel_dep(name = "examples_cc_external", version = "0.0.1")
+
+local_path_override(
+    module_name = "rules_xcodeproj",
+    path = "../.."
+)
+
+local_path_override(
+    module_name = "examples_cc_external",
+    path = "external"
+)
--- a/xcodeproj/extensions.bzl
+++ b/xcodeproj/extensions.bzl
@@ -1,0 +1,5 @@
+"""Module extension for loading dependencies not yet compatible with bzlmod."""
+
+load("//xcodeproj:repositories.bzl", "xcodeproj_rules_dependencies")
+
+non_module_deps = module_extension(implementation = lambda _: xcodeproj_rules_dependencies(include_bzlmod_ready_dependencies = False))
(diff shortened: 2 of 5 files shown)


brentleyjones:
FYI: https://github.com/buildbuddy-io/rules_xcodeproj/pull/1503

brentleyjones:
FYI, the repo now uses Bazel 6.0.

brentleyjones:
As part of this do we want to enable bzlmod in `bazel_6.bazelrc`? I think `local_provisioning_profile` breaks right now, so we might have to disable it for `examples/integration/.bazelrc`.

BalestraPatrick:
Yeah, I think that's a good idea. We might also need to create a `WORKSPACE.bzlmod` in order to bring in some of the other dependencies in an easier way, or just try to get everything done in the `MODULE.bazel`. I can take a look.

brentleyjones:
> We might also need to create a WORKSPACE.bzlmod in order to bring in some of the other dependencies in an easier way

If we can do that instead of disabling bzlmod, I would prefer that.

tools/bazel
```diff
@@ -12,4 +12,6 @@ else
   readonly bazel_version_bazelrc="$root_dir/bazel_6.bazelrc"
 fi
 
-exec "$BAZEL_REAL" "--bazelrc=$bazel_version_bazelrc" "$@"
```
brentleyjones, line 15: Can you open a PR with this change? Is disjoint from this PR.

BalestraPatrick: Yes, just trying to get this one green and then will split it up.

examples/cc/MODULE.bazel
```diff
@@ -0,0 +1,15 @@
+bazel_dep(name = "rules_apple", repo_name = "build_bazel_rules_apple", version = "2.0.0")
+bazel_dep(name = "bazel_skylib", version = "1.3.0")
+bazel_dep(name = "rules_xcodeproj", repo_name = "com_github_buildbuddy_io_rules_xcodeproj", version = "1.5.0")
```
brentleyjones, line 3: Version number is wrong. Should we put a placeholder like 0.0.0 here?

xcodeproj/extensions.bzl
```diff
@@ -0,0 +1,8 @@
+"""Module extension for loading dependencies not yet compatible with bzlmod."""
+
+load("//xcodeproj:repositories.bzl", "xcodeproj_rules_dependencies")
+
+def _non_module_deps_impl(_):
+    xcodeproj_rules_dependencies(include_bzlmod_ready_dependencies = False)
+
+non_module_deps = module_extension(implementation = _non_module_deps_impl)
```
brentleyjones, line 8: I think you can use a `lambda` for this.

brentleyjones (approved):

PR #1524, opened 2023-01-04
Run some Workflow actions on arm64 machines

brentleyjones:
Ideally not specifying the `arch` would use both archs, but until then, we manually set a few of the actions to run on arm64 instead of x86_64.


brentleyjones:
I'll increase the number of actions after a bug related to sizing is addressed.

PR #1525, opened 2023-01-04
Adjust Workflow resource requests and arm64 assignments

brentleyjones:
This will allow more actions per Workflow Mac.


brentleyjones:
I'm going to tweak the value per action, since I sorta know that some need less memory (like docs and linting).

PR #1530, opened 2023-01-06
Partially-revert 650c2ff49a4f0a8b1246d0d5f243414b1f15e76e

brentleyjones:
We still need the `swift_debug_settings.py` in the case of unfocused targets (as those are built with bazel). The part we aren't reverting is the incorrect setting of the vfs overlay.


erikkerber:
This still prints with `po` for us

PR #1569, opened 2023-01-12
Let through more Bazel placeholders

brentleyjones:


tools/params_processors/link_params_processor.py
```diff
@@ -117,15 +117,19 @@ def _process_linkopt(opt):
         if opt.endswith(".o"):
             return None
 
+        # Xcode adds system library search paths
+        if opt.startswith("-L__BAZEL_XCODE_"):
+            return None
+
```
brentleyjones, line 123: @keith There's no harm if these are set twice, right? I should probably just not filter it in case someone does something additional.

xcodeproj/internal/opts.bzl
```diff
@@ -221,7 +221,8 @@ def _process_base_compiler_opts(
         if skip_next:
             skip_next -= 1
             continue
-        if "__BAZEL_XCODE_" in opt:
+        if (opt.startswith("-F__BAZEL_XCODE_") or
+            opt.startswith("-I__BAZEL_XCODE_")):
```
brentleyjones, line 225: @keith There's no harm if these are set twice, right? I should probably just not filter it in case someone does something additional.

tools/params_processors/link_params_processor.py
keith: theoretically maybe the order matters in the case that multiple directories have the overlapping frameworks, but realistically shouldn't happen much

brentleyjones: https://github.com/buildbuddy-io/rules_xcodeproj/pull/1572

PR #1585, opened 2023-01-15
Add BCR app config

BalestraPatrick:
This PR adds the configuration files and templates necessary to get the [Publish To BCR](https://github.com/bazel-contrib/publish-to-bcr) GitHub app to work with rules_xcodeproj.

There are few steps to make sure this works correctly:
- [x] Approve my request to add the app to this repo.
- [x] Fork https://github.com/bazelbuild/bazel-central-registry under the account defined in `.bcr/config.yaml`. We can either remove the `fixedReleader` and have a fork of the BCR in the `buildbuddy-io` org, or we can use a personal fork. For example, for `rules_swift`, `rules_apple` and `apple_support` we're using [my own fork](https://github.com/balestrapatrick/bazel-central-registry) at the moment.
- [x] Test that the `.bcr/presubmit.yaml` is actually working. I think the first release should probably be checked-in manually to decide which targets to run as validation with bzlmod directly in the BCR.

### DIFF
--- a/.bcr/source.template.json
+++ b/.bcr/source.template.json
@@ -1,0 +1,5 @@
+{
+    "integrity": "",
+    "strip_prefix": "",
+    "url": "https://github.com/buildbuddy-io/rules_xcodeproj/releases/download/{TAG}/release.tar.gz"
+}
--- a/.bcr/config.yml
+++ b/.bcr/config.yml
@@ -1,0 +1,3 @@
+fixedReleaser:
+  login: BalestraPatrick
+  email: me@patrickbalestra.com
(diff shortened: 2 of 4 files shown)


brentleyjones:
Step 1 is done.

brentleyjones:
For step 2 we can use your fork.

.bcr/source.template.json
```diff
@@ -0,0 +1,5 @@
+{
+    "integrity": "",
+    "strip_prefix": "",
+    "url": "https://github.com/buildbuddy-io/rules_xcodeproj/releases/download/{TAG}/rules_xcodeproj.{TAG}.tar.gz"
```
brentleyjones, line 4: This isn't the release archive. Should probably be `https://github.com/buildbuddy-io/rules_xcodeproj/releases/download/{TAG}/release.tar.gz` instead?

brentleyjones (approved):

PR #1610, opened 2023-01-19
Add Lyft to users list

keith:


jpsim (approved):
+1 although I think the convention is to only have a link if it's relevant to rules_xcodeproj, such as an open source project where you can see how it's integrated, or a presentation about it.

jpsim (approved):

README.md
```diff
@@ -58,6 +58,7 @@ rules.
 - Spotify
 - [SwiftLint](https://github.com/realm/SwiftLint)
 - [Tinder](https://tinder.com)
+- Lyft
```
brentleyjones, line 61: nit: can you place this after `Envoy Mobile` to keep it alphabetical?

PR #1625, opened 2023-01-22
WIP: Asan config without O1

chiragramani:
https://github.com/buildbuddy-io/rules_xcodeproj/issues/1613

Bazel asan feature by default applies O1 but we might also want to catch violations for non-optimised code-paths. https://github.com/buildbuddy-io/rules_xcodeproj/issues/1613


brentleyjones:
Hmm, with this way we will get `-D_FORTIFY_SOURCE=1` because we won't hit `with_features = [with_feature_set(not_features = ["asan"])]`.

@keith I see you added this feature to Bazel. Why did you set `-O1`? Xcode doesn't set it.

Issue #1477, opened 2022-12-15
Feature Request: Possibility to profile the app in BwB mode


aleksandergrzyb:
Right now when generating Xcode project (with `--apple_generate_dsym` and `--output_groups=+dsyms` options) the `.dSYM` file is not placed in `DerivedData/<Project>/Build/Products/Debug-iphoneos` directory. This causes stack trace not being symbolicated when profiling the app on device.

Would be nice to have profiling functionality working in BwB mode. Thank you!

brentleyjones:
Note to self: probably will need `DBGSourcePathRemapping` as well: https://lldb.llvm.org/use/symbols.html#embedding-uuid-property-lists-inside-the-dsym-bundles

brentleyjones:
@CognitiveDisson is working on this

PR #1626, opened 2023-01-23
Copy Bazel produced dSYMs in BwB mode

CognitiveDisson:
Part of #1477.


examples/integration/.bazelrc
```diff
@@ -13,6 +13,12 @@ build:rules_xcodeproj_integration --define=foo=bar
 # Required for sharing cache between Intel and Apple silicon
 build:rules_xcodeproj --@build_bazel_rules_swift//swift:universal_tools
 
+build:rules_xcodeproj --apple_generate_dsym
+build:rules_xcodeproj --output_groups=+dsyms
+
+build --apple_generate_dsym
+build --output_groups=+dsyms
```
CognitiveDisson, line 20: for testing

examples/integration/BUILD
```diff
@@ -47,7 +47,7 @@ exports_files(["README.md"])
     xcodeproj(
         name = "xcodeproj-{}".format(simulator_cpu),
         associated_extra_files = ASSOCIATED_EXTRA_FILES,
-        build_mode = "xcode",
+        build_mode = "bazel",
```
CognitiveDisson, line 50: for testing

examples/integration/xcodeproj.bazelrc
```diff
@@ -3,3 +3,6 @@ import %workspace%/../../xcodeproj.bazelrc
 
 # Disabling archived bundles for fixture coverage
 build:rules_xcodeproj --define=apple.experimental.tree_artifact_outputs=1
+
+build:rules_xcodeproj --apple_generate_dsym
+build:rules_xcodeproj --output_groups=+dsyms
```
CognitiveDisson, line 8: for testing

xcodeproj/internal/bazel_integration_files/copy_outputs.sh
```diff
@@ -19,6 +19,23 @@ else
...
+        rsync \
+        --copy-links \
+        --recursive \
+        --times \
+        --archive \
+        --delete \
+        ${exclude_list:+--exclude-from="$exclude_list"} \
+        --chmod=u+w \
+        --out-format="%n%L" \
+        $dsym_files \
```
CognitiveDisson, line 34: This is a space separated list of files, so no quotes

xcodeproj/internal/top_level_targets.bzl
```diff
@@ -455,7 +459,7 @@ def process_top_level_target(
     if not ctx.var["COMPILATION_MODE"] == "dbg":
         build_settings["DEBUG_INFORMATION_FORMAT"] = ""
     elif not cpp.apple_generate_dsym:
-        build_settings["DEBUG_INFORMATION_FORMAT"] = "dwarf"
+        build_settings["DEBUG_INFORMATION_FORMAT"] = "dwarf-with-dsym"
```
CognitiveDisson, line 462: Not sure if it's worth changing, I'll check it out.

examples/integration/xcodeproj.bazelrc
brentleyjones: In the end, all you will need is `build:rules_xcodeproj --apple_generate_dsym`, only in this file.

xcodeproj/internal/bazel_integration_files/copy_outputs.sh
brentleyjones: But it could have spaces. So you'll need to handle it like `PREVIEW_FRAMEWORK_PATHS` below, where each path is quoted, and use `xargs -n1 sh -c` in order to split it into an array while preserving spaces within the quotes.

xcodeproj/internal/bazel_integration_files/copy_outputs.sh
```diff
@@ -19,6 +19,23 @@ else
       suffix=/Payload
     fi
 
+    if [[ -n ${BAZEL_OUTPUTS_DSYM:-} ]]; then
+      if [[ "$BAZEL_OUTPUTS_DSYM" = *.dSYM ]]; then
```
brentleyjones, line 23: Do we need this check? We have it above to allow for it to be zipped first, in which case we have to unzip it. Is it possible that it's zipped as well?

xcodeproj/internal/output_files.bzl
```diff
@@ -42,6 +42,8 @@ def _create(
     direct_products = []
     indexstore = None
 
+    dsym_files = None
```
brentleyjones, line 45: nit: move this into the group above.

xcodeproj/internal/output_files.bzl
```diff
@@ -112,6 +117,8 @@ def _create(
                 ))
         ],
     )
+    if dsym_files:
+        transitive_products = depset(transitive_products.to_list() + dsym_files.to_list())
```
brentleyjones, line 121: `to_list()` is very bas for performance. I think you just want to add the dsym files to the `depset(` creation above, as part of the array that is being sent into it.

xcodeproj/internal/output_files.bzl
```diff
@@ -214,22 +223,39 @@ def _get_outputs(*, id, product, swift_info):
...
+    if _has_dsym(debug_outputs) and output_group_info and "dsyms" in output_group_info:
+        dsym_files = output_group_info["dsyms"]
+
     return struct(
         id = id,
         is_framework = is_framework,
         product = product.file if product else None,
         product_path = product.path if product else None,
         product_file_path = product.actual_file_path if product else None,
+        dsym_files = dsym_files,
```
brentleyjones, line 237:
```suggestion
        dsym_files = dsym_files,
        id = id,
        is_framework = is_framework,
        product = product.file if product else None,
        product_path = product.path if product else None,
        product_file_path = product.actual_file_path if product else None,
```

xcodeproj/internal/output_files.bzl
```diff
@@ -260,10 +288,13 @@ def _collect_output_files(
         An opaque `struct` that should be used with `output_files.to_dto` or
         `output_files.to_output_groups_fields`.
     """
+
     outputs = _get_outputs(
         id = id,
         product = top_level_product,
         swift_info = swift_info,
+        debug_outputs = debug_outputs,
+        output_group_info = output_group_info,
```
brentleyjones, line 297:
```suggestion
        debug_outputs = debug_outputs,
        id = id,
        output_group_info = output_group_info,
        product = top_level_product,
        swift_info = swift_info,
```

xcodeproj/internal/top_level_targets.bzl
brentleyjones: `not cpp.apple_generate_dsym`, so we need an `else` case here to account for `dwarf-with-dsym`.

xcodeproj/internal/xcode_targets.bzl
```diff
@@ -185,6 +185,9 @@ def _to_xcode_target_outputs(outputs):
         product_path = (
             direct_outputs.product_path if direct_outputs else None
         ),
+        dsym_files =  (
+            direct_outputs.dsym_files if direct_outputs else None
+        ),
```
brentleyjones, line 190: nit: move this up to the top

xcodeproj/internal/xcode_targets.bzl
```diff
@@ -290,6 +293,7 @@ def _merge_xcode_target_outputs(*, src, dest):
         swift_generated_header = src.swift_generated_header,
         product_file = dest.product_file,
         product_path = dest.product_path,
+        dsym_files = dest.dsym_files,
```
brentleyjones, line 296: nit: move this up to the top

xcodeproj/internal/xcode_targets.bzl
```diff
@@ -327,11 +331,22 @@ def _set_bazel_outputs_product(
...
 
-    path = xcode_target.outputs.product_path
-    if not path:
-        return
-
-    build_settings["BAZEL_OUTPUTS_PRODUCT"] = path
+    product_file = xcode_target.outputs.product_file
+    if product_file:
+        build_settings["BAZEL_OUTPUTS_PRODUCT"] = product_file.path
+
```
brentleyjones, line 337: This regressed. I changed the logic to use `xcode_target.outputs.product_path`.

xcodeproj/internal/bazel_integration_files/copy_outputs.sh
CognitiveDisson: No, I will remove it.

xcodeproj/internal/top_level_targets.bzl
brentleyjones: Actually, I didn't do that, because that is the default. Did you find yourself needing this change at all?

CognitiveDisson: According to [this](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html) documentation, the default is `dwarf`.
I don't notice any issues with this value, but with the current state in the master, `DEBUG_INFORMATION_FORMAT` value just an empty string, which just looks confusing if `apple_generate_dsym` is enabled.

So I think it make sense to change it to:
```
elif cpp.apple_generate_dsym:
        build_settings["DEBUG_INFORMATION_FORMAT"] = "dwarf-with-dsym"
```

I quickly checked and it looks like it should be this way.

brentleyjones: In `CreateProject` in the generator let's set `DEBUG_INFORMATION_FORMAT = "dwarf"` to make this explicit then.

xcodeproj/internal/xcode_targets.bzl
CognitiveDisson: Fixed

xcodeproj/internal/output_files.bzl
CognitiveDisson: Should be resolved. There is still `to_list()`, but only for `dsym files`. Guess that's the only way.

xcodeproj/internal/bazel_integration_files/copy_outputs.sh
```diff
@@ -19,6 +19,21 @@ else
       suffix=/Payload
     fi
 
+    if [[ -n ${BAZEL_OUTPUTS_DSYM:-} ]]; then
+      read -a dsym_files <<< "$BAZEL_OUTPUTS_DSYM"
```
brentleyjones, line 23: I don't think this will be good enough. I really think you need to go the `xargs` route like I did for `PREVIEW_FRAMEWORK_PATHS`. This way won't handle spaces correctly. The entries need to be quoted items.

xcodeproj/internal/opts.bzl
```diff
@@ -791,11 +791,11 @@ def _process_compiler_opts(
...
     if len(has_debug_infos) == 1:
-        # We don't set "DEBUG_INFORMATION_FORMAT" for "dwarf"-with-dsym",
-        # as that's Xcode's default
         if not has_debug_infos[0]:
             build_settings["DEBUG_INFORMATION_FORMAT"] = ""
-        elif not cpp_fragment.apple_generate_dsym:
+        elif cpp_fragment.apple_generate_dsym:
+            build_settings["DEBUG_INFORMATION_FORMAT"] = "dwarf-with-dsym"
+        else:
             build_settings["DEBUG_INFORMATION_FORMAT"] = "dwarf"
```
brentleyjones, line 799: We can remove this part now, since it's the default. You'll have to adjust the bzl tests. You can change the top comment to:

```
        # We don't set "DEBUG_INFORMATION_FORMAT" for "dwarf", as we set that at
        # the project level.
```

xcodeproj/internal/output_files.bzl
```diff
@@ -101,7 +106,7 @@ def _create(
     # products. Until then we need them, to allow `Copy Bazel Outputs` to be
     # able to copy the products of transitive dependencies.
     transitive_products = depset(
-        direct_products if direct_products else None,
+        direct_products + dsym_files.to_list() if direct_products else None,
```
brentleyjones, line 109: Don't need to do `to_list()`, you can add it to `transitive` below with `+ [dsym_files]`.

xcodeproj/internal/top_level_targets.bzl
```diff
@@ -449,12 +453,10 @@ def process_top_level_target(
 
     # We don't have access to `CcInfo`/`SwiftInfo` here, so we have to make
     # a best guess at `-g` being used
-    # We don't set "DEBUG_INFORMATION_FORMAT" for "dwarf"-with-dsym",
-    # as that's Xcode's default
```
brentleyjones, line 453:
```
        # We don't set "DEBUG_INFORMATION_FORMAT" for "dwarf", as we set that at
        # the project level.
```

xcodeproj/internal/xcode_targets.bzl
```diff
@@ -324,11 +328,22 @@ def _set_bazel_outputs_product(
...
+        dsym_paths = []
+        for file in dsym_files.to_list():
+            file_path = file.path
+            # dSYM files contain plist and DWARF.
+            if not file_path.endswith('Info.plist'):
+                # ../Product.dSYM/Contents/Resources/DWARF/Product
+                dsym_path =  "/".join(file_path.split('/')[:-4])
+                dsym_paths.append(dsym_path)
+        if dsym_paths:
+            build_settings["BAZEL_OUTPUTS_DSYM"] = "\n".join(dsym_paths)
```
brentleyjones, line 346: Per my other comment, each path should be wrapped in `"`, in case it has spaces in it.

examples/integration/xcodeproj.bazelrc
```diff
@@ -3,3 +3,5 @@ import %workspace%/../../xcodeproj.bazelrc
 
 # Disabling archived bundles for fixture coverage
 build:rules_xcodeproj --define=apple.experimental.tree_artifact_outputs=1
+
+#build:rules_xcodeproj --apple_generate_dsym
```
brentleyjones, line 7:
```suggestion
```

tools/generator/src/Generator/CreateProject.swift
```diff
@@ -59,6 +59,7 @@ $(SYMROOT)/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)
 $(INDEXING_BUILT_PRODUCTS_DIR__$(INDEX_ENABLE_BUILD_ARENA))
 """,
             "CONFIGURATION_BUILD_DIR": "$(BUILD_DIR)/$(BAZEL_PACKAGE_BIN_DIR)",
+            "DEBUG_INFORMATION_FORMAT": "dwarf",
```
brentleyjones, line 62: You'll need to update `CreateProjectTests` as well.

xcodeproj/internal/bazel_integration_files/copy_outputs.sh
```diff
@@ -19,6 +19,21 @@ else
...
+      --copy-links \
+      --recursive \
+      --times \
+      --archive \
+      --delete \
+      ${exclude_list:+--exclude-from="$exclude_list"} \
+      --chmod=u+w \
+      --out-format="%n%L" \
+      ${dsym_files[@]} \
+      "$TARGET_BUILD_DIR"
```
brentleyjones, line 34: Glad you found the other examples.

I don't think we need to create an array for this one though, I think instead you can do this inline:

```suggestion
      while IFS='' read -r file; do dsym_files+=("${file}"); done < <(xargs -n1 <<< "$BAZEL_OUTPUTS_DSYM")
      rsync \
      --copy-links \
      --recursive \
      --times \
      --archive \
      --delete \
      ${exclude_list:+--exclude-from="$exclude_list"} \
      --chmod=u+w \
      --out-format="%n%L" \
      $(xargs -n1 <<< "$BAZEL_OUTPUTS_DSYM") \
      "$TARGET_BUILD_DIR"
```

Which is how the preview frameworks used to be: https://github.com/buildbuddy-io/rules_xcodeproj/blob/b6144935ad30cb933d92f9c5579426f4868f93c2/xcodeproj/internal/bazel_integration_files/copy_outputs.sh#L72

examples/integration/BUILD
brentleyjones: You should be able to revert this and update the fixtures, right? Anything else missing before this is no longer a draft?

CognitiveDisson: Reverted. I updated the fixtures.

>Anything else missing before this is no longer a draft?
`DBGSourcePathRemapping`. 

And maybe some additional tests or is the current one enough?

brentleyjones (approved):
Approving to run CI

examples/integration/BUILD
brentleyjones: I can't think of what additional tests we could do at this point, but if you have ideas, I'm all for them.

CognitiveDisson: I was thinking about some kind of E2E test that would check that dSYM is present, but I don't see how to add this type of test right now. 

It might be better to implement the `DBGSourcePathRemapping` logic in a separate PR, so I'm going to mark this PR as "Ready for Review".

I've fixed some linter errors, so I think you can restart CI.

brentleyjones (approved):
Can you add `--apple_generate_dsym` to `example/integration`'s `.bazelrc` and update the fixture after?

examples/integration/test/fixtures/bwb.xcodeproj/project.pbxproj
```diff
@@ -7325,6 +7333,28 @@
 				"BAZEL_COMPILE_TARGET_ID[sdk=iphoneos*]" = "//iOSApp/Source:iOSApp.library ios-arm64-min15.0-applebin_ios-ios_arm64-dbg-STABLE-4";
 				"BAZEL_COMPILE_TARGET_ID[sdk=iphonesimulator*]" = "$(BAZEL_COMPILE_TARGET_ID)";
 				BAZEL_LABEL = "@//iOSApp/Source:iOSApp";
+				BAZEL_OUTPUTS_DSYM = (
+					"\"bazel-out/applebin_ios-ios_x86_64-dbg-STABLE-7/bin/iOSApp/Source/CoreUtilsObjC/CoreUtilsObjC.framework.dSYM\"",
+					"\"bazel-out/applebin_ios-ios_x86_64-dbg-STABLE-7/bin/Lib/LibFramework.iOS.framework.dSYM\"",
+					"\"bazel-out/applebin_ios-ios_x86_64-dbg-STABLE-7/bin/UI/UIFramework.iOS.framework.dSYM\"",
+					"\"bazel-out/applebin_ios-ios_x86_64-dbg-STABLE-7/bin/WidgetExtension/WidgetExtension.appex.dSYM\"",
+					"\"bazel-out/applebin_ios-ios_x86_64-dbg-STABLE-7/bin/AppClip/AppClip.app.dSYM\"",
+					"\"bazel-out/applebin_watchos-watchos_x86_64-dbg-STABLE-9/bin/Lib/LibFramework.watchOS.framework.dSYM\"",
```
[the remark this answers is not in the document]
CognitiveDisson: Yes, it should use Spotlight to find the, so we should be able to copy it to sub-directory per-target.

I don't think it's a good idea to move it to the profile  pre-action, there may be other reasons for having dSYM's.

Alternatively, we can copy the dSYM for the current product only. How about this approach?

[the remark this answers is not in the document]
brentleyjones: > Alternatively, we can copy the dSYM for the current product only. How about this approach?

We would be missing transitive dSYMs for app extensions that aren't added to the scheme. So we need them to be transitive.

> I don't think it's a good idea to move it to the profile pre-action, there may be other reasons for having dSYM's.

What about the launch pre-action as well? And what other reasons are there for the rules_xcodeproj project to copy over dSYMs? lldb doesn't use them, and you can't submit them to the App Store.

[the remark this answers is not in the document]
CognitiveDisson: > What about the launch pre-action as well?
That will be fine. Or only for the launch target also should be fine. If would prefer only for the launch instead of pre-actions. Which one do you think is better?

> Though, what other reasons are there for the rules_xcodeproj project to copy over dSYMs?
Maybe for crash symbolization and for Crashlytics upload 🤔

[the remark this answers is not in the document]
brentleyjones: > Maybe for crash symbolization and for Crashlytics upload

You shouldn't use rules_xcodeproj built things for any submissions. We reserve the right to make slight changes for better debugging and/or integration: https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/docs/faq.md#does-the-archive-action-work

> Or only for the launch target also should be fine.

Can't be just the Launch pre-action script, as that isn't run when Profiling. I actually think this should be a Profile pre-action script, because I don't think there are any other use cases outside of profiling, and that reduces the amount of copying for people that aren't profiling. You can add a script that does this to this array: https://github.com/buildbuddy-io/rules_xcodeproj/blob/9136dcb1dabe8e706d17a1cb9af33177617810d2/tools/generator/src/Extensions/XCScheme%2BExtensions.swift#L108.

tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -103,10 +107,23 @@ extension XCScheme {
...
+            let scriptTitle: String
+            let scriptText: String
+            if buildMode == .xcode {
+                scriptTitle = "Update .lldbinit"
+                scriptText = XCScheme.ExecutionAction.createLLDBInitScript
+            } else {
+                scriptTitle = "Update .lldbinit and copy dSYMs"
+                scriptText = XCScheme.ExecutionAction.createLLDBInitScript +
+                    "\n" + XCScheme.ExecutionAction.copyDSYMsScript
+            }
```
brentleyjones, line 119: I made the change to only copy the dSYMs in BwB mode. A also merged the actions into one, since they have the same `buildableReference`.

brentleyjones (approved):

brentleyjones:
Thank you @CognitiveDisson!

PR #1663, opened 2023-01-26
Remove `top_level_only` from `scheme_autogeneration_mode` docs

mattrobmattrob:
Update docs for `scheme_autogeneration_mode` to match:

https://github.com/buildbuddy-io/rules_xcodeproj/blob/97a4f4254ff5bece0c88f2c77cc80bdc06b65c37/xcodeproj/internal/xcodeproj_rule.bzl#L1588-L1592

And:

https://github.com/buildbuddy-io/rules_xcodeproj/blob/97a4f4254ff5bece0c88f2c77cc80bdc06b65c37/tools/generator/src/DTO/SchemeAutogenerationMode.swift#L1-L5

### DIFF
--- a/docs/design/custom-xcode-schemes.md
+++ b/docs/design/custom-xcode-schemes.md
@@ -255,9 +255,6 @@ __Values__
 - `none`: No schemes are automatically generated.
 - `all`: A scheme is generated for every buildable target even if custom schemes
   are provided.
-- `top_level_only`: A scheme is generated for every top-level target even if
-  custom schemes are provided. A top-level target in this context is one that is
-  not depended upon by any other target in the Xcode project.
 
 The default value for `scheme_autogeneration_mode` is `auto`.
 


docs/design/custom-xcode-schemes.md
```diff
@@ -255,9 +255,6 @@ __Values__
 - `none`: No schemes are automatically generated.
 - `all`: A scheme is generated for every buildable target even if custom schemes
   are provided.
-- `top_level_only`: A scheme is generated for every top-level target even if
-  custom schemes are provided. A top-level target in this context is one that is
-  not depended upon by any other target in the Xcode project.
```
brentleyjones, line 260: So, this document is a historical design document. It doesn't reflect reality. All actual docs are under `docs/bazel`.

That said, we do plan to add `top_level_only`, we just haven't gotten around to it.

I wonder if we should keep this document around? And if we shouldn't, I should probably take the info in the high level design document and make it something more up-to date at a different location.

brentleyjones: https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/docs/bazel.md#xcodeproj-scheme_autogeneration_mode. We should update the values here to match what you found here. I didn't realize they weren't coming though.

mattrobmattrob: Ah, gotcha. I went down the path of trying it a bit and was surprised it was in the docs and not supported. Guess I wasn't properly understanding the docs.

mattrobmattrob: Ah, finally see your comment now, yes!

brentleyjones:
https://github.com/buildbuddy-io/rules_xcodeproj/pull/1735

PR #1701, opened 2023-01-30
Update Sanitizer related feature flags

chiragramani:
Fixes https://github.com/buildbuddy-io/rules_xcodeproj/issues/1613.

Bazel's `asan` feature by default applies `-O1` but we might also want to catch violations for non-optimised code-paths. 

This PR attempts to remove the usage of `--features=asan`, `--features=tsan`, and `--features=ubsan`, so that we can have parity with Xcode builds (which do not apply `-O1`).


xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
```diff
@@ -102,14 +102,15 @@ fi
 
 # Runtime Sanitizers
 if [[ $apply_sanitizers -eq 1 ]]; then
+  readonly common_sanitizer_flags="--copt -fno-omit-frame-pointer --copt -fno-sanitize-recover=all"
   if [ "${ENABLE_ADDRESS_SANITIZER:-}" == "YES" ]; then
-    build_pre_config_flags+=(--features=asan)
+    build_pre_config_flags+=($common_sanitizer_flags --copt=-fsanitize=address --linkopt=-fsanitize=address --swiftcopt=-sanitize=address --copt -D_FORTIFY_SOURCE=0)
```
brentleyjones, line 107:
```suggestion
    build_pre_config_flags+=($common_sanitizer_flags --copt=-fsanitize=address --linkopt=-fsanitize=address --swiftcopt=-sanitize=address --copt -Wno-macro-redefined --copt -D_FORTIFY_SOURCE=0)
```

brentleyjones (approved):
So I'm a little conflicted about this change.

Ideally, Bazel merges this change: https://github.com/bazelbuild/bazel/pull/17355

But even if they do, it seems they are reluctant to make this change, which means we are applying the sanitizer changes to exec tools as well: https://github.com/bazelbuild/bazel/pull/16626

Which means that this change is probably better for our use case?

PR #1739, opened 2023-02-02
Add external example links

brentleyjones:


examples/README.md
```diff
@@ -10,3 +10,12 @@ This directory holds several examples using rules_xcodeproj. To open an example,
 
 * **sanitizers**
     <br> Contains targets to test Sanitizers in BwB mode. The main purpose is to make sure sanitizers in BwB mode work and give UI feedback the same way sanitizers do in BwX builds.
+
+## External
+
+Here are some examples outside of this repository that show how to use
+rules_xcodeproj in various ways:
```
jpsim, line 17: Might want to say something about this not necessarily being up to date since they're permalinks, or to check the latest versions in those project's `main` branch.

README.md
```diff
@@ -55,7 +55,7 @@ the ruleset.
 ## Projects using rules_xcodeproj
 
 - Cash App
-- [Envoy Mobile](https://github.com/envoyproxy/envoy-mobile)
+- [Envoy Mobile](https://github.com/envoyproxy/envoy/tree/main/mobile)
```
jpsim, line 58: Now that the source config is linked directly in the external examples, maybe this should link to the project's online home rather than the source directory in GitHub?

```suggestion
- [Envoy Mobile](https://envoymobile.io)
```

jpsim: Yeah that's fine

jpsim: Then you should probably also remove the SwiftLint link.

However, then this list becomes a less useful reference because the examples links is in another doc.

Maybe you combine the lists here?

I see presentations, blog posts, tweets and examples as being in the same category of reference materials.

PR #1747, opened 2023-02-03
Fix `cp -c` issue with other volumes

### LINKED ISSUE #1746: Bug: Copy with clonefile fails when output_base and workspace are on different volumes
## Describe the bug
When setting the bazel `output_base` to a different volume than the one where the workspace to build resides, a couple of operations using `cp -c` fail with `clonefile failed: Cross-device link`. This happens for example when using a RAM disk as output_base, which is a different volume.

Removing the `-c` flag from the following locations fixes the issue:
https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/xcodeproj/internal/bazel_integration_files/BUILD#L100
https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/xcodeproj/internal/filelists.bzl#L19
https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/xcodeproj/internal/installer.template.sh#L141
https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/xcodeproj/internal/installer.template.sh#L147
https://github.com/buildbuddy-io/rules_xcodeproj/blob/main/xcodeproj/internal/product.bzl#L40

To not lose the advantages of using clonefile, a proper fix would be to check if source and destination reside on the same volume to decide whether or not to use `-c`.  Something along the lines of

```
if [[ $(stat -f '%d' "$from") == $(stat -f '%d' "$to") ]]; then
```

## Reproduction steps
1. Set `output_base` to a different volume than the current workspace
2. Generate project (BwB mode)
3. Build and run in Xcode

## Expected behavior
Copying across volumes should work without errors

## Versions
- Bazel: 6.0.0
- rules_xcodeproj: 1.0.0

brentleyjones:
Fixes #1746.

### DIFF
--- a/xcodeproj/internal/bazel_integration_files/BUILD
+++ b/xcodeproj/internal/bazel_integration_files/BUILD
@@ -97,7 +97,14 @@ genrule(
     srcs = ["//tools/swiftc_stub:universal_swiftc_stub"],
     outs = ["swiftc"],
     # Make the swiftc_stub have the right name
-    cmd = "cp -c $< $@",
+    cmd = """\
+readonly output="$@"
+if [[ $$(stat -f '%d' "$<") == $$(stat -f '%d' "$${output%/*}") ]]; then
+  cp -c "$<" "$@"
+else
+  cp "$<" "$@"
+fi
+""",
     message = "Renaming swiftc stub",
     tags = [
         "manual",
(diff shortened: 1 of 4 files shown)


brentleyjones:
@arrizer can you confirm this fixes your issue?

arrizer:
@brentleyjones just tested with the branch and I did not encounter any errors. Thanks for the quick fix!

PR #1749, opened 2023-02-06
Adjust BCR Publish releaser

brentleyjones:
By having it be me, I can amend release commits easier. Later we should be able to use an organization fork instead, so any maintainers can amend the release commits.

### DIFF
--- a/.bcr/config.yml
+++ b/.bcr/config.yml
@@ -1,3 +1,3 @@
 fixedReleaser:
-  login: BalestraPatrick
-  email: me@patrickbalestra.com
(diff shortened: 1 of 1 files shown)


.bcr/config.yml
```diff
@@ -1,3 +1,3 @@
 fixedReleaser:
-  login: BalestraPatrick
-  email: me@patrickbalestra.com
+  login: brentleyjones.com
```
BalestraPatrick, line 2: I think this needs to be your username, so without `.com`

PR #1758, opened 2023-02-11
Fix broken BwX link in README

cpsauer:
Happened to notice a broken link while browsing the README, so I figured I'd give it a quick fix to try to leave things even better than I found them.

Thanks for all the wonderful things you're doing here!
Chris


brentleyjones:
Thank you! I merged a change that covered a couple other links as well.

cpsauer:
Even better! Wish only I'd caught the others for you. Thanks, @brentleyjones, for all you do.

PR #1762, opened 2023-02-13
Set full range for `SUPPORTED_PLATFORMS` in normal builds

brentleyjones:
In 0c65cdde21007c53595a60138b623749b15644e6 we filtered `SUPPORTED_PLATFORMS` to prevent build errors in Index Build. This causes problems with `ARCHS` not being set correctly if a project contains macOS and {i,tv,watch}OS targets. With this change, we partially revert the filtering by having it only apply to Index Build.


brentleyjones:
Requested reviews are just to verify that a normal build continues to work for y'all. @BalestraPatrick can you be sure to test projects both with and without watchOS targets included?

BalestraPatrick (approved):
Normal builds for iOS with and without watchOS app work for us 👍

PR #1764, opened 2023-02-13
Use `swift.file_prefix_map` by default

brentleyjones:
The primary benefit for us is that index store data is reproducible, reducing extra uploads and downloads for the “same” index data.

Requires Xcode 14 and rules_swift 1.5.2.


brentleyjones:
Requires a release of rules_swift with https://github.com/bazelbuild/rules_swift/pull/1001 (which I assume would be 1.5.2 or 1.6.0).

I don't think we want to bump our minimum version just yet, but I wanted to submit this PR now so I don't lose track of it.

brentleyjones:
https://github.com/bazelbuild/rules_swift/pull/1014

PR #1767, opened 2023-02-20
Patch dSYM path

CognitiveDisson:
To improve some lldb handling of relative source paths, we create a plist file inside the dSYM file with `DBGSourcePathRemapping` which will remap relative paths to absolute paths (see [documentation](https://lldb.llvm.org/use/symbols.html#id2)).

This logic has been added to the `copy_dsyms.sh` script, which will be run as Launch and Profiling pre-actions.

### DIFF
--- a/xcodeproj/internal/bazel_integration_files/copy_dsyms.sh
+++ b/xcodeproj/internal/bazel_integration_files/copy_dsyms.sh
@@ -2,6 +2,45 @@
 
 set -euo pipefail
 
+function patch_dsym() {
+  readonly dsym_name="$1"
+  readonly binary_name="${dsym_name%%.*}"
+  readonly binary_path="${dsym_name}/Contents/Resources/DWARF/${binary_name}"
+
+  if [[ ! -f "$binary_path" ]]; then
+    echo "dSYM DWARF ${binary_path} does not exist." \
+    "Skip dSYM patch."
+    return 1
+  fi
+
+  dwarf_uuid=$(dwarfdump --uuid "${binary_path}" | cut -d ' ' -f 2)
+  readonly dwarf_uuid
+  if [[ -z "${dwarf_uuid// /}" ]]; then
+    echo "Failed to get dSYM uuid." \
+    "Skip dSYM patch."
+    return 1
+  fi
+
+  cat > "${dsym_name}/Contents/Resources/${dwarf_uuid}.plist" <<EOF
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+  <dict>
+    <key>DBGVersion</key>
+    <string>3</string>
+    <key>DBGSourcePathRemapping</key>
+    <dict>
+      <key>./bazel-out/</key>
+      <string>$BAZEL_OUT/</string>
+      <key>./external/</key>
+      <string>$BAZEL_EXTERNAL/</string>
+      <key>./</key>
+      <string>$SRCROOT/</string>
+    </dict>
+  </dict>
+</plist>
+EOF
+}
+
 if [[ -n "${BAZEL_OUTPUTS_DSYM:-}" ]]; then
   cd "${BAZEL_OUT%/*}"
 
@@ -17,4 +56,11 @@ if [[ -n "${BAZEL_OUTPUTS_DSYM:-}" ]]; then
     --out-format="%n%L" \
     $(xargs -n1 <<< "$BAZEL_OUTPUTS_DSYM") \
     "$TARGET_BUILD_DIR"
+
+  cd "${TARGET_BUILD_DIR}"
+
+  export -f patch_dsym
+  # shellcheck disable=SC2016
+  xargs -n1 sh -c 'patch_dsym $(basename "$1")' _ \
+    <<< "$BAZEL_OUTPUTS_DSYM"
 fi


tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -43,7 +43,7 @@ extension XCScheme {
                 buildActionInfo: buildActionInfo,
                 otherPreActions: otherPreActions
             )
-            
+
```
CognitiveDisson, line 46: SwiftFormat 🤔

xcodeproj/internal/bazel_integration_files/copy_dsyms.sh
```diff
@@ -2,6 +2,47 @@
...
+  cat > "${plist_path}" <<EOF
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+  <dict>
+    <key>DBGVersion</key>
+    <string>3</string>
+    <key>DBGSourcePathRemapping</key>
+    <dict>
+      <key>$execution_root</key>
+      <string>$workspace_root</string>
```
brentleyjones, line 39: The remapping should be the same as in `.lldbinit`, changing relative paths to absolute. In particular, it should be these three:

- `./bazel-out/` -> `$BAZEL_OUT`
- `./external/` -> `$BAZEL_EXTERNAL`
- `./` -> `$SRCROOT`

Might also need to do the same with `RESOLVED_REPOSITORIES` (and if so, might need to do the ordering trick as well).

tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -83,15 +83,25 @@ extension XCScheme {
...
+            let scriptTitle: String
+            let scriptText: String
+            if buildMode == .xcode {
+                scriptTitle = "Update .lldbinit"
+                scriptText = XCScheme.ExecutionAction.createLLDBInitScript
+            } else {
+                scriptTitle = "Update .lldbinit and copy dSYMs"
+                scriptText = XCScheme.ExecutionAction.createLLDBInitScript +
+                    "\n" + XCScheme.ExecutionAction.copyDSYMsScript
+            }
```
brentleyjones, line 95: Why are you adding anything to the Launch action? We are already calling this in the Profile action. Any change to that logic should be its own PR (but I thought we agreed that it shouldn't be the launch action).

tools/generator/src/Extensions/XCScheme+Extensions.swift
brentleyjones: Can you have this be its own PR?

tools/generator/src/Extensions/XCScheme+Extensions.swift
CognitiveDisson: I wanted to check if this solves [this problem](https://github.com/buildbuddy-io/rules_xcodeproj/issues/1336#issuecomment-1428749738).

xcodeproj/internal/bazel_integration_files/copy_dsyms.sh
CognitiveDisson: Note:
I think there is still something wrong with the path, the Instruments are complaining about finding the source.
Will investigate this.

tools/generator/src/Extensions/XCScheme+Extensions.swift
CognitiveDisson: That seems works

xcodeproj/internal/bazel_integration_files/copy_dsyms.sh
```diff
@@ -2,6 +2,51 @@
 
 set -euo pipefail
 
+# This is for testing, removed before merging
+set -x
+exec > >(tee -i ~/logfile.txt)
+exec 2>&1
+#############################################
+
```
brentleyjones, line 10: Remove

xcodeproj/internal/bazel_integration_files/copy_dsyms.sh
```diff
@@ -2,6 +2,51 @@
...
+function patch_dsym() {
+  readonly dsym_name="$1"
+  readonly binary_name="${dsym_name%%.*}"
+  readonly binary_path="${dsym_name}/Contents/Resources/DWARF/${binary_name}"
+  if [[ ! -f "$binary_path" ]]; then
+    echo "dSYM DWARF ${binary_path} does not exist." \
+    "Skip dSYM patch."
+    return 1
+  fi
+  dwarf_uuid=$(dwarfdump --uuid "${binary_path}" | cut -d ' ' -f 2)
```
brentleyjones, line 20:
```suggestion

  dwarf_uuid=$(dwarfdump --uuid "${binary_path}" | cut -d ' ' -f 2)
```

xcodeproj/internal/bazel_integration_files/copy_dsyms.sh
```diff
@@ -2,6 +2,51 @@
...
+  readonly dwarf_uuid
+  if [[ -z "${dwarf_uuid// /}" ]]; then
+    echo "Failed to get dSYM uuid." \
+    "Skip dSYM patch."
+    return 1
+  fi
+  readonly plist_path="${dsym_name}/Contents/Resources/${dwarf_uuid}.plist"
+  readonly execution_root="$PROJECT_DIR"
+  readonly workspace_root="$BAZEL_WORKSPACE_ROOT"
+  cat > "${plist_path}" <<EOF
```
brentleyjones, line 30:
```suggestion

  cat > "${dsym_name}/Contents/Resources/${dwarf_uuid}.plist" <<EOF
```

brentleyjones (approved):
After the comments and rebasing, we can merge this.

brentleyjones:
I didn't see this fix https://github.com/buildbuddy-io/rules_xcodeproj/issues/1284 either ☹️. It's fine to merge though, since it's still correct to do.

PR #1773, opened 2023-02-21
Merge dest targets with single Xcode src target

mattrobmattrob:
# Summary
This is a slight change to pre-filter the `src_ids` by those that actually produce Xcode targets and then do the merge for those destinations with a single library dependency.

In my case, this merges "test helper" libraries that are depended on by a single unit/UI test target but not another static library.

This is work towards https://github.com/buildbuddy-io/rules_xcodeproj/issues/1724 but doesn’t solve it yet.

### DIFF
--- a/xcodeproj/internal/xcodeproj_rule.bzl
+++ b/xcodeproj/internal/xcodeproj_rule.bzl
@@ -498,13 +498,11 @@ actual targets: {}
     target_merges = {}
     target_merge_srcs_by_label = {}
     for dest, src_targets in target_merge_dests.items():
-        focused_src_targets = [src_target for src_target in src_targets if focused_targets.get(src_target.id)]
-
-        if len(focused_src_targets) > 1:
+        if len(src_targets) > 1:
             # We can only merge targets with a single Xcode-target library dependency
             continue
 
-        src_target = focused_src_targets[0]
+        src_target = src_targets[0]
         target_merges.setdefault(src_target.id, []).append(dest)
         target_merge_srcs_by_label.setdefault(src_target.target.label, []).append(src_target.id)
 
--- a/xcodeproj/internal/xcodeproj_rule.bzl
+++ b/xcodeproj/internal/xcodeproj_rule.bzl
@@ -403,15 +403,14 @@ targets.
         dest_target = unprocessed_targets[merge.dest]
         dest_label = bazel_labels.normalize(dest_target.label)
         if (sets.contains(unfocused_labels, src_label) or
-            sets.contains(unfocused_labels, dest_label)):
+            sets.contains(unfocused_labels, dest_label) or
+            not src_target):
             continue
-        target_merge_dests.setdefault(merge.dest, []).append(merge.src.id)
+        target_merge_dests.setdefault(merge.dest, []).append(struct(id = merge.src.id, target = src_target))
 
-    for dest, src_ids in target_merge_dests.items():
-        src_targets = [struct(src = src, target = unprocessed_targets[src]) for src in src_ids if unprocessed_targets.get(src, None)]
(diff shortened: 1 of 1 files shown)


xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -408,7 +408,9 @@ targets.
         target_merge_dests.setdefault(merge.dest, []).append(merge.src.id)
 
     for dest, src_ids in target_merge_dests.items():
-        if len(src_ids) > 1:
+        src_targets = [struct(src = src, target = unprocessed_targets[src]) for src in src_ids if unprocessed_targets.get(src, None)]
+
+        if len(src_targets) > 1:
             # We can only merge targets with a single library dependency
```
brentleyjones, line 414:
```suggestion
            # We can only merge targets with a single Xcode-target library dependency
```

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -408,7 +408,9 @@ targets.
         target_merge_dests.setdefault(merge.dest, []).append(merge.src.id)
 
     for dest, src_ids in target_merge_dests.items():
-        if len(src_ids) > 1:
+        src_targets = [struct(src = src, target = unprocessed_targets[src]) for src in src_ids if unprocessed_targets.get(src, None)]
```
brentleyjones, line 411: You can do this filtering and `struct` creation up in the `for` loop above.

brentleyjones (changes requested):

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -480,30 +481,32 @@ actual targets: {}
...
 
     target_merges = {}
     target_merge_srcs_by_label = {}
-    for dest, src_ids in target_merge_dests.items():
-        if len(src_ids) > 1:
+    for dest, src_targets in target_merge_dests.items():
+        focused_src_targets = [src_target for src_target in src_targets if focused_targets.get(src_target.id)]
+
+        if len(focused_src_targets) > 1:
             # We can only merge targets with a single library dependency
```
mattrobmattrob, line 504:
```suggestion
            # We can only merge targets with a single Xcode-target library dependency
```

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -480,30 +481,32 @@ actual targets: {}
...
-        target_merge_dests[dest] = new_srcs_ids
+        target_merge_dests[dest] = new_src_targets
 
     target_merges = {}
     target_merge_srcs_by_label = {}
-    for dest, src_ids in target_merge_dests.items():
-        if len(src_ids) > 1:
-            # We can only merge targets with a single library dependency
+    for dest, src_targets in target_merge_dests.items():
+        focused_src_targets = [src_target for src_target in src_targets if focused_targets.get(src_target.id)]
```
brentleyjones, line 501: Is this check needed? Shouldn't the above `for` loop already filter out ones that aren't focused?

mattrobmattrob: Totally right.

brentleyjones:
Would you be able to add something to `examples/integration` to test this new logic? Something that would fail before but now works (and is captured in the fixtures)?

mattrobmattrob:
🤦 I'm now realizing after reproducing again that https://github.com/buildbuddy-io/rules_xcodeproj/pull/1769 actually fixed the underlying problem that I was seemingly working around with this. Will close and re-open with more target merge changes in another PR.

Thanks for the reviews anyway, @brentleyjones!

PR #1784, opened 2023-02-24
Use the `--reuse_sandbox_directories` flag

brentleyjones:
This is only a speed up if sandboxing is used, which is the default for us currently (outside of worker actions, like `SwiftCompile`).


brentleyjones:
https://github.com/bazelbuild/bazel/issues/10702#issuecomment-1443399680

PR #1807, opened 2023-03-03
Have the runner create a unique package for the generator

brentleyjones:
This will be leveraged by future changes to allow for dynamic creation of `xcodeproj` rules, in order to support user-defined Xcode configurations.


xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -1776,15 +1765,18 @@ The text of a script that will be run before the build. For example:
             doc = "A dictionary of global project options.",
             mandatory = True,
         ),
+        "runner_build_file": attr.string(
+            doc = "The path to the runner's BUILD file.",
+        ),
         "runner_label": attr.string(doc = "The label of the runner target."),
         "scheme_autogeneration_mode": attr.string(
             doc = "Specifies how Xcode schemes are automatically generated.",
```
brentleyjones, line 1779: These docs are never seen by a user. I'm not sure how much value they provide to a developer. I'm wondering if we should keep them? And if we do, should they be duplicated on the runner (and in some case triplicated if also exposed to the user in the macro) ?

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -1838,7 +1830,7 @@ dependencies of the targets specified in the `top_level_targets` attribute with
 a matching label will be excluded from the generated project. This overrides any
 targets specified in the `focused_targets` attribute.
 """,
-            default = [],
+            mandatory = True,
```
brentleyjones, line 1842: By making these mandatory, we can catch errors of the runner forgetting to set the values.

xcodeproj/internal/xcodeproj_runner.bzl
```diff
@@ -103,6 +110,108 @@ def _write_extra_flags_bazelrc(name, actions, attr, config):
...
+def _write_generator_defz_bzl(*, actions, attr, is_fixture, name, template):
+    output = actions.declare_file("{}.generator.defs.bzl".format(name))
+
+    build_mode = attr.build_mode
+
+    loads = [
+        """\
+# buildifier: disable=bzl-visibility
+load(
+    # FIXME: Make this path work with any bzlmod repo name
```
brentleyjones, line 129: FIXME

xcodeproj/internal/xcodeproj_runner.bzl
```diff
@@ -111,17 +220,28 @@ def _write_runner(
...
     is_bazel_6 = hasattr(apple_common, "link_multi_arch_static_library")
 
+    generator_package_directory = (
+        ".rules_xcodeproj/" +
+        paths.join(
+            paths.dirname(generator_build_file.short_path),
+            name,
+        ).replace("/", "_")
+    )
+    generator_label = "//{}:{}".format(generator_package_directory, name)
```
brentleyjones, line 243: By having a unique package per target, this supports concurrent project generations (assuming each has their own bazel lock).

.gitignore
```diff
@@ -1,5 +1,7 @@
 /bazel-*
 
+**/.rules_xcodeproj
```
brentleyjones, line 3: I'll need to add something to the FAQ or elsewhere talking about what this directory is, so people can ignore it just like this.

brentleyjones:
Friendly ping @BalestraPatrick @erikkerber @jpsim @mattrobmattrob 🙏.

.gitignore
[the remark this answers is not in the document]
BalestraPatrick: I think given people might already have various patterns with `.xcodeproj` ignored (like it is the case in various default gitignores for Apple developers), I'd say this is a bit more descriptive and might help people connect this name with the project.

xcodeproj/internal/xcodeproj_rule.bzl
BalestraPatrick: Would this prevent having the docs generated in the markdown files? If they're not supposed to be exposed to a consumer, then we could get rid of them.

brentleyjones: These ones aren't generated in the markdown currently. I only have them for developers.

.gitignore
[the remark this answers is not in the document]
brentleyjones: I might be able to get away with using an external repository for this, but if not, I agree.

mattrobmattrob (commented):
Probably worth including that `.rules_xcodeproj/` should be lumped into a `package_group` /the `visibility` param to make this all work:

```
ERROR: /Users/matt.robinson/Developer/REPO/.rules_xcodeproj/xcode-projectgen_xcodegen_XcodeProject/BUILD:6:10: in xcodeproj rule //.rules_xcodeproj/xcode-projectgen_xcodegen_XcodeProject:XcodeProject: target '<REDACTED>' is not visible from target '//.rules_xcodeproj/xcode-projectgen_xcodegen_XcodeProject:XcodeProject'. Check the visibility declaration of the former target if you think the dependency is legitimate
```

We already do this for our auto-discovery functionality's generated `BUILD.bazel` file as well so it was easy to add but fairly substantial change compared to the current implementation WRT visibility.

brentleyjones:
Good callout. It's the one change I had to make to targets to make this work, and it's not entirely clear what the exact value should be. I'll add something to the FAQ.

jpsim (approved):
+1 did some light testing with Envoy Mobile

Issue #1794, opened 2023-02-28
Cleanup: `custom_xcode_schemes.json` files should be as minimal as possible


brentleyjones:
Similar to #237, we should not send over defaults, and for keys we do send over, they should be a single character.

PR #1815, opened 2023-03-08
Add ability to set Xcode configurations

### LINKED ISSUE #1059: Feature Request: Support multiple Xcode configurations
Currently there is only the Debug configuration in generated projects. We should support defining multiple configurations if possible.

(If possible, because currently `--config` can only be supported if the output paths don't change, since we can't transition on `--config`. One possible way around this is to generate a different sub-`spec.json` for each configuration, each from a separate `bazel build` call. This might/probably will thrash the analysis cache and will be super slow. I'm trying to think of other ways to handle this.)

brentleyjones:
Resolves #1059.


xcodeproj/internal/xcodeproj_macro.bzl
```diff
@@ -242,6 +250,15 @@ def xcodeproj(
             hash of all transitive dependencies of the targets specified in the
             `top_level_targets` argument with the `"simulator"`
             `target_environment`, even if they aren't watchOS targets.
+        xcode_configurations: Optional. A `dict` mapping Xcode configuration
+            names to transition settings dictionaries. For example,
+            `{"Dev": {"//command_line_option:compilation_mode": "dbg"}, "AppStore": {"//command_line_option:compilation_mode": "opt"}}`,
```
jpsim, line 255: This example is useful, but more examples for common cases would also be good. You could say that's out of scope here and you do link out to the bazel docs which is great, but maybe being able to answer "how can I specify a bazel config (aka a `--config $foo`) in an Xcode configuration" would be really useful.

Maybe I'm misunderstanding something and a transition can't apply an arbitrary bazel config in that way.

xcodeproj/internal/xcodeproj_macro.bzl
```diff
@@ -266,6 +283,20 @@ in your `.bazelrc` or `xcodeproj.bazelrc` file.""")
         project_name = name
     if not project_options:
         project_options = _default_project_options()
+    if not xcode_configurations:
+        xcode_configurations = {"Debug": {}}
```
jpsim, line 287: Out of the box, Xcode projects and SwiftPM packages have the ability to be built in "Debug" mode and "Release" mode. Why not mirror that here?

jpsim (approved):
Woohoo! Did some light testing with Envoy Mobile and seems to work well, at least for the simple case of Debug / Release configurations.

tools/generator/xcodeproj_targets.bzl
```diff
@@ -12,6 +12,15 @@ _TOOL_TARGET = "//tools/swiftc_stub:swiftc"
 
 TOP_LEVEL_TARGETS = [_APP_TARGET, _TEST_TARGET, _TOOL_TARGET]
 
+XCODE_CONFIGURATIONS = {
+    "Debug": {
+        "//command_line_option:compilation_mode": "dbg",
+    },
+    "Profile": {
```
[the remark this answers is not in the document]
brentleyjones: Mainly to show off that you can use more than just the "default" names

xcodeproj/internal/xcodeproj_macro.bzl
brentleyjones: Correct, you can't use `--config`. The main reason I linked out, is because the rules around what is supported in a transition is out of our hands, and it probably will change with time.

xcodeproj/internal/xcodeproj_macro.bzl
brentleyjones: Each additional configuration is a multiplier on the number of targets analyzed, serialized into the spec.json, and processed in the generator. Besides for being slower with each configuration added, for very large projects we might even run into memory limits.

Since most people survived without the release config, and because you shouldn't really be "releasing" from this generated project, I felt it was ideal not to change the default at this time.

tools/generator/src/DTO/XcodeScheme.swift
```diff
@@ -237,7 +237,7 @@ No `BazelLabel` values were provided to `XcodeScheme.TestAction`.
 
 extension XcodeScheme.TestAction: Decodable {
     enum CodingKeys: String, CodingKey {
-        case buildConfigurationName
+        case buildConfigurationName = "buildConfiguration"
```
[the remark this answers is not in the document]
brentleyjones: `buildConfigurationName` is the name that `XcodeProj` uses, but in Starlark I felt it was better as `build_configuration`. With #1794 this would just be `"b"`.

xcodeproj/internal/xcodeproj_macro.bzl
brentleyjones: I guess the large projects can choose to remove the extra configuration. I just think that might be hard to debug if it's an issue, and also puts rules_xcodeproj at an unfair disadvantage when being compared with other generators that don't do this by default.

We can always change this default later, though it's harder to remove the extra config.

I just got some benchmarks for one large project, where total generation (analysis + generator) went from 43s to 80s (as expected). Maybe after we are able to spend time optimizing that more we can change the default?

jpsim: I don't feel super strongly, but do lean more on the side of having Debug/Release options enabled by default. Larger projects likely have more time to devote to tweaking and optimizing things, whereas the default should be what is a good starting point without much customization.

As far as other generators go, it's not an apples-to-apples comparison, but XcodeGen and Xcode projects auto-generated from SwiftPM packages both would create Debug/Release configurations by default and I'd lean more on feature parity than performance parity for defaults.

It's true you don't "release" from this generated project, but Swift and Objective-C code in particular often varies in behavior and certainly performance when building with or without optimizations.

brentleyjones: Conversation being continued in slack (sorry everyone that doesn't have access): https://bazelbuild.slack.com/archives/C0429BN0F8V/p1678378765794209?thread_ts=1677165331.047499&cid=C0429BN0F8V. I'll post back here with whatever is concluded there though.

brentleyjones: Deciding not to change the default for these reasons:
- The performance increase currently is a concern
- Opting out of this extra configuration is clunky
- Things with Bazel usually need to be explicit, and having this apply some automatic assumptions can actually be confusing
- By not having multiple configurations it allows for easier progressive disclosure, leading to new Bazel users to learn about configs and flags
- While it is an advanced topic, cache warming with the command-line API would by default build twice as many things (we should add a topic to `usage.md` regardless about this so people understand how to do it and what impacts its performance)

jpsim: I think that's fair, thanks for considering it and driving a discussion around it.

erikkerber:
Does this change not blend well with conditional dependencies? I'm switching on a custom flag of ours to test:

```python
xcode_configurations = {
    "Retail": {
        "@slack-objc//:product": "retail",
    },
    "Intune": {
        "@slack-objc//:product": "intune",
    },
}
```

And after a few warnings it flips a precondition error:

```
WARNING: Was unable to consolidate target groupings "[@//App:Slack applebin_ios-ios_arm64-dbg-ST-09660800eb41, @//App:Slack applebin_ios-ios_sim_arm64-dbg-ST-efefcef94047], [@//App:Slack applebin_ios-ios_arm64-dbg-ST-7457851afbd5, @//App:Slack applebin_ios-ios_sim_arm64-dbg-ST-b4da387d22bf]" since they have conditional dependencies (e.g. `deps`, `test_host`, `watch_application`, etc.)
WARNING: Was unable ...
...
`xcodeproj.default_xcode_configuration` "Retail" not one of the configurations of "Key(targetIDs: Set([@//App:SlackEndToEndTests.__internal__.__test_bundle applebin_ios-ios_sim_arm64-dbg-ST-b4da387d22bf]))".
Please file a bug report at https://github.com/buildbuddy-io/rules_xcodeproj/issues/new?template=bug.md
```

If I use the example `opt`/`dbg` flags, it generates just fine.

brentleyjones:
> Does this change not blend well with conditional dependencies?

Ahh, yeah, that would be an issue. It can't consolidate the two configurations into a single target because their deps are different, and there is no way to specify in Xcode "for this configuration use these deps".

With all that said, I believe a separate change should be able to only include deps that are actually set in Xcode in the consideration. By this I mean, in BwB mode, we set very few target dependencies, so if that pruning would change the calculation, it could resolve this for you.

That said, what sort of conditional dependencies do you have here?

brentleyjones:
> after a few warnings it flips a precondition error

I'll have to adjust that to fail in a better way.

One option, technically, is to have the two configurations show up as separate Xcode targets. I think we can even set just the one config, but I'm not sure how Xcode handles that (and if it can't do that, we can just mirror the config for the missing ones). I'm not sure I want to go that route though.

erikkerber:
> That said, what sort of conditional dependencies do you have here?

Our flags are going to conditionally include:

- Some libraries and SDKs depending on what product we are building (our "Intune" is Slack + Intune stuff)
- Debugging libraries for Debug

brentleyjones:
How would you represent that in a normal Xcode project?

erikkerber:
I suppose that might be the hard stop. Xcode doesn't have conditional dependencies, so you'd need a copy of every target for every set of dependencies it might have.

I mean it would be possible, but disgusting.

brentleyjones:
Looks like I accidentally merged this in #1828. I'll address the conditional dependency thing in a followup.

brentleyjones:
I'll revert that actually.

brentleyjones:
I also ran into an issue when two Xcode configurations were basically the same. The target consolidation code incorrectly failed there. I'll need to fix that as well.

brentleyjones:
(I haven't merged this yet because I need to address the above stuff, but I'm busy doing some optimizations to reduce the time it takes to generate a project, since this change makes that even more important.

brentleyjones:
I believe I've addressed all the weird issues for split Xcode configuration targets (where a Bazel target with different Xcode configurations can't be consolidated because of conditional deps). Going to merge now 🤞.

PR #1832, opened 2023-03-09
Replace `__BAZEL` placeholders in compiler flags in the generator

brentleyjones:
This is an optimization since we can do a single replacement pass for each build setting. Swift is faster than Starlark here as well. Finally, in the future when we can parallelize the work we do for each target, this can be done in parallel instead of all serially by Starlark.


tools/generator/src/Generator/AddBazelDependenciesTarget.swift
```diff
@@ -207,8 +207,22 @@ $(INDEXING_SUPPORTED_PLATFORMS__$(INDEX_ENABLE_BUILD_ARENA))
...
+
+  # Replace build settings with their values
+  s/
+    \$             # Match a dollar sign
+    (\()?          # Optionally match an opening parenthesis and capture it
+    ([a-zA-Z_]\w*) # Match a variable name and capture it
+    (?(1)\))       # If an opening parenthesis was captured, match a closing parenthesis
+  /$ENV{$2}/gx;    # Replace the entire matched string with the value of the corresponding environment variable
+
+' "$SCRIPT_INPUT_FILE_0" > "$SCRIPT_OUTPUT_FILE_0"
```
brentleyjones, line 225: Thank you ChatGPT (for the "explain this command").

PR #1846, opened 2023-03-10
Fix issues in README BUILD example

luispadron:
Fixes some load and label issues in the example BUILD file. This still obviously requires the user to add some sources, etc to actually build. But should be clear of other issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ load(
 )
 load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")
 load(
-    "@rules_xcodeproj//xcodeproj:defs.bzl",
+    "@com_github_buildbuddy_io_rules_xcodeproj//xcodeproj:defs.bzl",
     "top_level_target",
     "xcodeproj",
 )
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,7 +114,7 @@ load(
 )
 load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")
 load(
-    "@rules_xcodeproj//xcodeproj:defs.bzl",
+    "@com_github_buildbuddy_io_rules_xcodeproj//xcodeproj:defs.bzl",
```
brentleyjones, line 117: This has changed. I'll mention something in the next release notes, but `@rules_xcodeproj` is now the recommended way.

brentleyjones: (It's odd how the README.md and the most-recent release are mismatched, but I hope to release a new release by Monday EOD.)

README.md
```diff
@@ -147,7 +147,7 @@ ios_unit_test(
     name = "Tests",
     bundle_id = "com.example.tests",
     minimum_os_version = "15.0",
-    test_host = "//App",
+    test_host = ":App",
```
brentleyjones, line 150: This is valid, and we should also add the new visibility changes to the `App` target as well.

README.md
luispadron: I imagine most folks go and checkout the great examples anyway

PR #1857, opened 2023-03-15
Handle very long linker input args

yongjincho92:
Resolves https://github.com/buildbuddy-io/rules_xcodeproj/issues/1855

PR #1862, opened 2023-03-15
Use sub-params files for linker params processing

brentleyjones:
Fixes #855.
Closes #1857.

It seems the `Args` object that the linker creates might not always result in a params file being created. So now we force that to happen, and account for possibly multiple `Args` files being created, and for them to mention params files themselves.

### DIFF
--- a/tools/params_processors/link_params_processor.py
+++ b/tools/params_processors/link_params_processor.py
@@ -13,7 +13,7 @@ def _main(
         self_linked_path: str,
         swift_triple: str,
         output_path: str,
-        args: List[str]
+        args_files: List[str]
     ) -> None:
     with open(xcode_generated_paths_path, encoding = "utf-8") as fp:
         xcode_generated_paths = json.load(fp)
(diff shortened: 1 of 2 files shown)


tools/params_processors/link_params_processor.py
```diff
@@ -36,10 +36,22 @@ def _main(
...
+            lines = fp.read().splitlines()
+            if lines[0].startswith("@"):
+                # Sometimes those arguments might be also be a redirect
+                with open(lines[0][1:], encoding = "utf-8") as f:
+                    args.extend(f.read().splitlines())
+            else:
+                # First argument is the tool name
+                args.extend(lines[1:])
+
+    print(args)
```
yongjincho92, line 53: nit : probably delete this print unless it's here for debugging purposes? @brentleyjones

brentleyjones: Thanks, I'll remove.

PR #1876, opened 2023-03-16
Create generated generators in an external repository

brentleyjones:
This removes the need to handle a temporary directory in-repo.


erikkerber (commented):
I think this would work well, or at least can't now think of a reason it's a bad idea

PR #1902, opened 2023-03-22
Allow non-Xcode targets to forward first-order potential target merges

mattrobmattrob:
This change does a handful of things:

1. Enable each target type (`library_targets`, `non_xcode_targets`, and `top_level_targets`) to forward its possible target merges upstream.
2. Add `mergable_xcode_library_targets` to `XcodeProjInfo` to have a direct list of potential target merges to use in the processing of top level targets. This allows each target to bubble up desired possible merges for itself.

Work towards https://github.com/buildbuddy-io/rules_xcodeproj/issues/1724. When using https://github.com/mattrobmattrob/rules_xcodeproj_playground/tree/mr/rules_ios.swift.fwk, this allows `@//SmallFramework:SmallFrameworkTests_swift` (the underlying test library's Swift code) to be merged into `@//SmallFramework:SmallFrameworkTests.__internal__.__test_bundle` (the unit test target) in Xcode because `@//SmallFramework:SmallFrameworkTests_objc` is `not should_create_xcode_target` so it only has a single possible merge source (after redoing https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/1903, which will come in a follow up).

|Before|After (with https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/1903)|
|---|---|
|Screenshot 2023-03-22 at 12 53 47 PM|Screenshot 2023-03-22 at 12 53 25 PM|


xcodeproj/internal/providers.bzl
```diff
@@ -151,6 +151,11 @@ added to the project.
 """,
         "label": "The `Label` of the target.",
         "lldb_context": "A value returned from `lldb_context.collect`.",
+        "direct_potential_target_merges": """\
```
brentleyjones, line 154: nit:  move up to below `dependencies`

xcodeproj/internal/top_level_targets.bzl
```diff
@@ -425,6 +425,18 @@ def process_top_level_target(
...
+        # providers resulted in an empty set of potential target merges.
+        if not potential_target_merges:
+            potential_target_merges = [
+                struct(
+                    src = mergeable_target.src,
+                    dest = id,
+                )
+                for info in deps_infos
+                for mergeable_target in info.direct_potential_target_merges
+            ]
```
brentleyjones, line 439: Does this only go one layer down? Ideally this would go as many layers as needed.

With the `dependencies` handling here: https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/2667f91abe73112f3552543922edd0dec6e6a46b/xcodeproj/internal/target_properties.bzl#L41-L51

we handle this at the level of the target being excluded, not at the top-level (like this)

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -442,13 +442,19 @@ actual targets: {}
...
         src_label = bazel_labels.normalize_label(src_target.label)
         dest_target = unprocessed_targets[merge.dest]
         dest_label = bazel_labels.normalize_label(dest_target.label)
         if src_label in unfocused_labels or dest_label in unfocused_labels:
             continue
-        raw_target_merge_dests.setdefault(merge.dest, []).append(merge.src.id)
+
+        # Only include `should_create_xcode_target` targets.
+        if merge.src.id in focused_targets:
+            raw_target_merge_dests.setdefault(merge.dest, []).append(merge.src.id)
```
brentleyjones, line 457: I think this change stands on its own? If so, can you open a PR for just this part?

mattrobmattrob: ✅  https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/1903

xcodeproj/internal/top_level_targets.bzl
mattrobmattrob: Yup, only one layer. In this case, it's sorta handled by the target _and_ the top-level. I didn't want to join up the second-order potential merge targets unless the first-order resulted in none being present vs. conjoining multiple first-order and potentially second-order merge targets at the same time. Possible that's not a valid concern though.

But let me look at the linked code snippet to see if I can integrate it a different way. 👍

mattrobmattrob:
No need to review. Trying to see how the tests end up with this change.

xcodeproj/internal/xcodeproj_rule.bzl
brentleyjones: Just a reminder that #1903 was reverted, so you might have to re-apply something similar.

brentleyjones (commented):
Looks good. I want @chiragramani and @luispadron to verify it on their projects as well.

Is there a way to adjust `examples/integration` to show something merging that wouldn't previously merge (so we see this in our fixtures and ensure we don't regress it)?

mattrobmattrob:
> Looks good. I want @chiragramani and @luispadron to verify it on their projects as well.
> 
> Is there a way to adjust `examples/integration` to show something merging that wouldn't previously merge (so we see this in our fixtures and ensure we don't regress it)?

Without redoing https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/1903, this won't show any change right now, AFAIK. So I can add any `example/integration` additions as part of that work, if you're cool with that.

brentleyjones (approved):

mattrobmattrob:
> Ahh, I missed this line in the description:

Thought to add it for clarity then added it after you asked. Sorry. 🤦

luispadron:
Going to check this out now!

thiagohmcruz:
I pulled and re-checked the same thing I did last time using a target in `rules_ios` and looks good!

chiragramani:
Thanks a lot, @mattrobmattrob. Just tested the PR and I confirm that this fixes the target merging issue that we have had. For us also, the underlying test library wasn't merged with the test bundle. glad to see the fix!

chiragramani (approved):

PR #1905, opened 2023-03-23
Remove unused Swift declarations in `tools/generator/...`

jpsim:
This was found with SwiftLint's upcoming improved dead code analysis.

Let me know if some of this stuff should be kept for future use.

Signed-off-by: JP Simard <jp@jpsim.com>


mattrobmattrob (commented):
Will def to @brentleyjones here. 👍

brentleyjones (approved):
I would love to have a fast dead code checker to run ok every commit 😉.

PR #1906, opened 2023-03-23
Add tests for `DefaultLogger`

jpsim:
Not terribly exciting right now, but this will be useful to validate the coloring support I'm working on.


tools/generator/src/Logger.swift
```diff
@@ -10,33 +10,57 @@ protocol Logger {
...
+    }
+}
+
 /// The logger that is used when not running tests.
-class DefaultLogger: Logger {
+final class DefaultLogger: Logger {
     private var standardError = StderrOutputStream()
+    private var testOutputStream: StringOutputStream?
+
+    func overrideOutputStreamForTests(_ outputStream: StringOutputStream) {
```
brentleyjones, line 33: Instead of this, could you have the initializer take in the output streams (stdout (which would be new) and stderr), with default values set? Having test specific code feels a bit off to me.

tools/generator/src/Generator/Main.swift
```diff
@@ -7,7 +7,7 @@ import ZippyJSON
 extension Generator {
     /// The entry point for the `generator` tool.
     static func main() async {
-        let logger = DefaultLogger()
+        let logger = DefaultLogger(standardError: StderrOutputStream(), standardOutput: StdoutOutputStream())
```
brentleyjones, line 10: Can you have these be the default values so the normal call site doesn't have to change?

tools/generator/src/Logger.swift
```diff
@@ -10,33 +9,46 @@ protocol Logger {
...
+final class StdoutOutputStream: TextOutputStream {
+    func write(_ string: String) {
+        fputs(string, stdout)
+    }
+}
+
 /// The logger that is used when not running tests.
-class DefaultLogger: Logger {
-    private var standardError = StderrOutputStream()
+final class DefaultLogger<E: TextOutputStream, O: TextOutputStream>: Logger {
```
brentleyjones, line 26: Does it need to be generic? Does `print()` not work with protocols?

tools/generator/src/Generator/Main.swift
jpsim: Unfortunately not, because `DefaultLogger` is now generic over `E` and `O`.

There's also a Swift compiler crash if I try to use type erased protocols. 🤦

jpsim: Here's the crash. I'm trying to find a nicer workaround now.

```
1.	Apple Swift version 5.7.2 (swiftlang-5.7.2.135.5 clang-1400.0.29.51)
2.	Compiling with the current language version
3.	While evaluating request ExecuteSILPipelineRequest(Run pipelines { PrepareOptimizationPasses, EarlyModulePasses, HighLevel,Function+EarlyLoopOpt, HighLevel,Module+StackPromote, MidLevel,Function, ClosureSpecialize, LowLevel,Function, LateLoopOpt, SIL Debug Info Generator } on SIL for generator)
4.	While running pass #1224490 SILFunctionTransform "DCE" on SILFunction "@$s9generator9GeneratorC20setXCCurrentVersions33_EC5E7CBC9BE37AFF2E062D2830B09400LL15xcVersionGroups17xccurrentversions6loggerySDyAA8FilePathV9XcodeProj14XCVersionGroupCG_SayAA0dO0VGAA6Logger_ptKFZ".
 for 'setXCCurrentVersions(xcVersionGroups:xccurrentversions:logger:)' (at tools/generator/src/Generator/CreateFilesAndGroups.swift:606:13)
Stack dump without symbol names (ensure you have llvm-symbolizer in your PATH or set the environment var `LLVM_SYMBOLIZER_PATH` to point to it):
0  swift-frontend           0x00000001071ef5b0 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) + 56
1  swift-frontend           0x00000001071ee5b4 llvm::sys::RunSignalHandlers() + 112
2  swift-frontend           0x00000001071efc34 SignalHandler(int) + 344
3  libsystem_platform.dylib 0x000000019e5d02a4 _sigtramp + 56
4  swift-frontend           0x0000000103174de8 (anonymous namespace)::DCE::markValueLive(swift::SILValue) + 244
5  swift-frontend           0x0000000103174de8 (anonymous namespace)::DCE::markValueLive(swift::SILValue) + 244
6  swift-frontend           0x0000000103172500 (anonymous namespace)::DCEPass::run() + 3648
7  swift-frontend           0x00000001030f0470 swift::SILPassManager::runFunctionPasses(unsigned int, unsigned int) + 4020
8  swift-frontend           0x00000001030ed0c4 swift::SILPassManager::executePassPipelinePlan(swift::SILPassPipelinePlan const&) + 148
9  swift-frontend           0x0000000103108670 swift::SimpleRequest<swift::ExecuteSILPipelineRequest, std::__1::tuple<> (swift::SILPipelineExecutionDescriptor), (swift::RequestFlags)1>::evaluateRequest(swift::ExecuteSILPipelineRequest const&, swift::Evaluator&) + 56
10 swift-frontend           0x00000001030f5858 llvm::Expected<swift::ExecuteSILPipelineRequest::OutputType> swift::Evaluator::getResultUncached<swift::ExecuteSILPipelineRequest>(swift::ExecuteSILPipelineRequest const&) + 504
11 swift-frontend           0x00000001030f79a0 swift::runSILOptimizationPasses(swift::SILModule&) + 368
12 swift-frontend           0x00000001028a9694 swift::CompilerInstance::performSILProcessing(swift::SILModule*) + 524
13 swift-frontend           0x000000010283a5cc performCompileStepsPostSILGen(swift::CompilerInstance&, std::__1::unique_ptr<swift::SILModule, std::__1::default_delete<swift::SILModule> >, llvm::PointerUnion<swift::ModuleDecl*, swift::SourceFile*>, swift::PrimarySpecificPaths const&, int&, swift::FrontendObserver*) + 1064
14 swift-frontend           0x000000010283cce8 swift::performFrontend(llvm::ArrayRef<char const*>, char const*, void*, swift::FrontendObserver*) + 7284
15 swift-frontend           0x00000001027dd294 swift::mainEntry(int, char const**) + 3940
16 dyld                     0x000000019e277e50 start + 2544
```

jpsim: Weird, doesn't crash if I have my colorize integration. It's small enough that I'll just extend this PR to also do that here.

tools/generator/src/Logger.swift
jpsim: See the other thread on the swift compiler crash.

tools/generator/src/Generator/Main.swift
jpsim: Hmm, still crashed in some cases. Filed https://github.com/apple/swift/issues/64559.

brentleyjones (approved):

PR #1907, opened 2023-03-23
Colorize logs

jpsim:
Colors and styling should match main Bazel logs.

* Debug: yellow
* Info: green
* Warning: magenta
* Error: red & bold

Some sample screenshots:

image

Can be disabled with `--@rules_xcodeproj//xcodeproj:color=no` for cases where ansi control codes won't be rendered properly.


tools/generator/src/Logger.swift
```diff
@@ -33,22 +33,24 @@ final class DefaultLogger<E: TextOutputStream, O: TextOutputStream>: Logger {
     }
 
     func logDebug(_ message: @autoclosure () -> String) {
-        // TODO: Colorize
```
brentleyjones, line 36: Thanks for doing this!

One of the main reasons I didn't tackle this yet, was I wanted to make sure we _didn't_ colorize in certain cases. For a long time I believed that would correlate with Bazel's setting, but then I realized that the output would be cached so maybe it shouldn't? Then I realized we don't cache the generation for various reasons.

All of that is to say I still think we should conditionally do this. I think we should have two implementations, versus a bool. And for the condition, ideally we can detect the color mode for Bazel, but if not, we can add a Starlark build setting ourselves. The main reason I think we should allow it to be turned off, is some people run this on CI that doesn't support color codes, and it will make the output all funny for them.

tools/generator/src/Logger.swift
```diff
@@ -33,22 +33,24 @@ final class DefaultLogger<E: TextOutputStream, O: TextOutputStream>: Logger {
...
     func logWarning(_ message: @autoclosure () -> String) {
-        // TODO: Colorize
-        print("WARNING: \(message())", to: &self.standardError)
+        print("WARNING: \(message())".yellow, to: &self.standardError)
     }
 
     func logError(_ message: @autoclosure () -> String) {
-        // TODO: Colorize
-        print("ERROR: \(message())", to: &self.standardError)
+        print("ERROR: \(message())".red, to: &self.standardError)
```
brentleyjones, line 48: To match Bazel, can we only colorize the prefix? And also, to match Bazel, let's make error bold, info green, debug yellow (as well as warning as it is now).

tools/generator/src/Logger.swift
jpsim: I started writing a utility that would detect if the generator was running in a TTY that supported colors, but ultimately that wouldn't work because Bazel doesn't pass through the `TERM` environment variable.

So after speaking with @brentleyjones , he suggested that we add an `xcodeproj` macro argument for `colorize`, which is now what I've done.

docs/BUILD
```diff
@@ -50,7 +50,7 @@ done
 
 diff_test(
     name = "diff_test",
-    failure_message = "\nPlease update the docs by running\n    bazel run //docs:update",
+    failure_message = "\nPlease update the docs by running\n    bazel run --noenable_bzlmod //docs:update",
```
brentleyjones, line 53: Can you remove in favor of #1910?

tools/generator/src/Generator/Main.swift
```diff
@@ -7,10 +7,17 @@ import ZippyJSON
...
-        let logger = DefaultLogger(standardError: StderrOutputStream(), standardOutput: StdoutOutputStream())
+        let logger = DefaultLogger(
+            standardError: StderrOutputStream(),
+            standardOutput: StdoutOutputStream(),
+            colorize: true
+        )
 
         do {
             let arguments = try parseArguments(CommandLine.arguments)
+            if !arguments.colorize {
```
luispadron, line 18: Nice!

luispadron (approved):

tools/generator/src/Logger.swift
```diff
@@ -22,33 +22,66 @@ final class StdoutOutputStream: TextOutputStream {
...
+            to: &self.standardError
+        )
     }
 
-    func logError(_ message: @autoclosure () -> String) {
-        // TODO: Colorize
-        print("ERROR: \(message())", to: &self.standardError)
+    func logError(_ message: String) {
+        print(
+            self.format(color: .red, prefix: "ERROR", message: message),
```
brentleyjones, line 83: Just so it's still tracked: we want this to be bold as well.

jpsim: It _is_ bold, all colors are bold now. It's the `;1` ANSI code.

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -1818,6 +1823,7 @@ def make_xcodeproj_rule(
             default = Label("//xcodeproj/internal:installer.template.sh"),
         ),
         "_is_fixture": attr.bool(default = is_fixture),
+        "_colorize": attr.bool(default = colorize),
```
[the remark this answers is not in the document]
brentleyjones: Sure. Sorry for not adding more detail.

To make the attribute public, remove the `_`. That will mean that the generated `BUILD` file `xcodeproj` will need a new line for `colorize`. You would add that [here](https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/5b705c26632bb5458f86c70db847f7c3875237c1/xcodeproj/internal/generator.BUILD.template.bazel#L11) and [here](https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/5b705c26632bb5458f86c70db847f7c3875237c1/xcodeproj/internal/xcodeproj_runner.bzl#L227).

In `xcodeproj_runner.bzl`, in order to pass the value to the template, you would use `str(ctx.attr._colorize[BuildSettingInfo].value)`.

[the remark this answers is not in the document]
jpsim: ok, I think I got it, sorry for needing so much hand holding.

brentleyjones (approved):

xcodeproj/internal/xcodeproj_rule.bzl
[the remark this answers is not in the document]
brentleyjones: No problem! Thanks for tackling this.

PR #1917, opened 2023-03-24
Respect `--output_base` set on command-line

brentleyjones:


xcodeproj/internal/runner.template.sh
```diff
@@ -101,11 +102,7 @@ fi
 
 installer_flags+=(--bazel_path "$bazel_path")
 
-if [[ %is_fixture% -eq 1 ]]; then
-  execution_root=$("$bazel_path" info --config=fixtures execution_root)
-else
-  execution_root=$("$bazel_path" info execution_root)
-fi
```
brentleyjones, line 108: This technically speeds up generation a bit (super minor), since the file is cached

mattrobmattrob (approved):
Stoked about this also not calling `bazel info` anymore too. 🥳

yongjincho92:
Thanks so much for fixing this!

PR #1926, opened 2023-03-29
Fix handling of bzlmod external targets

brentleyjones:
Closes #1925.

Our regex wasn’t account for the fact that external repos under bzlmod would have a `@@` prefix instead of a single `@`.


mattrobmattrob (approved):

BalestraPatrick (approved):

keith (approved):
validated this fixes the issue in our actual setup as well, thanks!

PR #1930, opened 2023-03-30
Associate non-Xcode target source files with the correct label

brentleyjones:
Before this change the extra files would be associated with the next target(s) down the dependency tree. This now allows you to properly unfocus the source files of targets that won't become Xcode targets.


chiragramani (approved):
Thanks a lot! 🙏🏼 this fixes the target focus issue for us!

thii (approved):

Issue #1922, opened 2023-03-27
Feature Request: Can we support linking against shared libraries (.dylib)?


yrom:
See my test codes in fork:

https://github.com/yrom/rules_xcodeproj/commit/4d8cf7012f67a0dc778d3546e841c7f50ac08463

```sh
cd examples/cc
bazel run //:xcodeproj # failed
```

<details><summary> ERROR message</summary>
<pre>
Generating "cc.xcodeproj"
ERROR: ~/rules_xcodeproj/examples/cc/lib3/BUILD:32:11: in @rules_xcodeproj~override//xcodeproj/internal:default_automatic_target_processing_aspect.bzl%default_automatic_target_processing_aspect,@rules_xcodeproj~override//xcodeproj/internal:xcodeproj_rule.bzl%bwb_xcodeproj_aspect aspect on cc_library rule //lib3:lib3: 
Traceback (most recent call last):
	File "~/rules_xcodeproj/examples/cc/bazel-output-base/rules_xcodeproj/build_output_base/external/rules_xcodeproj~override/xcodeproj/internal/xcodeproj_aspect.bzl", line 65, column 36, in _xcodeproj_aspect_impl
		info = create_xcodeprojinfo(
	File "~/rules_xcodeproj/examples/cc/bazel-output-base/rules_xcodeproj/build_output_base/external/rules_xcodeproj~override/xcodeproj/internal/xcodeprojinfo.bzl", line 566, column 44, in create_xcodeprojinfo
		info_fields = _create_xcodeprojinfo(
	File "~/rules_xcodeproj/examples/cc/bazel-output-base/rules_xcodeproj/build_output_base/external/rules_xcodeproj~override/xcodeproj/internal/xcodeprojinfo.bzl", line 422, column 50, in _create_xcodeprojinfo
		processed_target = process_library_target(
	File "~/rules_xcodeproj/examples/cc/bazel-output-base/rules_xcodeproj/build_output_base/external/rules_xcodeproj~override/xcodeproj/internal/library_targets.bzl", line 74, column 47, in process_library_target
		linker_inputs = linker_input_files.collect(
	File "~/rules_xcodeproj/examples/cc/bazel-output-base/rules_xcodeproj/build_output_base/external/rules_xcodeproj~override/xcodeproj/internal/linker_input_files.bzl", line 55, column 65, in _collect_linker_inputs
		primary_static_library = _compute_primary_static_library(
[...]

PR #1942, opened 2023-03-30
Place codesigned executable next to the base executable

brentleyjones:
Part of #1922.

This is needed in order to allow Bazel-set rpaths to work (e.g. to support `linkshared`).


mattrobmattrob (approved):

brentleyjones:
I need to land one stacked changed to account for the filename changing.

brentleyjones:
Not sure why this is now causing an ASan violation and crashes in the tests, but I assume a swiftc miscompile. I'll dig in more soon.

PR #1944, opened 2023-03-30
Symlink binaries instead of copying them

brentleyjones:
Part of #1922.

This is needed to allow for Bazel-set rpaths (BwB or unfocused BwX targets) to work.


brentleyjones:
A downside to this: you can't easily switch between BwX and BwB without cleaning derived data, since Xcode won’t replace the symlink with a binary. You sorta have to do that anyway when switching modes for other reasons, but still a downside.

brentleyjones:
Pretty sure this broke sources appearing correctly in Instruments. I'm looking into it more.

brentleyjones:
Was unrelated.

Issue #1724, opened 2023-02-01
Feature Request: Hide extraneous targets created by `rules_ios` macros


luispadron:
`rules_ios` seems to create a target `_objc` and/or `_swift` for each framework which in turn causes them to show up in the Xcode target navigator.

It would be nice to support hiding these somehow as a default.

`rules_xcodeproj`: 9ab3f0d3751eada941e3d1207b392c6e9b8deb01

<img width="331" alt="Screenshot 2023-02-01 at 2 47 36 PM" src="https://user-images.githubusercontent.com/13840545/216148026-a8b0e929-0c70-4f6d-aa94-3be56ac9b534.png">

brentleyjones:
This would require changes to target merging. Not sure if it's possible or not, but I can look into it.

As for just not creating the schemes, custom schemes or eventually https://github.com/buildbuddy-io/rules_xcodeproj/issues/857 would prevent the schemes from being created.

mattrobmattrob:
As mentioned in the Bazel Slack, super simple replication for this issue in https://github.com/mattrobmattrob/rules_xcodeproj_playground/tree/mr/rules_ios.swift.fwk (just `bazel run :Project` to exercise `rules_xcodeproj`).

Actively working on this.

mattrobmattrob:
This isn't necessarily done but should now have `ios_unit_test`/`ios_ui_test` merged. Next up, would be the actual intra-graph merges for `apple_framework`'s _swift and _objc suffixed targets.

brentleyjones:
@luispadron is this addressed with incremental generation?

luispadron:
Yes! I think we can close it 🚀

brentleyjones:
Thanks! I'll tie it to that milestone and close it then.

PR #1947, opened 2023-03-30
Add ability to merge `xcode_target`s to dest

mattrobmattrob:
Work towards https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/1724.

### DIFF
--- a/xcodeproj/internal/xcode_targets.bzl
+++ b/xcodeproj/internal/xcode_targets.bzl
@@ -3,7 +3,7 @@
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("@bazel_skylib//lib:structs.bzl", "structs")
-load(":collections.bzl", "set_if_true", "uniq")
+load(":collections.bzl", "flatten", "set_if_true", "uniq")
 load(
     ":files.bzl",
     "build_setting_path",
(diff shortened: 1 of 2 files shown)


xcodeproj/internal/xcode_targets.bzl
```diff
@@ -261,45 +261,59 @@ def _to_xcode_target_product(product):
...
             dest = dest.product,
         ),
-        is_swift = src.is_swift,
+        is_swift = is_swift,
         test_host = dest._test_host,
         build_settings = build_settings,
-        conlyopts = src._conlyopts or dest._conlyopts,
-        cxxopts = src._cxxopts or dest._cxxopts,
-        swiftcopts = src._swiftcopts or dest._swiftcopts,
+        conlyopts = flatten([src._conlyopts for src in srcs]) or dest._conlyopts,
```
brentleyjones, line 306: Since we are going to guarantee that there is only one swift and one non-swift, can we instead select the swift copts from the swift one and the non-swift copts from the other target? Then we don't need to flatten.

xcodeproj/internal/xcode_targets.bzl
```diff
@@ -261,45 +261,59 @@ def _to_xcode_target_product(product):
...
+        cxxopts = flatten([src._cxxopts for src in srcs]) or dest._cxxopts,
+        swiftcopts = flatten([src._swiftcopts for src in srcs]) or dest._swiftcopts,
         search_paths = dest._search_paths,
-        modulemaps = src._modulemaps,
-        swiftmodules = src._swiftmodules,
-        inputs = _merge_xcode_target_inputs(
-            src = src.inputs,
+        modulemaps = flatten([src._modulemaps for src in srcs]),
+        swiftmodules = flatten([src._swiftmodules for src in srcs]),
```
brentleyjones, line 311: Same for these two: only gather from the swift target.

xcodeproj/internal/xcode_targets.bzl
```diff
@@ -308,26 +322,51 @@ def _merge_xcode_target(*, src, dest):
...
+            ],
             dest = dest.outputs,
         ),
         lldb_context = dest.lldb_context,
         xcode_required_targets = dest.xcode_required_targets,
     )
 
-def _merge_xcode_target_inputs(*, src, dest):
+def _merge_xcode_targets_inputs(*, src_inputs, dest):
+    pchs = [src.pch for src in src_inputs if src.pch]
```
brentleyjones, line 340: Same comment, but this should only come from the non-swift target. I would pass on two inputs, one for each of the two targets.

xcodeproj/internal/xcode_targets.bzl
```diff
@@ -261,45 +261,59 @@ def _to_xcode_target_product(product):
         _additional_files = product.framework_files,
     )
 
-def _merge_xcode_target(*, src, dest):
-    """Creates a new `xcode_target` by merging the values of `src` into `dest`.
+def _merge_xcode_targets(*, srcs, dest):
```
brentleyjones, line 264: Based on my other comments, I would pass in two arguments here: `swift_src` and `non_swift_src`. We can make it clear that we can merge up to 2 targets, and grab specific things from each one.

mattrobmattrob:
Sounds good. Trying to see what failure this will cause and can refactor to make it more portable. 👍

mattrobmattrob:
Going to replace with a more wholistic approach with @brentleyjones's suggestions.

PR #1950, opened 2023-04-03
Set a more restrictive `PATH` for Bazel

brentleyjones:
We now set the resolved path in the generated project, along with `BAZEL_REAL` if it was set, to prevent the need to have a more permissive `PATH`.


thiagohmcruz (approved):
Ran the same tests as last time while we were debugging this and looks good!

PR #1952, opened 2023-04-03
Add `rules_ios` example and tests

mattrobmattrob:
Loosely copy off of the `examples/integration` but switch to `rules_ios` as the ruleset. This removes various things like app clip, watchOS app/extensions, and so on to keep it simpler for now.

Intentionally get coverage on the following:
- Mixed, Swift, and ObjC language unit test.
- Mixed, Swift, and ObjC language `apple_framework`.
- UI test.
- `link_dynamic = True` & `link_dynamic = False` declared `apple_framework`.
- iOS extension via `//WidgetExtension`.


mattrobmattrob:
Fixture generation is currently broken in my local so will continue to iterate on this as I have time.

examples/rules_ios/.bazelrc
```diff
@@ -0,0 +1,29 @@
...
+build:rules_xcodeproj --apple_generate_dsym
+
+build --test_env=INTEGRATION_TEST_ENV_VAR=TRUE
+
+# rules_ios
+build --features apple.virtualize_frameworks
+build --features apple.arm64_simulator_use_device_deps
+
+# Use a user.bazelrc if it exists
+try-import %workspace%/user.bazelrc
```
mattrobmattrob, line 29: Should this just be `try-import %workspace%/../integration/.bazelrc` for simplicity/lack of duplication?

examples/rules_ios/.bazelrc
```diff
@@ -0,0 +1,29 @@
+# Import parent workspace settings
+import %workspace%/../../shared.bazelrc
+
+# https://github.com/bazel-ios/rules_ios/issues/672
+build --noexperimental_strict_conflict_checks
+
+# Exercise the extra flags feature
```
brentleyjones, line 7: I would make this the bare minimum (so see what is in Sanitizers?) and then add what is needed for rules_ios.

examples/rules_ios/BUILD
```diff
@@ -0,0 +1,73 @@
...
+#        --generator_output_groups=all_targets \
+#        'build --remote_download_minimal'
+#
+#      bazel run --config=cache //:xcodeproj-sim_arm64 \
+#        -- \
+#        --generator_output_groups=all_targets \
+#        'build --remote_download_minimal'
+#
+[
+    xcodeproj(
```
brentleyjones, line 48: I wouldn't do the complicated cache setup here. So removing all the stuff related to multiple cpu types to make it super simple is best.

examples/rules_ios/Lib/README.md
```diff
@@ -0,0 +1,4 @@
+# CommandLineTool
+
+This is an example of an extra file that is added to the Xcode project via
```
brentleyjones, line 3: I need for this. In general all the "integration" stuff can get culled. A concise example will be better than the kitchen sink that Integration is trying to be.

mattrobmattrob: SGTM

examples/rules_ios/iOSApp/Source/Utils/BUILD
```diff
@@ -0,0 +1,23 @@
+load(
+    "@build_bazel_rules_ios//rules:framework.bzl",
+    rules_ios_apple_framework = "apple_framework",
+)
+
+rules_ios_apple_framework(
+    name = "Utils",
+    srcs = glob(
+        ["**/*.m"],
+        # Headers included here on purpose, to test `hdrs` collection logic
```
brentleyjones, line 10: Same

examples/rules_ios/test/fixtures/BUILD
```diff
@@ -0,0 +1,34 @@
...
+)
+
+xcodeproj_fixture(
+    scheme_autogeneration_mode = SCHEME_AUTOGENERATION_MODE,
+    top_level_targets = XCODEPROJ_TARGETS,
+)
+
+_FIXTURES = [
+    ":xcodeproj_bwb",
+    # Disable since VFS is enabled by default for 'rules_ios' examples.
```
brentleyjones, line 20:
```suggestion
    # Disabled since VFS is enabled by default for 'rules_ios' examples.
```

examples/rules_ios/xcodeproj.bazelrc
```diff
@@ -0,0 +1,5 @@
+# Import parent workspace settings
+import %workspace%/../../xcodeproj.bazelrc
+
+# Disabling archived bundles for fixture coverage
+build:rules_xcodeproj --define=apple.experimental.tree_artifact_outputs=1
```
brentleyjones, line 5: Probably don't need?

examples/rules_ios/xcodeproj_targets.bzl
```diff
@@ -0,0 +1,23 @@
...
+    "top_level_targets",
+)
+
+XCODEPROJ_TARGETS = [
+    top_level_targets(
+        labels = [
+            "//iOSApp",
+            "//Lib:LibDynamic",
+        ],
+        target_environments = ["simulator"],
```
brentleyjones, line 14: This is the default, do you want `device` here as well? I would assume so to show off device support for rules_ios?

mattrobmattrob: I had left it off to avoid provisioning profile stuff. Happy to include it if you think it's valuable. That stuff just ends up forwarded to the underlying `rules_apple` things so there's likely not a ton of benefit explicitly WRT `rules_ios` functionality.

brentleyjones: I would like it, yes.

examples/rules_ios/iOSApp/Source/Utils/BUILD
mattrobmattrob: Same comment as above.

examples/rules_ios/Lib/BUILD
```diff
@@ -0,0 +1,44 @@
...
+    name = "LibDynamic",
+    bundle_id = "rules-xcodeproj.LibFramework",
+    infoplists = ["//Lib:Info.plist"],
+    link_dynamic = True,
+    module_name = "Lib",
+    platforms = {
+        "ios": "15.0",
+    },
+    visibility = ["@rules_xcodeproj//xcodeproj:generated"],
+    xcconfig = {
```
luispadron, line 40: Do we want to test the `_by_build_setting` ones for this and infoplists

examples/rules_ios/xcodeproj_targets.bzl
thiagohmcruz: Yep, there's not much going on as far as I remember in `rules_ios` re "on device" attributes. But +1 to adding it anyways if it's simple enough, catching potential future regressions being one benefit here IMO.

luispadron (approved):
Nice! Thanks for adding this will be great for reports and stuff going forward 

Some follow ups we could add:
- new app clip support
- testing xcframework rules

examples/rules_ios/Lib/BUILD
mattrobmattrob: You want to add it in a follow up, @luispadron? We don't use that internally so I'm less familiar with its usage.

thiagohmcruz (approved):
As you mentioned there's more we can cover here but I think this is a great starting point. Thx for adding this!

examples/rules_ios/Lib/BUILD
luispadron: Yep can do!

thiagohmcruz: The `_by_build_setting` attrs are loosely tied to intended usage of `--config` which is a bit different then how `rules_xcodeproj` creates "Xcode Configurations". I think it makes sense to tackle that separately, ideally we would not only cover the `_by_build_setting` attributes but also how to have `rules_ios` run with `rules_xcodeproj`'s "Xcode Configurations" feature.

Probably a good segway to revisit the `_by_build_setting` story in `rules_ios` and see if there's any improvements we can do there to make it more generic and inclusive of other Xcode project generators, it was initially closely related to `rules_ios`'s own generator.

PR #1955, opened 2023-04-04
Fix BwB test action env variables for custom schemes without launch actions

brentleyjones:


tools/generator/src/Extensions/XCScheme+Extensions.swift
```diff
@@ -108,11 +79,44 @@ extension XCScheme {
...
+                        title: "Update .lldbinit",
+                        scriptText: ExecutionAction.createLLDBInitScript,
+                        buildableReference: aTargetInfo.buildableReference
+                    )
+                )
+            }
+            testAction = try .init(
+                buildMode: buildMode,
+                testActionInfo: testActionInfo,
+                launchActionHasRunnable: launchActionHasRunnable,
```
mattrobmattrob, line 112: Would it be a bit simpler just directly testing the actual value?
```suggestion
                launchActionHasRunnable: launchAction.runnable != nil,
```

mattrobmattrob (approved):

tools/generator/src/Extensions/XCScheme+Extensions.swift
brentleyjones: Yeah, probably.

PR #1962, opened 2023-04-04
Exclude more files from `_process_cc_info_headers` processing

brentleyjones:


xcodeproj/internal/input_files.bzl
```diff
@@ -415,7 +415,7 @@ def _collect_input_files(
              compilation_context.direct_textual_headers),
             exclude_headers = exclude_headers,
             pch = pch,
-            srcs = srcs,
+            srcs = srcs + non_arc_srcs + hdrs,
```
yongjincho92, line 418: are the new elements in `non_arc_srcs + hdrs` already being added as params for `_process_cc_info_headers` (e.g. `compilation_context.direct_private_headers + compilation_context.direct_public_headers + compilation_context.direct_textual_headers`)? And I assume adding them to srcs would exclude them from `_process_cc_info_headers`.

brentleyjones: If something is in `srcs`, `non_arc_srcs`, or `hdrs`, the then we are already processing them in some way and we don't want them added to `extra_files`.

yongjincho92 (approved):

xcodeproj/internal/input_files.bzl
brentleyjones: Thinking about this more, I can probably extend it more by having all files that go through `_handle_file` get put in a dictionary (`exclude`) and exclude even more that way.

luispadron (approved):

Issue #1963, opened 2023-04-04
Bug: Index compilation for a source file in a objc_library with implementation_deps fails 


yongjincho92:
### Description

At Snap, we added a change (https://github.com/bazelbuild/bazel/pull/17962) in Bazel fork to support `implementation_deps` for `objc_library` rule. 

We have found that rules_xcodeproj generated Xcode fails to compile a source file that imports a header defined in its implementation_deps. 

Say, ATest is a unit test target and depends on target A. And the target A is an implementation deps of ATest. Building ATest works fine, however, opening `TestA.m` in Xcode editor tap triggers Xcode to index-compile the source file, which fails because path to A.h is not passed as a copt. 

### Reproduction steps

1. Download repro project from https://drive.google.com/file/d/1iixYoxc6Tek8q-VlIqF6udS8_mexCbqs/view?usp=share_link 
2. Extract it and go into dir 
3. run `./bazel run //lib:A.xcodeproj`  (bazel binary bundled together has change https://github.com/bazelbuild/bazel/pull/17962) 
4. run `open lib/A.xcodeproj` 
5. open `TestA.m` and check in Xcode's Index log that the compilation failed due to missing "A.h"

### Expected behavior

When opening TestA.m, Xcode should be able to find "A.h" and compile `TestA.m` successfully. 

### rules_xcodeproj version

1.3.2

### Xcode version

14.1

### Bazel version

upstream/main (b24930f6f90934a0665c7454e12cc6c4e4ecc7b3) + https://github.com/bazelbuild/bazel/pull/17962

brentleyjones:
Thanks for the repro. I'll look into this soon!

yongjincho92:
the bazel binary I bundled together is arm64 executable, so let me know if you need intel cpu version.

brentleyjones:
Apple Silicon is perfect, thanks.

PR #1965, opened 2023-04-04
works but not robust

yongjincho92:
An attempt to fix https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/1963

### DIFF
--- a/xcodeproj/internal/opts.bzl
+++ b/xcodeproj/internal/opts.bzl
@@ -106,7 +106,8 @@ def _get_unprocessed_compiler_opts(
         build_mode,
         has_c_sources,
         has_cxx_sources,
-        target):
+        target,
+        transitive_infos):
     """Returns the unprocessed compiler options for the given target.
 
     Args:
@@ -170,11 +171,16 @@ def _get_unprocessed_compiler_opts(
                 ctx.disabled_features + _UNSUPPORTED_CC_FEATURES
             ),
         )
+        include_directories = depset(
+            compilation_context.includes.to_list(), 
+            transitive = [info.compilation_providers.cc_info.compilation_context.includes for attr, info in transitive_infos if info.compilation_providers.cc_info]
+        )
+
         variables = cc_common.create_compile_variables(
             feature_configuration = feature_configuration,
             cc_toolchain = cc_toolchain,
             user_compile_flags = user_copts,
-            include_directories = compilation_context.includes,
+            include_directories = include_directories,
             quote_include_directories = compilation_context.quote_includes,
             system_include_directories = compilation_context.system_includes,
             framework_include_directories = (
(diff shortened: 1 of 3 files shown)


xcodeproj/internal/opts.bzl
```diff
@@ -170,11 +171,16 @@ def _get_unprocessed_compiler_opts(
                 ctx.disabled_features + _UNSUPPORTED_CC_FEATURES
             ),
         )
+        include_directories = depset(
+            compilation_context.includes.to_list(), 
+            transitive = [info.compilation_providers.cc_info.compilation_context.includes for attr, info in transitive_infos if info.compilation_providers.cc_info]
+        )
```
yongjincho92, line 177: There are a set of hacky and unsafe assumptions in these lines.

brentleyjones: If instead y'all add `local_includes` to https://bazel.build/rules/lib/CompilationContext, similar to https://bazel.build/rules/lib/CompilationContext#local_defines, then we could use that instead of `compilation_context.includes`, and it would work for both normal and implementation deps.

brentleyjones: @ivan-golub

yongjincho92: Good point, let me check on that option!

ivan-golub: thx for suggestion @brentleyjones. This is an option for sure but Im wondering how rules_xcodeproj are handling cc_library implementation deps now? Given that we use same approach for objc_library implementaiton deps, so I would assume we can use same approach instead of introducing new concept for compilation context. Main reason here is that Im actually not 100% sure any of objc impl deps changes will be accepted upstream before objc rules are extracted to starlark by bazel core team and changing base compilation contexts reduced that possibility even further

brentleyjones: We might not be handling them correctly, I haven't checked. And if we aren't, then we would have to do something similar to this (or something similar to Swift, where we grab the arguments from the produced action... but I would really prefer not to go there), which is annoying since `cc_complication_context` currently has all of the information needed to compile the target.

brentleyjones: Looking into the source code, yeah, we probably don't support `cc_library.implementation_deps` either. I'll take a stab at a version of this change.

brentleyjones: We need to do more than includes (like defines as well), so I think what we need to do here is call `cc_common.merge_cc_infos`. Seems we already do that, in one case at least, in `compilation_providers.merge`. I'll adjust that code and pipe it through to fix this.

brentleyjones: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/1967 should fix it.

yongjincho92:
closing in favor of https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/1967

Issue #1956, opened 2023-04-04
Bug: Test suite targets not supported as top level targets


thiagohmcruz:
### Description

If you hit [this code path](https://github.com/bazel-ios/rules_ios/blob/master/rules/test.bzl#L128) in `rules_ios` you'll get errors during project generation:
```sh
ERROR: /private/var/tmp/_bazel_thiago/122885c1fe4a2c6ed7635584956dfc9d/rules_xcodeproj/build_output_base/external/rules_xcodeproj_generated/generator/BUILD:6:10: in xcodeproj rule @rules_xcodeproj_generated//generator:generator: 
Traceback (most recent call last):
        File "/private/var/tmp/_bazel_thiago/122885c1fe4a2c6ed7635584956dfc9d/rules_xcodeproj/build_output_base/external/rules_xcodeproj/xcodeproj/internal/xcodeproj_rule.bzl", line 1390, column 26, in _xcodeproj_impl
                i = [_process_dep(dep) for dep in targets]
        File "/private/var/tmp/_bazel_thiago/122885c1fe4a2c6ed7635584956dfc9d/rules_xcodeproj/build_output_base/external/rules_xcodeproj/xcodeproj/internal/xcodeproj_rule.bzl", line 121, column 13, in _process_dep
                fail("""
Error in fail: 
'@//tests/macos/xcodeproj:Single-Application-RunnableTestSuite' is not a top-level target, but was listed in `top_level_targets`. Only list top-level targets (e.g. binaries, apps, tests, or distributable frameworks) in `top_level_targets`. Schemes and `focused_targets`/`unfocused_targets` can refer to dependencies of targets listed in `top_level_targets`, and don't need to be listed in `top_level_targets` themselves.

If you feel this is an error, and `test_suite` targets should be recognized as top-level targets, file a bug report here: https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/new?template=bug.md
```

### Reproduction steps

* Checkout this branch: [thiago/repro-test-suite-not-supported-rules-xcodeproj](https://github.com/bazel-ios/rules_ios/compare/master...thiago/repro-test-suite-not-supported-rules-xcodeproj)
* Run `bazel run tests/ios/app:foo`
* You should hit the same err reported here

### Expected behavior

[...]

PR #1968, opened 2023-04-05
Add support for ios_*_test_suite targets as top level targets

thiagohmcruz:
Resolves https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/1956

- [ ] TBD: Add tests (both for integration and rules_ios)

Initial intend was to add support for `test_suite` rules and map each runner in `runners` to targets in Xcode, this is doable but it creates indexing issues due to two targets sharing paths. Ended up changing the approach to simply skip the runners (like it's done for non-suite rules already) and let Xcode be the runner, only propagating the bundle. This is probably similar to how Tulsi handled this based on the [docs](https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-ios.md#ios_unit_test):

>Builds and bundles an iOS Unit .xctest test bundle. Runs the tests using the provided test runner when invoked with bazel test. When using Tulsi to run tests built with this target, runner will not be used since Xcode is the test runner in that case.

The only catch here is that `replacement _labels` will contain many underlying labels associated with the same target ID. So added logic to unify that and bring back a `id <=> label` 1-1 relationship, e.g., if these are found in this id => labels mapping exists in the replacement labels:
```py
{
  "@//tests/macos/xcodeproj:Single-Application-RunnableTestSuite.__internal__.__test_bundle applebin_ios-ios_sim_arm64-dbg-ST-d1716b12dfa6": [
    Label("//tests/macos/xcodeproj:Single-Application-RunnableTestSuite_iPad-Air-2__16.2"), 
    Label("//tests/macos/xcodeproj:Single-Application-RunnableTestSuite_iPhone-13-Pro__16.2")
  ]
}
```

Then the longest common value between the labels and the label contained in the id will be used to bring back a 1-1 relationship:
```py
{
  "@//tests/macos/xcodeproj:Single-Application-RunnableTestSuite.__internal__.__test_bundle applebin_ios-ios_sim_arm64-dbg-ST-d1716b12dfa6": Label("@//tests/macos/xcodeproj:Single-Application-RunnableTestSuite"),
}
```

### DIFF
--- a/xcodeproj/internal/target_id.bzl
+++ b/xcodeproj/internal/target_id.bzl
@@ -37,10 +37,11 @@ def _longest_common_prefix(labels):
     return res
 
 def calculate_replacement_label(*, id, replacement_labels):
-    """Calculates a single replacement label for a given id and list of replacement labels
+    """Calculates a single replacement label for a given id and list of \
+    replacement labels.
 
-    It considers the list of labels received plus the label contained in the id itself, then it finds
-    the longest common prefix between those
+    It considers the list of labels received plus the label contained in the id
+    itself, then it finds the longest common prefix between those.
 
     Args:
         id: An id that uniquely represents a target (see `get_id`)
--- a/xcodeproj/internal/target_id.bzl
+++ b/xcodeproj/internal/target_id.bzl
@@ -17,11 +17,13 @@ def get_id(*, label, configuration):
 def _longest_common_prefix(labels):
     if not labels:
         return None
+
     # Assume the shortest string is the answer
     res = labels[0]
     for s in labels:
         if len(s) < len(res):
             res = s
+
     # Compare one character at a time with the rest,
     # as soon as any difference is found the result holds the longest common prefix
     index = 0
@@ -48,10 +50,10 @@ def calculate_replacement_label(*, id, replacement_labels):
         The longest common prefix between the list of labels and the label contained in the id,
         if a longest common prefix is not found the label contained in the id is returned
     """
-    id_label = id.split(" ")[0] # This assumes the id follows the creation pattern in `get_id`
+    id_label = id.split(" ")[0]  # This assumes the id follows the creation pattern in `get_id`
 
     res = _longest_common_prefix(
-        [id_label] + ["%s" % label for label in replacement_labels]
+        [id_label] + ["%s" % label for label in replacement_labels],
     )
     if not res:
         return None
--- a/xcodeproj/internal/target_id.bzl
+++ b/xcodeproj/internal/target_id.bzl
@@ -1,5 +1,7 @@
 """Functions dealing with Target IDs."""
 
+load(":bazel_labels.bzl", "bazel_labels")
+
 def get_id(*, label, configuration):
     """Generates a unique identifier for a target.
 
@@ -11,6 +13,51 @@ def get_id(*, label, configuration):
         An opaque string that uniquely identifies the target.
     """
     return "{} {}".format(label, configuration)
+
+def _longest_common_prefix(labels):
+    if not labels:
+        return None
+    # Assume the shortest string is the answer
+    res = labels[0]
+    for s in labels:
+        if len(s) < len(res):
+            res = s
+    # Compare one character at a time with the rest,
+    # as soon as any difference is found the result holds the longest common prefix
+    index = 0
+    for c in res.elems():
+        for label in labels:
+            if label[index] != c:
+                return res[:index]
+        index += 1
+
+    # If the above never returns the shortest string is the result
+    return res
+
+def calculate_replacement_label(*, id, replacement_labels):
+    """Calculates a single replacement label for a given id and list of replacement labels
+
+    It considers the list of labels received plus the label contained in the id itself, then it finds
+    the longest common prefix between those
+
+    Args:
+        id: An id that uniquely represents a target (see `get_id`)
+        replacement_labels: A list of replacement labels
+
+    Returns:
+        The longest common prefix between the list of labels and the label contained in the id,
+        if a longest common prefix is not found the label contained in the id is returned
+    """
+    id_label = id.split(" ")[0] # This assumes the id follows the creation pattern in `get_id`
+
+    res = _longest_common_prefix(
+        [id_label] + ["%s" % label for label in replacement_labels]
+    )
+    if not res:
+        return None
+
+    res = bazel_labels.normalize_label(res)
+    return Label(res)
 
 def write_target_ids_list(*, actions, name, target_dtos):
     """Writes the list of target IDs for a set of `xcode_target`s to a file.
--- a/xcodeproj/internal/target_id.bzl
+++ b/xcodeproj/internal/target_id.bzl
@@ -54,7 +54,7 @@ def calculate_replacement_label(*, id, replacement_labels):
     id_label = id.split(" ")[0]  # This assumes the id follows the creation pattern in `get_id`
 
     res = _longest_common_prefix(
-        [id_label] + ["%s" % label for label in replacement_labels],
+        [id_label] + [bazel_labels.normalize_label(label) for label in replacement_labels],
     )
     if not res:
         return None
--- a/xcodeproj/internal/xcodeproj_rule.bzl
+++ b/xcodeproj/internal/xcodeproj_rule.bzl
@@ -26,7 +26,7 @@ load(":platforms.bzl", "platforms")
 load(":project_options.bzl", "project_options_to_dto")
 load(":providers.bzl", "XcodeProjInfo")
 load(":resource_target.bzl", "process_resource_bundles")
-load(":target_id.bzl", "write_target_ids_list")
+load(":target_id.bzl", "write_target_ids_list", "calculate_replacement_label")
 load(":xcode_targets.bzl", "xcode_targets")
 
 # Utility
@@ -1485,12 +1485,27 @@ configurations: {}""".format(", ".join(xcode_configurations)))
     )
     focused_labels = {label: None for label in ctx.attr.focused_targets}
     unfocused_labels = {label: None for label in ctx.attr.unfocused_targets}
-    replacement_labels = {
-        r.id: r.label
-        for r in depset(
-            transitive = [info.replacement_labels for info in infos],
-        ).to_list()
+
+    replacement_labels_infos = depset(
+        transitive = [info.replacement_labels for info in infos],
+    ).to_list()
+
+    # Finds ids associated with multiple labels
+    multiple_replacement_labels = {
+        r.id: [x.label for x in replacement_labels_infos if x.id == r.id] # all labels associated with this id
+        for r in replacement_labels_infos
+        if len([s.id for s in replacement_labels_infos if s.id == r.id]) > 1 # Check to only consider ids associated with multiple labels
     }
+    # Ensures the id <=> label relationship is 1-1
+    replacement_labels = {}
+    for r in replacement_labels_infos:
+        if r.id in multiple_replacement_labels:
+            replacement_labels[r.id] = calculate_replacement_label(
+                id = r.id,
+                replacement_labels = multiple_replacement_labels[r.id],
+            )
+        else:
+            replacement_labels[r.id] = r.label
 
     (
         targets,
--- a/xcodeproj/internal/xcodeproj_rule.bzl
+++ b/xcodeproj/internal/xcodeproj_rule.bzl
@@ -26,7 +26,7 @@ load(":platforms.bzl", "platforms")
 load(":project_options.bzl", "project_options_to_dto")
 load(":providers.bzl", "XcodeProjInfo")
 load(":resource_target.bzl", "process_resource_bundles")
-load(":target_id.bzl", "write_target_ids_list", "calculate_replacement_label")
+load(":target_id.bzl", "calculate_replacement_label", "write_target_ids_list")
 load(":xcode_targets.bzl", "xcode_targets")
 
 # Utility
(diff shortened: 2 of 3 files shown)


brentleyjones (approved):
Approving to run CI

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -71,6 +71,7 @@ _BUILD_TEST_RULES = {
     "macos_build_test": None,
     "tvos_build_test": None,
     "watchos_build_test": None,
+    "test_suite": None,
```
brentleyjones, line 74: nit: this should only be the `*_build_test` rules.

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -259,6 +260,15 @@ def _skip_target(
...
 
+    test_suite_replacement_labels = []
+    if ctx.rule.kind == "test_suite":
+        for info in valid_transitive_infos:
+            for xcode_target in info.xcode_targets.to_list():
+                if xcode_target.id.count(".__internal__.__test_bundle"):
+                    test_suite_replacement_labels.append(
+                        struct(id = xcode_target.id, label = target.label),
+                    )
+
```
thiagohmcruz, line 271: @brentleyjones curious if you see a better approach here. Without this the scheme is Xcode gets named after the latest label from the underlying runners: https://github.com/bazel-ios/rules_ios/blob/master/tests/macos/xcodeproj/BUILD.bazel#L55 (which was the `iPhone` one based on my testing)

But when using a test suite I'd say the expected behaviour is to see the name of the respective test suite in Xcode: https://github.com/bazel-ios/rules_ios/blob/master/tests/macos/xcodeproj/BUILD.bazel#LL41C13-L41C49

All replacements are happening on top of the same label for the internal `.__internal__.__test_bundle` target, so adding this as the last option below as `+ test_suite_replacement_labels` fixed it for me but maybe there's a better approach here

Also this fixed adding the `test_suite` rule as a custom scheme in `schemes`, otherwise if fails since it can't find the `test_suite` label in the transitive deps and finds only the `iPhone/iPad` ones, so It ends up hitting this: https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/main/tools/generator/src/Extensions/XcodeScheme%2BExtensions.swift#L27

xcodeproj/internal/xcodeprojinfo.bzl
thiagohmcruz: 👍 will put this in a separate constant

xcodeproj/internal/xcodeprojinfo.bzl
brentleyjones: When it's a test suite, I think we shouldn't set `depset(direct)` below in `replacement_labels`, since we want the labels to be left alone, right?

thiagohmcruz: >When it's a test suite, I think we shouldn't set depset(direct) below in replacement_labels, since we want the labels to be left alone, right?

Ok, let me look into that!

>I don't think we want the test suite to appear in Xcode, I think the various sub-tests should just be included.

Maybe I'm thinking about this backwards, I thought it made sense to have the test suite there and Xcode is the "runner" but now I think that missed the point of having different runners already configured in `BUILD` files. Ok, let me explore the approach above.

luispadron: > I don't think we want the test suite to appear in Xcode, I think the various sub-tests should just be included.

It could become an auto-generated scheme that builds & tests all the tests under it maybe? But right, I think the default would be that it simply includes whatever test targets are declared in the suite.

thiagohmcruz: Yep, having all underlying tests in Xcode seems to be the correct behaviour indeed, discussing possible approaches offline to not pollute this thread too much but the TL;DR is that both underlying tests depend on the same `.__internal__.__test_bundle` test bundle target which is causing the latest one in `replacement_labels` to override all the others and we end up with a single test target in Xcode

thiagohmcruz: >It could become an auto-generated scheme that builds & tests all the tests under it maybe

This seems wroth exploring, haven't looked into auto-generated schemes at all but will do

thiagohmcruz: Ok, after some discussion I'll remove this and go with a different approach, will detailed it better in the PR descr once it's ready for review

xcodeproj/internal/target_id.bzl
```diff
@@ -12,6 +14,53 @@ def get_id(*, label, configuration):
...
+        The longest common prefix between the list of labels and the label contained in the id,
+        if a longest common prefix is not found the label contained in the id is returned
+    """
+    id_label = id.split(" ")[0]  # This assumes the id follows the creation pattern in `get_id`
+
+    res = _longest_common_prefix(
+        [id_label] + ["%s" % label for label in replacement_labels],
+    )
+    if not res:
+        return None
```
thiagohmcruz, line 59: Should probably `fail` here since ideally this would always return something?

xcodeproj/internal/target_id.bzl
```diff
@@ -12,6 +14,53 @@ def get_id(*, label, configuration):
...
+        for label in labels:
+            if label[index] != c:
+                return res[:index]
+        index += 1
+
+    # If the above never returns the shortest string is the result
+    return res
+
+def calculate_replacement_label(*, id, replacement_labels):
+    """Calculates a single replacement label for a given id and list of replacement labels
```
brentleyjones, line 40:
```suggestion
    """Calculates a single replacement label for a given id and list of \
    replacement labels
```

brentleyjones, line 40:
```suggestion
    """Calculates a single replacement label for a given id and list of \
    replacement labels.
```

xcodeproj/internal/target_id.bzl
```diff
@@ -12,6 +14,53 @@ def get_id(*, label, configuration):
...
+
+    It considers the list of labels received plus the label contained in the id itself, then it finds
+    the longest common prefix between those
```
brentleyjones, line 43:
```suggestion
    It considers the list of labels received plus the label contained in the id
    itself, then it finds the longest common prefix between those.
```

brentleyjones, line 43:
```suggestion
    It considers the list of labels received plus the label contained in the id
    itself, then it finds the longest common prefix between those.
```

xcodeproj/internal/target_id.bzl
```diff
@@ -12,6 +14,53 @@ def get_id(*, label, configuration):
...
+
+    Args:
+        id: An id that uniquely represents a target (see `get_id`)
+        replacement_labels: A list of replacement labels
+
+    Returns:
+        The longest common prefix between the list of labels and the label contained in the id,
+        if a longest common prefix is not found the label contained in the id is returned
+    """
+    id_label = id.split(" ")[0]  # This assumes the id follows the creation pattern in `get_id`
```
brentleyjones, line 53: Why do you need to do this part? Can't `_longest_common_prefix` just be called with `replacement_labels`?

Related: the format of a target id is supposed to be opaque, and will probably change in the future to be a  non-identifying hash.

xcodeproj/internal/target_id.bzl
```diff
@@ -12,6 +14,53 @@ def get_id(*, label, configuration):
...
+
+    res = _longest_common_prefix(
+        [id_label] + ["%s" % label for label in replacement_labels],
```
brentleyjones, line 56:
```suggestion
        [id_label] + [str(label) for label in replacement_labels],
```

brentleyjones, line 56:
```suggestion
        [id_label] + [bazel_labels.normalize_label(label) for label in replacement_labels],
```

xcodeproj/internal/target_id.bzl
```diff
@@ -12,6 +14,53 @@ def get_id(*, label, configuration):
...
+    )
+    if not res:
+        return None
+
+    res = bazel_labels.normalize_label(res)
```
brentleyjones, line 61: No need to normalize here, as you should do it above (since `bazel_labels.normalize_label` basically does `str()` already, with a little more magic).

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -1485,13 +1485,29 @@ configurations: {}""".format(", ".join(xcode_configurations)))
...
+    # Ensures the id <=> label relationship is 1-1
+    replacement_labels = {}
+    for r in replacement_labels_infos:
+        if r.id in multiple_replacement_labels:
+            replacement_labels[r.id] = calculate_replacement_label(
+                id = r.id,
+                replacement_labels = multiple_replacement_labels[r.id],
+            )
+        else:
+            replacement_labels[r.id] = r.label
```
brentleyjones, line 1509: I think it can be expressed more efficiently this way:
```suggestion
    raw_replacement_labels = {}
    for r in replacement_labels_infos:
        raw_replacement_labels.setdefault(r.id, []).append(r.label)

    replacement_labels = {}
    for id, labels in raw_replacement_labels.items():
        if len(labels) > 1:
            replacement_labels[id] = calculate_replacement_label(
                 id = id,
                 replacement_labels = labels,
             )
        else:
            replacement_labels[id] = labels[0]
```

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -71,6 +71,11 @@ _BUILD_TEST_RULES = {
     "macos_build_test": None,
     "tvos_build_test": None,
     "watchos_build_test": None,
+    "test_suite": None,
```
brentleyjones, line 74: Still needs to be removed.

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -88,6 +93,8 @@ def _should_skip_target(*, ctx, target):
     """
     if ctx.rule.kind in _BUILD_TEST_RULES:
         return True
+    if ctx.rule.kind in _TEST_SUITE_RULES:
+        return True
```
brentleyjones, line 97: I think this, and the replacement labels stuff, can be two separate PRs.

- Replacement labels: can add an example to `examples/integrations` for an `ios_unit_test` with multiple runners (or just adjust one of existing ones to have multiple runners).
- `test_suite`: Can add a `test_suite` around two tests in `examples/integration` and list it in `top_level_targets`

xcodeproj/internal/target_id.bzl
thiagohmcruz: good catch, we don't need to pass only the label here, the common prefix will be the same anyways 🤦 

>the format of a target id is supposed to be opaque, and will probably change in the future to be a non-identifying hash.
Ok good to know, I wasn't sure if we wanted to assume anything about that format or not, it makes sense to ensure it's opaque though, will update this code!

xcodeproj/internal/xcodeprojinfo.bzl
thiagohmcruz: Oops

xcodeproj/internal/xcodeprojinfo.bzl
thiagohmcruz: Ok, after we reverted the approach of bringing all runner variations into Xcode I thought this could go back to being a single change but I could be missing something:

>Replacement labels: can add an example to examples/integrations for an ios_unit_test with multiple runners (or just adjust one of existing ones to have multiple runners).

I think `ios_unit_test` [only takes a single runner](https://github.com/bazelbuild/rules_apple/blob/master/doc/rules-ios.md#ios_unit_test)? 

>test_suite: Can add a test_suite around two tests in examples/integration and list it in top_level_targets

That would be adding `ios_unit_test_suite` and `ios_ui_test_suite` with `runners` set and then `rules_ios`'s `ios_unit_test` with `runners` set (`rules_ios` call the above two under the hood). IIUC this already covers what you intended to cover in the previous item making it unnecessary?

brentleyjones: Got confused. Can land `test_suites` support first (just raw `test_suite`, not `ios_unit_test_suite`), then this change with the `replacement_labels` fix can come after to add support for `ios_unit_test_suite`/`rules_ios`'s `ios_unit_test.runners`.

thiagohmcruz: Ok thx for clarifying, I had only the `ios_*` rules in mind, didn't realize we could (and should) test `test_suite` rules explicitly first. Decoupling things like this makes sense from the perspective of `rules_xcodeproj` regardless how each repo abstracts these, thx for clarifying.

Will split this into the two PRs then!

luispadron: +1 to supporting test_suite. On Cash we use the native rule along with multiple ios_unit_test and `target_compatible_with` to split up our snapshot tests by device. 

Each unit test target specifies a runner and they're all part of the same suite. Ideally, this would show up as a single target in Xcode where the runner is whatever Xcode has selected. 

Right now I see all 4 targets get loaded and running it spins up the entire test suite, which spins up 4 sims + 1 sim from Xcode

thiagohmcruz: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2184

Just one Q there about custom schemes

xcodeproj/internal/target_id.bzl
thiagohmcruz: When passing the `id` the results works if the labels are all in the same format, suggesting a fix for that here first: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2188

This way I can compare the labels with the label in the id on the same basis to find the common prefix

thiagohmcruz:
All the discussion and feedback here was very helpful, thx @brentleyjones for the offline suggestions/guidance too!

Closing in favour of: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2196

PR #1969, opened 2023-04-05
Add `xcodeproj.bazel_path_env` attribute

brentleyjones:
This allows overriding what the `PATH` should be when invoking the `bazel_path` binary.


yongjincho92 (approved):

mattrobmattrob (approved):
A lot of duplication of `/usr/bin:/bin` in the various places it's used as a default but 👍

brentleyjones:
> A lot of duplication of /usr/bin:/bin in the various places it's used as a default

Same for all of the things that we default, since we want to no send it over in the spec if it’s the default. I’m cool with those being extracted into constants somewhere, but there are so few locations, and the various tests would break if we forgot to modify it somewhere, so I didn’t see the need to change it right now.

luispadron:
Testing this out now and wondering if this is how we expect it to be setup:

```python
    xcodeproj(
        name = name,
        ...
        bazel_path = "bazelisk",
        bazel_path_env = "/usr/bin:/bin:/opt/homebrew/bin/bazelisk:/usr/local/bin/bazelisk",
    )
```

or 

```python
    xcodeproj(
        name = name,
        ...
        bazel_path = "/opt/homebrew/bin/bazelisk",
    )
```

I noticed that this was the only setup that worked after this change, as in, invoking the wrapper and running Bazel. Without it our wrapper was failing because `BAZEL_REAL` wasn't set which indicates it wasnt running through Bazelisk.

I tried where `bazel_path = "bazel"` and `bazel_path_env = "/usr/bin:/bin:/opt/homebrew/bin/bazel:/usr/local/bin/bazel"` but that seemed to skip running through Bazelisk (or at least setting up the environment vars it sets up)

Using `bazel_path=./Tools/bazel` also seems to not set the `BAZEL_REAL` environment either.

brentleyjones:
I think this is an issue with your wrapper? But if you can create a small repro for the issue (not using this new attribute if possible per your latest edit), I can look into it.

brentleyjones:
I see what you are saying about `BAZEL_REAL` not being set. I didn't think that would be needed in the runner, but I probably should set it if it is. I'll get a PR up for that soon.

brentleyjones:
And related, maybe I expand this change to be a dictionary, to allow setting more than just PATH, since it's the only way to impact the Xcode based compilation. That way I can fixed a long standing `ANDROID_HOME` bug as well.

luispadron:
Yeah totally an issue in our wrapper which in trying to fix but we need that env var to run the correct binary. 

I don't know how other folks are doing this and maybe using BAZEL_REAL is incorrect? I do know what we had before in the wrapper (calling bazelisk) was causing the wrapper to execute twice. 

If we expose that env var then this should be good! I like the idea of a dict here as well.

brentleyjones:
In the wrapper you should use `BAZEL_REAL` in your final `exec` at the end, so if you are running into an issue with that not being set, that is a rules_xcodeproj problem. I'll get that fixed up soon.

brentleyjones:
So, looking into this, I think things are working as expected. If you don't have `BAZEL_REAL` set, you need to make sure your wrapper is passing it through (it wasn't in the repro I have available).

- In `runner.sh`, we don't filter the env, just change `PATH`, so if `BAZEL_REAL` was set by `bazelisk`, it will still be set when we run `bazel run //:generator`.
- In `runner.sh`, we resolve `bazel_path` with `which` (that takes into account runner.sh’s `PATH`, _not_ `bazel_path_env`). This means if you have `bazel_path = "bazel"`, it should find any globally installed `bazel`/`bazelisk`. Then later when it runs `bazel run //:generator` the `PATH` is filtered to `bazel_path_env` (by default `/usr/bin:/bin`). In your wrapper you were trying to find bazelisk, but you should instead use `BAZEL_REAL` if it’s set (which it should be), or you need to adjust `bazel_path_env` to be able to find what you want to find.
- In `bazel_build.sh` (in Xcode), we set the same `BAZEL_REAL` that was available when running `runner.sh`, and we use the same resolved path for `bazel_path`.

I confirm that global bazel/bazelisk, with or without a `tools/wrapper`, with `bazel_path = "bazel"` or `bazel_path = "bazelisk"`, worked. But I also improved the situation slightly (less `PATH` modification) [here](https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/1988).

I still plan to change `bazel_path_env` to `bazel_env`, which would be a dictionary of env variables you wish to modify or inherit from `runner.sh`’s env.

PR #1972, opened 2023-04-05
Stop adding folder-nested generated files to `extra_files`

brentleyjones:
`generated.xcfilelist` only needs to include files that are in the navigator, and actually only those that are referenced in `PBXBuildFile`s.

This also removes the need for `file_path.include_in_navigator`.

### DIFF
--- a/xcodeproj/internal/resources.bzl
+++ b/xcodeproj/internal/resources.bzl
@@ -95,8 +95,7 @@ def _add_structured_resources_to_bundle(
         *,
         nested_path,
         files,
-        generated,
-        extra_files):
+        generated):
     if nested_path:
         inner_dir = nested_path.split("/")[0]
     else:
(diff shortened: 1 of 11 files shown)


xcodeproj/internal/resources.bzl
```diff
@@ -131,9 +131,6 @@ def _add_structured_resources_to_bundle(
         )
         bundle.resources.append(fp)
 
-        if not file.is_source:
-            extra_files.append(file_path(file, include_in_navigator = False))
```
thiagohmcruz, line 135: Linter saying `extra_files` is not being used anymore, you'll probably have to remove from call sites of `_add_structured_resources_to_bundle` too

thiagohmcruz (approved):
All usages of `include_in_navigator` seem to be correctly removed here so LGTM!

PR #1989, opened 2023-04-07
Use `/bin/bash` in pre/post scripts

brentleyjones:


luispadron (approved):

yongjincho92 (approved):
Looks fine, but was it a different shell other than bash that it was using? (zsh?)

luispadron:
> Looks fine, but was it a different shell other than bash that it was using? (zsh?)
> 
> 

I think this is to make it match Xcode defaults when creating a new run script but could be wrong

brentleyjones:
It was using `/bin/sh` before. This change makes it so you don't have to wrap more complicated things in another script.

brentleyjones:
No one has asked for it, and my example isn't needed anymore, so I could just drop this. Or I could make it an option.

thii (approved):

mattrobmattrob (approved):

PR #1995, opened 2023-04-10
Allow {un,}focusing of non-Xcode targets

brentleyjones:
For example, before this change we couldn’t focus `@com_github_krzyzanowskim_cryptoswift//:CryptoSwift` in `examples/integration`.

### DIFF
--- a/xcodeproj/internal/xcodeproj_rule.bzl
+++ b/xcodeproj/internal/xcodeproj_rule.bzl
@@ -345,17 +345,17 @@ def _process_targets(
         for id, label in replacement_labels.items()
     }
 
-    target_labels = {
+    xcode_target_labels = {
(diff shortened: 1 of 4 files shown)


xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -345,23 +345,30 @@ def _process_targets(
         for id, label in replacement_labels.items()
     }
 
-    labels = {
+    target_labels = {
```
mattrobmattrob, line 348: What do you think about prefixing these two with `xcode_` to be abundantly clear compared to `label_strs`?

mattrobmattrob (approved):

mattrobmattrob (approved):

PR #2000, opened 2023-04-10
Add support for multiple projects per workspace

brentleyjones:
Before this change, if you generated project A, generated project B, then interacted with project A, it would use project B’s generated `BUILD` file.


xcodeproj/internal/xcodeproj_runner.bzl
```diff
@@ -366,8 +369,9 @@ def_env+='}}'""".format(
 
     is_bazel_6 = hasattr(apple_common, "link_multi_arch_static_library")
 
-    # TODO: Create a unique generator per target
-    generator_label = "{repo}//generator".format(
+    generator_package_name = paths.join("generator", package, name)
```
brentleyjones, line 372: I don't like that we have the name of the target as an extra package, but with all the associated generated files, and trying to have a single package host them all, I don't think there is an easy alternative right now.

BalestraPatrick (approved):

PR #2001, opened 2023-04-10
Skip unnecessary partition call in normalized_file_path

thiagohmcruz:
Ran `--starlark_cpu_profile` in one of our apps and noticed a small optimization we can do. This skips early if `extension not in path` to avoid an unnecessary `.partition()` call.

Before:
image
image

After:
image
image

I noticed that `file_path` has increased from `48.99s` to `58.45s` after this change but I couldn't find evidence for any change in behaviour here. I sent to a file all paths hitting both `return` statements in this code and I couldn't find any delta.

Still, I believe since things improve at the `normalized_file_path` level we can always come back and improve `file_path` in a separate PR. Open to suggestions here though!

### DIFF
--- a/xcodeproj/internal/files.bzl
+++ b/xcodeproj/internal/files.bzl
@@ -173,9 +173,9 @@ def normalized_file_path(file, *, folder_type_extensions):
     path = file.path
 
     for extension in folder_type_extensions:
-        prefix, ext, _ = path.partition(extension)
-        if not ext:
+        if extension not in path:
             continue
+        prefix, ext, _ = path.partition(extension)
         return file_path(
             file,
             path = prefix + ext,


thiagohmcruz:
`DCO` is failing, probably because I manually edited it to remove my company's email signature and use my personal. I'll re-commit+sign then, what are folks using here? Any preference for personal vs company signatures?

xcodeproj/internal/files.bzl
```diff
@@ -173,9 +173,9 @@ def normalized_file_path(file, *, folder_type_extensions):
     path = file.path
 
     for extension in folder_type_extensions:
-        prefix, ext, _ = path.partition(extension)
-        if not ext:
+        if extension not in path:
             continue
```
mattrobmattrob, line 177: Is `endswith` slower because of the Starlark method lookup cost?
```suggestion
        if not path.endswith(extension):
            continue
```

thiagohmcruz: I'd have to re-run the profile to make sure this doesn't slow things down, but given that it has better readability def worth re-running it. Will do and report back here.

thiagohmcruz: I tried `find` too as a suggestion from @brentleyjones but it was `1s` slower than `partition` so not sure yet, will re-run and see how `.endswith` goes

thiagohmcruz: That's why `partition` was being used to extract `prefix`, `ext` and leave out the rest

mattrobmattrob: Ah, perfect. Didn't look at the implementation of partition here but figured it'd be at the end. Ignore my suggestion then!

mattrobmattrob (approved):

brentleyjones (approved):

erikkerber:
Nice!

PR #2007, opened 2023-04-10
[DO NOT REVIEW] changes for cflag params

yongjincho92:
Draft change that passes copts as params file instead to avoid Argument List too long error in Index Builds

**high level changes list** 
1. in `xcode_targets.bzl#_xcode_target_to_dto()`, create params file for `copts`, `cxxopts`, `swiftcopts` 
2. in `xcodeproj_rule.bzl#_process_targets()`, for each processed `xcode_target`, collect `opt_files` and add the collected `opt_files` into `additional_generated` with a output group name `bg` (can I use the same group name for xcode build mode or should I create a new one for xcode different from existing ones?) 
3. in `xcodeproj_rule.bzl#_xcodeproj_impl()`, use `aditional_generated` to pass the opt params files into OutputGroupInfo provider for both Bazel and Xcode mode.

### DIFF
--- a/tools/generator/src/Generator/AddTargets.swift
+++ b/tools/generator/src/Generator/AddTargets.swift
@@ -100,6 +100,10 @@ Product for target "\(key)" not found in `products`
                 hasClangSearchPaths: target.hasClangSearchPaths,
                 files: files
             ),
+            try createCompileDependenciesScript(
+                in: pbxProj,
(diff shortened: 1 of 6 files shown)


yongjincho92:
closing in favor of https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2016

PR #2008, opened 2023-04-10
Remove `file_path` and `FilePath.PathType`

brentleyjones:
Creating `file_path` structs wasn't free, and in the end we don't need to pre-determine the type before `generator` processes them.

`createFilesAndGroups` has gotten pretty ugly after this. I have plans to clean it up a lot when I change the algorithm to support Swift Concurrency.


brentleyjones:
Looks to be a slight net win in when doing a clean (i.e. non-analysis cached) project generation. I like this change, but I'll wait to merge it until I'm ready with the `createFilesAndGroups` change which will make this a much better win.

brentleyjones:
Since this has a big memory win as well, and we cut the 1.4.0 release, I'll merge this now (after approvals), and dedicate to getting the `createFilesAndGroups` change in before 1.5.0.

thiagohmcruz (approved):
Ran some tests with this and there's def a good win here from a perf perspective for us, +1 to landing since the release was cut

mattrobmattrob (approved):

PR #2019, opened 2023-04-12
Fix command-line API to work with the `dump` and `shutdown` commands

brentleyjones:


ivan-golub (approved):

xcodeproj/internal/runner.template.sh
```diff
@@ -209,6 +210,11 @@ else
       "%generator_label%"
     )
   else
+    if [[ $cmd == "dump" ]]; then
```
ivan-golub, line 213: shutdown suffers from same issues

PR #2029, opened 2023-04-13
Remove `platform_info.is_same_type` check

brentleyjones:
This means that in BwB mode, for iOS schemes that have watchOS framework dependencies, and `adjust_schemes_for_swiftui_previews = True` (the default), then those schemes will include those watchOS framework targets. This has minimal impact on build time (just some extra rsync for their outputs per-build), but it reduces critical-path analysis phase time.


brentleyjones:
Closing in favor of #2030.

PR #2034, opened 2023-04-13
Process `swift_debug_settings.py` at execution time

brentleyjones:
This reduces the critical path in analysis by not expanding depsets and processing clang opts each time. Downside is approx. double the filesystem usage, and some front loaded I/O. Maybe we can recoup that later by splitting up the file and pulling them in on demand?


brentleyjones:
(I'm waiting for some cpu and memory profiles to see how this behaves on large projects.)

brentleyjones:
I think I can be more efficient with this, making it a single processor instead of being two parts. I'll iterate on it a bit more.

brentleyjones:
So this is current way less efficient. I can look at it more later.

brentleyjones:
Fixed the main issue, afaik.

brentleyjones:
Verified this works as expected. Ready for review @MobileNativeFoundation/rules_xcodeproj-maintainers.

thiagohmcruz (approved):

tools/params_processors/swift_debug_settings_processor.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/python3
```
thii, line 1: Not particular to this PR but I think the shebang is not needed, since the script is not invoked directly, and this file doesn't need to be executable.

brentleyjones: true, technically not needed, but I was testing it by running it directly so I wanted it.

erikkerber (approved):

PR #2036, opened 2023-04-14
Fix `create_lldbinit.sh` not appending content to a new line

### LINKED ISSUE #2035: Bug: Debugging and breakpoints broken when lldbinit exists
### Description

When a user already has an `~/.lldbinit` file (or equivalent) the generator will try to append a source line [here](https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/4a009a40bcbc2859bd1e937211165602258a6497/xcodeproj/internal/bazel_integration_files/create_lldbinit.sh#L83-L87) to setup lldb debugging / breakpoint support.

If this file previously had some content, especially without a final newline, the generator will append the source in the same line as the existing content. This causes the file to be invalid and thus breaks debugging and breakpoints in Xcode.

### Reproduction steps

1. Have an `~/.lldbinit` file with some content, for example
  
  ```sh
  command script import "/<user>/Application Support/Kaleidoscope/lldb_ksdiff.py"
  command script import /opt/homebrew/opt/chisel/libexec/fbchisellldb.pycommand
  ```

2. Generate a project
3. Notice the modified file is invalid:

  ```sh
  command script import "/<user>/Application Support/Kaleidoscope/lldb_ksdiff.py"
  command script import /opt/homebrew/opt/chisel/libexec/fbchisellldb.pycommand source ~/.lldbinit-rules_xcodeproj
  ```

### Expected behavior

The generator appends the source line to the end of the file, taking into account existing content.

### rules_xcodeproj version

1.4.0

### Xcode version

14.2.0

### Bazel version

6.1.1

### rules_apple version

2.2.0

### rules_swift version

1.7.1

### Additional information

_No response_

luispadron:
Closes #2035.

Updates the `grep` command to use `^` and `$` to match the entire line, if no match is found it adds the command to the end of the file.

Additionally, adss an extra step before we append the lldbinit content to ensure the file contains a new line.

```sh
tail -c 1 "$lldbinit" | read || echo >> "$lldbinit"
```

This will read last char of lldbinit but only runs echo to add a new line iff the read fails (i.e. EOF found before new line)

### Examples

```sh
# Before (without newline)
cat ~/.lldbinit                                                                                                 
command echo "hello world"%

# After running target in Xcode
command echo "hello world"
command source ~/.lldbinit-rules_xcodeproj
```

```sh
# Before (with newline)
cat ~/.lldbinit                                                                                                 
command echo "hello world"

# After running target in xcode
command echo "hello world"
command source ~/.lldbinit-rules_xcodeproj
```

```sh
# Before (corrupted file)
cat ~/.lldbinit                                                                                                 
command echo "hello world"command source ~/.lldbinit-rules_xcodeproj

# After running target in xcode
command echo "hello world"command source ~/.lldbinit-rules_xcodeproj
command source ~/.lldbinit-rules_xcodeproj

```

### DIFF
--- a/xcodeproj/internal/bazel_integration_files/create_lldbinit.sh
+++ b/xcodeproj/internal/bazel_integration_files/create_lldbinit.sh
@@ -81,7 +81,9 @@ fi
 touch "$lldbinit"
 
 readonly required_source='command source ~/.lldbinit-rules_xcodeproj'
-if ! grep -m 1 -q "$required_source" "$lldbinit"; then
+if ! grep -m 1 -q "^$required_source$" "$lldbinit"; then
+  # Add a newline if the file doesn't end with one
+  tail -c 1 "$lldbinit" | read || echo >> "$lldbinit"
   # Update `$lldbinit to source `~/.lldbinit-rules_xcodeproj`
   echo "$required_source" >> "$lldbinit"
 fi


xcodeproj/internal/bazel_integration_files/create_lldbinit.sh
```diff
@@ -82,6 +82,8 @@ touch "$lldbinit"
 
 readonly required_source='command source ~/.lldbinit-rules_xcodeproj'
 if ! grep -m 1 -q "$required_source" "$lldbinit"; then
```
brentleyjones, line 84: Can you also adjust this, so it only matches when it's a whole line? Otherwise this fix won't fix already corrupt files out there.

brentleyjones, line 84: I think this will do it?
```suggestion
if ! grep -m 1 -q "^$required_source" "$lldbinit"; then
```

luispadron: That doesn't seem to work, it adds to the same line due to `echo` below:


```sh
cat ~/.lldbinit
command echo "hello world"%       

cat ~/.lldbinit                                                                                                 
command echo "hello world"command source ~/.lldbinit-rules_xcodeproj
```

luispadron: Yep make sense, let me get that updated

luispadron: Was missing the `$` at end.

This works: `"^$required_source$"`

luispadron:
@brentleyjones Should be functioning for all cases now, `grep` update was the only thing needed.

luispadron:
Actually never mind just tried it again for the (no newline case and it still appends to the same line)

brentleyjones (approved):
Thanks!

brentleyjones:
Ahh, makes sense, so needs both changes

brentleyjones (approved):

luispadron:
Looks good now, tested for all three cases. The addition of checking if newline exists and adding the newline is required for the case where there is no newline in the file originally.

The updated grep will fix any corrupted files that may have had this added to the same line

PR #2042, opened 2023-04-17
Rewrite `createFilesAndGroups`

brentleyjones:
It’s around 5 times faster now (e.g. on a particular large project it went from 5.61s to 1.13s).


tools/generator/src/Generator/CreateFilesAndGroups.swift
brentleyjones, line 1: Not really worth looking at the diff. I would look at the function as a whole. I'll go through now and add more comments, since I only started to add them towards the end once I knew what i was going to do.

CognitiveDisson (approved):
Works great 👍

mattrobmattrob (approved):

PR #2060, opened 2023-04-19
Use `action.argv` for CC targets as well

brentleyjones:
This will generally make our compiler flags match better with reality. This is also an analysis phase wall-time reduction. Though it’s mainly a change that will let us use `action.args` for copt processing in the future, reducing our memory footprint.

### DIFF
--- a/xcodeproj/internal/input_files.bzl
+++ b/xcodeproj/internal/input_files.bzl
@@ -58,12 +58,12 @@ def _process_cc_info_headers(headers, *, exclude_headers, generated):
 
 # API
 
-_C_EXTENSIONS = {
+C_EXTENSIONS = {
     "c": None,
     "m": None,
 }
 
-_CXX_EXTENSIONS = {
+CXX_EXTENSIONS = {
     "cc": None,
     "cpp": None,
     "cxx": None,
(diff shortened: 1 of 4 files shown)


brentleyjones:
@MobileNativeFoundation/rules_xcodeproj-maintainers I've checked the difference in the example `{c,cxx}.compile.params` files, but I would like you to compare on your end as well to make sure nothing odd is coming through now, which could break indexing.

brentleyjones:
Looks like if we take this change (which I plan on doing) we are dropping Bazel 5 support.

jpsim:
> Looks like if we take this change (which I plan on doing) we are dropping Bazel 5 support.

Is it worth doing one last release prior to dropping Bazel 5? 1.4.1?

brentleyjones:
Yeah, I won't merge this for a bit knowing it breaks compatibility. And we will release 1.4.1 or 1.5.0 before the drop.

thiagohmcruz:
> Is it worth doing one last release prior to dropping Bazel 5? 1.4.1?

+1 for doing this, we're still on Bazel 5, planning to move to 6 in Q2 so we'll have to stick to this last release until then

thiagohmcruz (approved):
Thank you for keeping the Bazel 5 code path alive for a little longer!

brentleyjones:
Merging so I can open up stacked PRs.

PR #2064, opened 2023-04-19
Fix `import_indexstores.sh` when using `swift.remap_xcode_path`

chiragramani:
Post rules_swift 1.6.0 release, "DEVELOPER_DIR" remapping was enabled by default.
I was debugging a different indexing issue and hence came across an observation where this is not remapped to local machine dev directory.
I also haven't seen any issue with rules_xcodeproj indexing so in case this is not needed, happy to abandon it.

Before this change:
[ContentView.o-3P4ZTLDW8HNC.zip](https://github.com/MobileNativeFoundation/rules_xcodeproj/files/11278087/ContentView.o-3P4ZTLDW8HNC.zip)

After this change:
[ContentView.o-3P4ZTLDW8HNC.zip](https://github.com/MobileNativeFoundation/rules_xcodeproj/files/11278090/ContentView.o-3P4ZTLDW8HNC.zip)

### DIFF
--- a/xcodeproj/internal/bazel_integration_files/import_indexstores.sh
+++ b/xcodeproj/internal/bazel_integration_files/import_indexstores.sh
@@ -99,9 +99,9 @@ remaps=(
   # The only other type of path in the unit files are sysroot based. While
   # these should always be Xcode.app relative, our regex supports command-line
   # tools based paths as well.
-  -remap "^(?:.*?/[^/]+/Contents/Developer|/Library/Developer/CommandLineTools).*?/SDKs/([^\\d.]+)=$DEVELOPER_DIR/Platforms/\$1.platform/Developer/SDKs/\$1"
-
-  -remap "DEVELOPER_DIR=$DEVELOPER_DIR"
+  # `DEVELOPER_DIR` has an optional `./` prefix, because index-import adds `./`
+  # to all relative paths.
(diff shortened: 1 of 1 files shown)


xcodeproj/internal/bazel_integration_files/import_indexstores.sh
```diff
@@ -100,6 +100,8 @@ remaps=(
   # these should always be Xcode.app relative, our regex supports command-line
   # tools based paths as well.
   -remap "^(?:.*?/[^/]+/Contents/Developer|/Library/Developer/CommandLineTools).*?/SDKs/([^\\d.]+)=$DEVELOPER_DIR/Platforms/\$1.platform/Developer/SDKs/\$1"
+
+  -remap "DEVELOPER_DIR=$DEVELOPER_DIR"
```
brentleyjones, line 104: How about this instead, so we only remap when `DEVELOPER_DIR` is at the start, and when it's for the sysroot?

```suggestion
  -remap "^(?:.*?/[^/]+/Contents/Developer|^DEVELOPER_DIR|/Library/Developer/CommandLineTools).*?/SDKs/([^\\d.]+)=$DEVELOPER_DIR/Platforms/\$1.platform/Developer/SDKs/\$1"
```

brentleyjones: I tested my suggestion and it works.

chiragramani: Thanks @brentleyjones 🙏🏼

brentleyjones (approved):

brentleyjones:
You’ll need to sign your commit for the DCO check.

thii (approved):

PR #2068, opened 2023-04-20
Stop storing output group names

brentleyjones:
Reducing memory usage for slightly higher cpu usage.


brentleyjones:
Since we store the same strings in `output_group_list`, I don't think this will actually be a win. Testing it.

brentleyjones:
Not doing this for now. #2073 is better for now.

PR #2069, opened 2023-04-20
Quote one more source of linkopts

brentleyjones:
Also rename functions and add comments to make it clear when we should use which.


mattrobmattrob (approved):
❤️ 🧹

luispadron (approved):

PR #2070, opened 2023-04-20
Escape spaces in clang_opts

brentleyjones:
This won’t fix if `$(PROJECT_DIR)` has spaces in it, but it shouldn’t since Bazel currently doesn’t support the execution root having spaces in the path.


jpsim (approved):

mattrobmattrob (approved):
👏

PR #2078, opened 2023-04-21
Actually fix `PBXBuildFile` sorting

brentleyjones:
Old way wouldn’t take into account parent of parents, because it would end up hitting the `PBXFileElement` cache. Solution is to move the `PBXBuildFile` cache into `PBXFileElement` and use a different property recursively.


mattrobmattrob (approved):
Was experiencing this one yesterday. 👍

luispadron (approved):

PR #2082, opened 2023-04-21
Pre-filter `transitive_infos`

brentleyjones:
I should have done this a long time ago, but there used to be some more logic around these lists. This is mainly a CPU win, but also a huge readability win.


luispadron (approved):
Nice!

PR #2085, opened 2023-04-21
Change `indexstores_filelists` calculation to not depend on target id

brentleyjones:
This is needed to support a future change where we change the target id to be an opaque string.


xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
```diff
@@ -344,19 +361,20 @@ def _main(
...
                 guid_labels,
+                guid_package_bins,
                 guid_target_ids,
             )
 
     print("\n".join(
         [
-            f"{label}\n{prefix} {id}"
-            for label, id in labels_and_target_ids
+            f"{label}\n{configuration}\n{prefix} {id}"
```
brentleyjones, line 376:
```suggestion
            f"{label}\n{packageBin}\n{prefix} {id}"
```

xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
```diff
@@ -344,19 +361,20 @@ def _main(
...
+            for label, configuration, id in return_data
```
brentleyjones, line 377:
```suggestion
            for label, packageBin, id in return_data
```

brentleyjones:
Not going with this for now.

PR #2089, opened 2023-04-24
Prevent calculation of `bwx_unfocused_libraries` in BwB mode

brentleyjones:
Memory and cpu win.


mattrobmattrob (approved):
This is to skip the `if not bwx_unfocused_libraries` case below this change, correct?

thiagohmcruz (approved):

PR #2099, opened 2023-04-24
Stop setting `ENABLE_BITCODE = "NO"`

brentleyjones:
The default is `"NO"`. Results in smaller projects and reduced Starlark memory usage.


jpsim (approved):
Do we even need to support enabling bitcode any more since Apple no longer accepts App Store Connect submissions with it?

mattrobmattrob (approved):

brentleyjones:
> Do we even need to support enabling bitcode any more since Apple no longer accepts App Store Connect submissions with it?

We technically support older Xcode versions still, and people can do ad-hoc distribution, so yes, until we drop those.

PR #2107, opened 2023-04-24
Set `*_PARAMS_FILE` build settings in generator

brentleyjones:
Since we need to have these values set anyway, and the way the paths are calculated is predetermined, this is better. Is a Starlark memory and CPU reduction.


tools/generator/src/Generator/SetTargetConfigurations.swift
```diff
@@ -254,7 +254,10 @@ Target with id "\(id)" not found in `consolidatedTarget.uniqueFiles`
     ) throws -> [String: BuildSetting] {
         var buildSettings = target.buildSettings
 
-        if target.linkParams != nil {
+        if let linkParams = target.linkParams {
+            buildSettings.set("LINK_PARAMS_FILE", to: #"""
+$(BAZEL_OUT)\#(linkParams.path.string.dropFirst(9))
+"""#)
```
mattrobmattrob, line 260: Mind adding a comment about what you're using `.dropFirst(9)` to accomplish?
```suggestion
            // Drop the `bazel-out` prefix since we use the env var for this portion of the path.
            buildSettings.set("LINK_PARAMS_FILE", to: #"""
$(BAZEL_OUT)\#(linkParams.path.string.dropFirst(9))
"""#)
```

mattrobmattrob (approved):

PR #2111, opened 2023-04-25
Stop setting `PRODUCT_MODULE_NAME` when not needed

brentleyjones:
Starlark memory reduction.

### DIFF
--- a/xcodeproj/internal/xcode_targets.bzl
+++ b/xcodeproj/internal/xcode_targets.bzl
@@ -285,6 +285,14 @@ def _merge_xcode_target(*, src, dest):
         build_settings,
     )
 
+    product = _merge_xcode_target_product(
+        src = src.product,
+        dest = dest.product,
+    )
+
+    if build_settings.get("PRODUCT_MODULE_NAME", None) == product.name:
+        build_settings.pop("PRODUCT_MODULE_NAME")
+
     return _make_xcode_target(
         id = dest.id,
         label = dest.label,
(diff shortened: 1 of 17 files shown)


xcodeproj/internal/xcode_targets.bzl
```diff
@@ -285,17 +285,22 @@ def _merge_xcode_target(*, src, dest):
         build_settings,
     )
 
+    product = _merge_xcode_target_product(
+        src = src.product,
+        dest = dest.product,
+    )
+
+    if build_settings.get("PRODUCT_MODULE_NAME", None) == product.name:
+        build_settings.pop("PRODUCT_MODULE_NAME")
```
mattrobmattrob, line 294: This is just choosing to rely on the Xcode default for this value, correct? Is it possible to move this _just_ into the generation of `src._build_settings` or `dest._build_settings` instead to just never set it if it's the default behavior (this might be what the other file changes actually are, not sure)?

mattrobmattrob (approved):

xcodeproj/internal/xcode_targets.bzl
brentleyjones: This accounts for target merging, where until we merge them together the product name wouldn't match the module name. And yes, this (whole PR) is relying on the fact that `PRODUCT_MODULE_NAME = $(PRODUCT_NAME:c99extidentifier)`. And we already don't set it at the library and top_level when they are equal, but this catches the case where target merging would have them be equal.

brentleyjones: Now that I think about that, we might run into an issue where because of target merging we _do_ need `PRODUCT_MODULE_NAME` set, because it is _now_ different. 🤔

brentleyjones:
This breaks stuff in my testing. So not pushing it for the time, but I'll possibly get back to it later.

PR #2127, opened 2023-04-26
Include generator name in params files

brentleyjones:
This is to prevent issues when generating two projects at the same time. This isn’t possible for normal users, but it is in analysis tests (which I’m trying to get some in).


xcodeproj/internal/opts.bzl
```diff
@@ -680,15 +683,18 @@ def _create_cc_compile_params(
 
     return params
 
-def _create_swift_compile_params(*, actions, name, opts):
+def _create_swift_compile_params(*, actions, name, opts, generator_name):
```
mattrobmattrob, line 686: [nit]
```suggestion
def _create_swift_compile_params(*, actions, generator_name, name, opts):
```

xcodeproj/internal/opts.bzl
```diff
@@ -659,8 +661,9 @@ def _create_cc_compile_params(
     ]
 
     params = actions.declare_file(
-        "{}.rules_xcodeproj.{}.compile.params".format(
+        "{}.rules_xcodeproj-{}.{}.compile.params".format(
```
mattrobmattrob, line 664: I do wonder if the pattern could be switch to increase clarity just a bit. Mostly leaving the static `rules_xcodeproj` value by itself.
```suggestion
        "{}-{}.rules_xcodeproj.{}.compile.params".format(
```

xcodeproj/internal/opts.bzl
```diff
@@ -641,8 +642,9 @@ def _create_cc_compile_params(
 
     def _create_compiler_sub_params(idx, sub_args):
         sub_output = actions.declare_file(
-            "{}.rules_xcodeproj.{}.compile.sub-{}.params".format(
+            "{}.rules_xcodeproj-{}.{}.compile.sub-{}.params".format(
```
mattrobmattrob, line 645: Same here:
```suggestion
            "{}-{}.rules_xcodeproj.{}.compile.sub-{}.params".format(
```

xcodeproj/internal/opts.bzl
```diff
@@ -680,15 +683,18 @@ def _create_cc_compile_params(
...
+def _create_swift_compile_params(*, actions, name, opts, generator_name):
     if not opts or not actions:
         return None
 
     args = actions.args()
     args.add_all(opts)
 
     output = actions.declare_file(
-        "{}.rules_xcodeproj.swift.compile.params".format(name),
+        "{}.rules_xcodeproj-{}.swift.compile.params".format(
```
mattrobmattrob, line 694: And here:
```suggestion
        "{}-{}.rules_xcodeproj.swift.compile.params".format(
```

mattrobmattrob (approved):
Don't let the file(s) name change block if you disagree. 👍

xcodeproj/internal/opts.bzl
brentleyjones: I think `YourTargetTame-GeneratorTargetName.rules_xcodeproj.more_stuff` is more confusing than `YourTargetName.rules_xcodeproj-GeneratorTargetName.more_stuff`. This way you can sort of ignore everything from `.rules_xcodeproj` on. How about I change the `-` to a `.` so it's `YourTargetName.rules_xcodeproj.GeneratorTargetName.more_stuff`?

mattrobmattrob: I like that too. 👍

PR #2128, opened 2023-04-26
Include generated generators in fixtures

brentleyjones:
Will be used by future analysis tests.


mattrobmattrob (approved):
What's the pathway for updating these `BUILD` in the case we change the interface of `xcodeproj` and we get analysis/otherwise failures? Should `test/update_all_fixtures.sh` clear them before running anything just to be careful?

brentleyjones:
> What's the pathway for updating these `BUILD` in the case we change the interface of `xcodeproj` and we get analysis/otherwise failures? Should `test/update_all_fixtures.sh` clear them before running anything just to be careful?

They aren't referenced by anything currently, so there should be no analysis errors. And then the only thing that will reference them is tests in `test/internal`. I don't think `test/update_all_fixtures.sh` needs to clear anything ahead of time, since it can run even with them being invalid. To prove that I went and made various syntax errors in the `BUILD` file and `test/update_all_fixtures.sh` updated them without issue.

mattrobmattrob:
> To prove that I went and made various syntax errors in the `BUILD` file and `test/update_all_fixtures.sh` updated them without issue.

👏 Appreciate you being thorough.

PR #2132, opened 2023-04-27
Rename `xcode_target._*_params` to `xcode_target.*_params`

mattrobmattrob:
Transition `_swift_params`, `_c_params`, & `_cxx_params` to non-underscore prefix since they're used external to `xcode_targets.bzl`.

### DIFF
--- a/xcodeproj/internal/xcodeproj_rule.bzl
+++ b/xcodeproj/internal/xcodeproj_rule.bzl
@@ -528,8 +528,8 @@ targets.
             # Only merge if one src is swift and the other isn't.
             src_1 = src_ids[0]
             src_2 = src_ids[1]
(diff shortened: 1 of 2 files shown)


xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -744,8 +744,8 @@ targets.
             compile_params.append(xcode_target._c_params)
         if xcode_target._cxx_params:
```
brentleyjones, line 745: Can you do the same for the c/cxx params as well, since they are referenced here?

mattrobmattrob: Oh, geeze, whoops. Good point.

brentleyjones (approved):

PR #2134, opened 2023-04-27
Pre-filter `transitive_infos` in more places

brentleyjones:
Missed in d2cf7913931d4d57dbcc887fca9e3ad1863449a6.

### DIFF
--- a/xcodeproj/internal/xcodeprojinfo.bzl
+++ b/xcodeproj/internal/xcodeprojinfo.bzl
@@ -337,7 +337,7 @@ def _skip_target(
                     automatic_target_info = automatic_target_info,
                 )
                 for info in deps_transitive_infos
-            ] if automatic_target_info.envs else None,
+            ] if automatic_target_info.env else None,
             transitive = [
                 info.envs
                 for info in valid_transitive_infos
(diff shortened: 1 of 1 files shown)


xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -263,12 +268,8 @@ def _skip_target(
...
                     automatic_target_info = automatic_target_info,
                 )
-                for attr, info in transitive_infos
-                if (target and
-                    attr in deps_attrs and
-                    info.xcode_target and
-                    automatic_target_info.args)
-            ],
+                for info in deps_transitive_infos
+            ] if automatic_target_info.args else None,
```
mattrobmattrob, line 272: Guess it doesn't change the behavior so it's not necessary/never `None`?
```suggestion
            ] if target and automatic_target_info.args else None,
```

mattrobmattrob (approved):

xcodeproj/internal/xcodeprojinfo.bzl
brentleyjones: There was a time when `target` could be `None`, but now it's called only from the aspect, so it's always non-`None`.

thiagohmcruz (approved):

PR #2144, opened 2023-04-28
fix relative path

Kila2:
Fix generate Xcode project contains soft link ref. Because Xcode edit a soft link and save file will get a error.

### DIFF
--- a/tools/generator/src/Generator/CreateFilesAndGroups.swift
+++ b/tools/generator/src/Generator/CreateFilesAndGroups.swift
@@ -234,13 +234,33 @@ extension Generator {
                 explicitFileType = nil
             }
 
-            let file = PBXFileReference(
-                sourceTree: sourceTree,
-                name: name,
-                explicitFileType: explicitFileType,
-                lastKnownFileType: lastKnownFileType,
-                path: path
-            )
+            var realPath:Path! = nil
+            if isExternal {
+                realPath = directories.absoluteExternal + Path(String(filePathStr.dropFirst("external/".count)))
+            }
+            else {
+                realPath = directories.workspace+Path(filePathStr)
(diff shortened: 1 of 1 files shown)


brentleyjones:
Do you have a repro? This should be handled in `resolveFilePath` already.

brentleyjones:
In my recent refactor I might have lowered the count by one too many, because my examples all still worked. Might be all that is needed.

brentleyjones:
I think I get it now. This is additional symlinks in the various places (external, bazel-out, or even the workspace).

This is expensive to do, so I need to do some benchmarks, and think on this.

Kila2:
Try use this example this project has many symlinks. when use Xcode edit and save will emit error.
https://github.com/MobileNativeFoundation/XCBBuildServiceProxyKit/tree/main/Examples

brentleyjones:
Yeah, I thought of that after my initial comment. We will have to support this, either always or with a flag. I'm looking into it. Thanks for the bug report!

brentleyjones:
@Kila2 can you verify that #2147 fixes this for you?

Kila2:
> @Kila2 can you verify that #2147 fixes this for you?

I will test it tomorrow

Kila2:
I'm try use this project and edit PreviewInfoResponse.swift. The https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2147 can't fix it.
https://github.com/Kila2/XCBBuildServiceProxyKit/tree/main/Examples
image

brentleyjones:
Odd. I’ll try that example early next week and tweak the fix.

brentleyjones:
I wasn't handling when a symlink points to another symlink. Tested and it works now.

PR #2145, opened 2023-04-28
Fix `pop` error with multiple merge destinations

brentleyjones:
If a library target can merge into multiple top-level targets, then the second+ destination will fail `pop()` if we don’t have a default specified.


brentleyjones:
Ehh, my method is wrong. I'll leave it to @mattrobmattrob to fix.

PR #2147, opened 2023-04-28
Resolve symlinks for all paths

brentleyjones:
Closes #2144.

This fixes issues with symlinks that occur in other places outside of Bazel’s control, like the workspace, generated files, etc.

For a certain large project, this doubles the time that `createFilesAndGroups()` takes (1s to 2s), resulting in total `generator` time going from 8s to 9s.

This also changes symlink resolution to loop, in case a symlink points to another symlink.

### DIFF
--- a/tools/generator/src/Generator/CreateFilesAndGroups.swift
+++ b/tools/generator/src/Generator/CreateFilesAndGroups.swift
@@ -94,7 +94,7 @@ extension Generator {
             path: String
         ) {
             let relativePath: Path
-            let absolutePath: Path
+            var absolutePath: Path
             let addToResolvedRepositories: Bool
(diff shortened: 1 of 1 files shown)


brentleyjones:
@MobileNativeFoundation/rules_xcodeproj-maintainers thoughts on the slight performance regression? Maybe y'all can test a before and after and see if it's in line with my testing?

thiagohmcruz (approved):

thiagohmcruz:
If using a large project the perf impact is only 1s I wouldn't be too worried about it given the benefits 

But will run one test in one of our big apps to compare just in case

luispadron (approved):

BalestraPatrick (approved):

PR #2153, opened 2023-05-03
Rename `platform_info` to `platforms`

brentleyjones:
Better matches naming of our other Starlark modules.


xcodeproj/internal/xcode_targets.bzl
```diff
@@ -1,1024 +0,0 @@
-"""Module containing functions dealing with the `Target` DTO."""
```
thiagohmcruz, line 1: Is the intent to delete this file? I don't see this logic being moved somewhere else?

brentleyjones: No, something odd with my git-pile implementation. Fixing.

thiagohmcruz (approved):

mattrobmattrob (approved):

adincebic (approved):

PR #2155, opened 2023-05-04
Add `PBXProject` prefix `PBXProj` partial generator

brentleyjones:


brentleyjones:
@MobileNativeFoundation/rules_xcodeproj-maintainers I might not merge this for a bit, but I want us to have discussions about it now.

This is the first of the `.pbxproj` partial generators. They will be used in a new incremental project generation. It is the smallest, least complicated, etc. _But_ it will be used as a blueprint to copy for all of the other generators. So I want to use this PR as a place to discuss _everything_ about how the generator is coded, from its use of `ArgumentParser`, to how the tests are set up, to how things are documented, etc. Nothing is off limits, and now is the time to nit to your hearts content.

tools/pbxproj_generators/BUILD
```diff
@@ -0,0 +1,28 @@
+load("//xcodeproj:defs.bzl", "xcodeproj")
+load(
+    ":xcodeproj_targets.bzl",
+    "SCHEME_AUTOGENERATION_MODE",
+    "TOP_LEVEL_TARGETS",
+    "XCODE_CONFIGURATIONS",
+    "get_xcode_schemes",
+)
```
brentleyjones, line 8: I was using this pattern still, because I thought I would add fixtures for these... but I don't think we will be using fixtures by the time all of this is done. So I'll probably move those values into this file.

tools/pbxproj_generators/lib/PBXProj/src/Identifiers.swift
```diff
@@ -0,0 +1,7 @@
+public enum Identifiers {
```
brentleyjones, line 1: I need to add a doc comment to this, because it's not clear why this is needed at first glance.

tools/pbxproj_generators/lib/PBXProj/src/Logger.swift
brentleyjones, line 1: Copied (with formatting changes) from `generator`. I'll back-port the formatting changes.

tools/pbxproj_generators/lib/PBXProj/src/SemanticVersion.swift
brentleyjones, line 1: Copied (with formatting changes) from `generator`. I'll back-port the formatting changes.

tools/pbxproj_generators/project_prefix/src/Generator/Generator.swift
```diff
@@ -0,0 +1,36 @@
...
+/// partials. The `generate()` method is passed all the inputs needed to
+/// generate a partial.
+struct Generator {
+    private let environment: Environment
+
+    init(environment: Environment = .default) {
+        self.environment = environment
+    }
+
+    func generate(arguments: Arguments) throws {
```
brentleyjones, line 17: Does this need a doc comment?

tools/pbxproj_generators/project_prefix/test/CalculateProjectPrefixTests.swift
```diff
@@ -0,0 +1,179 @@
...
+        
+        let compatibilityVersion = "AppCode 42.7.4"
+        let developmentRegion = "en-GB"
+        let organizationName: String? = nil
+        let projectDir = "/some/execution_root"
+        let workspace = "/Users/TimApple/StarBoard"
+
+        let expectedProjectPrefix = #"""
+		000000000000000000000001 /* Project object */ = {
+			isa = PBXProject;
```
brentleyjones, line 20: This uses tabs for the indent, which is expected in the output. You can't easily see that though, and I bet it can be a pain for development. Currently I opted for here and in `Generator.calculateProjectPrefix()` to directly use them, since `\t` is hard to read, and developing an indent system seemed like a lot of overhead when we will always know the exact indent we want.

I'm open to feedback though. Maybe we just need a comment pointing out that it uses tabs and it's intentional?

tools/pbxproj_generators/project_prefix/src/Generator/Arguments.swift
```diff
@@ -0,0 +1,39 @@
+import ArgumentParser
+import Foundation
+import PBXProj
+
+extension Generator {
+    struct Arguments: ParsableArguments {
+        @Argument(
+            help: """
+Path to where the 'project_prefix' '.pbxproj' partial should be written.
```
brentleyjones, line 9: Everywhere I mention `.pbxproj` I can probably say `PBXProj` instead. A downside to that is besides for the filename it's never called that (unlike the other nested types like `PBXProject`). If people search though, they will find it called that in `XcodeProj` and the like.

```suggestion
Path to where the 'project_prefix' '.pbxproj' partial should be written.
```

tools/pbxproj_generators/lib/PBXProj/src/SemanticVersion.swift
```diff
@@ -0,0 +1,107 @@
...
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        try container.encode(full)
+    }
+}
+
+extension SemanticVersion: Decodable {
+    public init(from decoder: Decoder) throws {
+        let value = try decoder.singleValueContainer().decode(String.self)
+        guard let self0 = SemanticVersion(version: value) else {
```
CognitiveDisson, line 93: That variable name ( `self0` ) is a little confusing and doesn't make things clear to me.

CognitiveDisson: Maybe something like `instance` is a bit better

tools/pbxproj_generators/project_prefix/src/Generator/Generator.swift
```diff
@@ -0,0 +1,36 @@
...
+        try environment.writeProjectPrefix(
+            environment.calculateProjectPrefix(
+                /*compatibilityVersion:*/ environment
```
brentleyjones, line 20: Since we can't have labels for closures, I feel this adds to readability.

tools/pbxproj_generators/project_prefix/src/ProjectPrefix.swift
```diff
@@ -0,0 +1,33 @@
+import ArgumentParser
+import Foundation
+import PBXProj
+
+@main
+struct ProjectPrefix: ParsableCommand {
+    static var configuration = CommandConfiguration(
+        commandName: "project_prefix",
+        abstract: "Generates the 'project_prefix' '.pbxproj' partial."
```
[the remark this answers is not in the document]
brentleyjones: Now that these are under the `//tools/generators` package, the generators probably need a more descriptive name.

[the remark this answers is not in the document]
brentleyjones: Added a README that expands on the above. Still probably need better names.

tools/generators/pbxproject_prefix/README.md
brentleyjones, line 1: I'll fill this file in soon (Monday at this point). It will detail the inputs and show an example output.

tools/generators/README.md
```diff
@@ -0,0 +1,70 @@
...
+- `pbxproject_additional_elements`:
+  - The start of the `PBXProj` element
+  - All of the various `PBXProject` additional elements:
+    - `XCBuildConfiguration`
+    - `XCBuildConfigurationList`
+    - `PBXTargetDependency`
+    - `PBXContainerItemProxy`
+- [`pbxproject_prefix`](pbxproject_prefix/README.md):
+  - The start of `PBXProject` element
+  - Contains all `PBXProject` properties except for
```
thiagohmcruz, line 39: Nit: It was a bit confusing to see "all X except for Y" for something called "prefix", I'd expect a "prefix" to generate a piece (the start) of the thing and not almost "all" of it. This is explicitly stated in `pbxproject_prefix/README.md` and IMO a similar wording could be used here

tools/generators/pbxproject_prefix/README.md
```diff
@@ -0,0 +1,52 @@
...
+- Positional `minimum-xcode-version`
+- Option `--development-region <development-region>`
+- Option `--organization-name <organization-name>`
+- Flag `--colorize`
+
+Here is an example invocation:
+
+```shell
+$ pbxproject_prefix \
+    /tmp/pbxproj_partials/pbxproject_prefix \
```
thiagohmcruz, line 24: I've never used `ArgumentParser`, is it common there to use positional args like this and not specify something like `--output-path` instead? Or that is an option and we're choosing to not make that the default usage?

tools/generators/pbxproject_prefix/src/Generator/CalculatePBXProjectPrefix.swift
```diff
@@ -0,0 +1,40 @@
+import PBXProj
+
+extension Generator {
+    /// Calculates the `PBXProject` prefix `PBXProj` partial.
+    static func calculatePBXProjectPrefix(
```
thiagohmcruz, line 5: IIUC this is really "generating" the prefix partial and not "calculating" it correct?

tools/generators/pbxproject_prefix/src/Generator/CalculateProjectDir.swift
```diff
@@ -0,0 +1,14 @@
+import PBXProj
+
+extension Generator {
+    /// Calculates the `PBXProject.projectDir` string.
+    ///
+    /// - Parameters:
+    ///   - executionRoot: The absolute path to the Bazel execution root.
+    static func calculateProjectDir(executionRoot: String) -> String {
```
thiagohmcruz, line 8: See other comment on using the word "calculate", why not simply call it `projectDir`? No strong preference here btw, feel free to ignore!

tools/generators/pbxproject_prefix/src/Generator/Generator.swift
```diff
@@ -0,0 +1,36 @@
+import Foundation
+import PBXProj
+
+/// A type that generates and writes to disk a `PBXProject` prefix `PBXProj`
+/// partial.
+///
+/// The `Generator` type is stateless. It can be used to generate multiple
+/// partials. The `generate()` method is passed all the inputs needed to
+/// generate a partial.
+struct Generator {
```
thiagohmcruz, line 10: >It can be used to generate multiple partials.

Multiple partials of the same `pbxproject_prefix` "type" I assume correct? Given the `generate()` impl it looks like this is specific to `pbxproject_prefix`. 

Are we planning to implement a `Generator` class like this for all partial generators?

tools/generators/pbxproject_prefix/test/CalculateCompatibilityVersionTests.swift
```diff
@@ -0,0 +1,103 @@
...
+        XCTAssertEqual(
+            compatibilityVersion,
+            expectedCompatibilityVersion
+        )
+    }
+
+    func test_xcode14() {
+        // Arrange
+
+        let minimumXcodeVersion: SemanticVersion = "14.2.1"
```
thiagohmcruz, line 50: What if a user pass `14.2.0.14C18`? Is this something we should handle/error out nicely if it's not expected?

thiagohmcruz (approved):
Overall LGTM, I have some questions but all non-blocking, mostly for my own learning about intent vs impl approach.

tools/generators/README.md
```diff
@@ -0,0 +1,70 @@
+# `.xcodeproj` generators
+
+The `.xcodeproj` file bundle has multiple files inside of it. The generators
+in this package are responsible for creating those various files.
```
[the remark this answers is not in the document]
thiagohmcruz: Just saw this: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2155#issuecomment-1536354474

Yeah, tricky one to name I guess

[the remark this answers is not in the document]
brentleyjones: When the incremental generator is finished, `tools/generator` will no longer exist. I can instead move it into `tools/generators/legacy` or something?

tools/pbxproj_generators/project_prefix/test/CalculateProjectPrefixTests.swift
thiagohmcruz: I wouldn't over engineer it as well, IMO a comment with details should be enough

tools/generators/README.md
brentleyjones: What different wording? They seem identical to me.

tools/generators/pbxproject_prefix/README.md
brentleyjones: I'm only using flags like `--output-path` if required, to reduce Starlark memory usage. Times when they are required:

- Optional values
- To indicate the end of one list of values and the start of another

There is a slight readability loss, but the only place you should be constructing these are in the custom scheme definition (where there are comments), or in the Starlark code that creates the `args()` object (which will also have comments).

tools/generators/pbxproject_prefix/src/Generator/CalculatePBXProjectPrefix.swift
brentleyjones: In my mind generation is the whole thing, which is the combination of calculating it and then writing it. It's not generated until it's written to disk.

tools/generators/pbxproject_prefix/src/Generator/CalculateProjectDir.swift
brentleyjones: I'm open to better names. Is `projectDir(executionRoot:)` clear? I'm bad at function names, but if this is correct Swift style, I'm all for it!

tools/generators/pbxproject_prefix/src/Generator/Generator.swift
brentleyjones: In practice you wouldn't use the same instance of `PBXProjectPrefix.Generator` to generate multiple partials, but it's stateless, so you could.

> Are we planning to implement a `Generator` class like this for all partial generators?

Yes. Each generator tool will have its own `Generator` struct.

tools/generators/pbxproject_prefix/test/CalculateCompatibilityVersionTests.swift
brentleyjones: That will never be sent to this tool. We force a specific format for this version here: https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/925d111ccfb8d65f7701e352ae6daeddf9e61b96/xcodeproj/internal/xcodeproj_rule.bzl#L122-L129

tools/generators/pbxproject_prefix/src/Generator/CalculateProjectDir.swift
BalestraPatrick: Since this function doesn't have any side effects, I think simply calling it `projectDir` is fine.

BalestraPatrick (approved):

tools/generators/README.md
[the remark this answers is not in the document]
brentleyjones: I moved it.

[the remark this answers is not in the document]
thiagohmcruz: Ok, moving to `legacy` works to address what I mentioned above, thx!

tools/generators/README.md
thiagohmcruz: My confusion was because I don't see the word "all" in`pbxproject_prefix/README.md`:
> The `pbxproject_prefix` generator creates a `PBXProj` partial containing the
start (i.e. prefix) of the `PBXProject` element.

But I'm overcomplicating this, this `README` here does mention `start` in the first item and then details what to expect specifically: all `PBXProject` except for X and Y

I'm going to close this nvm, not even a nit tbh haha

tools/generators/pbxproject_prefix/README.md
thiagohmcruz: >There is a slight readability loss, but the only place you should be constructing these are in the custom scheme definition (where there are comments), or in the Starlark code that creates the args() object (which will also have comments).

Ok, perfect, this addresses exactly what I was concerned about and agree that it should be fine then

tools/generators/pbxproject_prefix/src/Generator/CalculateProjectDir.swift
thiagohmcruz: +1

tools/generators/pbxproject_prefix/src/Generator/Generator.swift
thiagohmcruz: Any value in making these `Generator` structs adopt a protocol so we know "a partial Generator" is expected to be able to "init with env, generate, <maybe more expected actions in the future?>" 

Probably not something for this PR, though, just wondering if it could end up evolving this way and requiring a common interface

tools/generators/pbxproject_prefix/test/CalculateCompatibilityVersionTests.swift
thiagohmcruz: 👍

tools/generators/pbxproject_prefix/src/Generator/Generator.swift
brentleyjones: I personally don't see a value in that. There is only a separation between `PBXProjectPrefix` and `Generator`, because we can't set `environment` in `PBXProjectPrefix.run()`. The slip makes it so `PBXProjectPrefix` parses the arguments, then those arguments get passed to `Generator`. The environment and arguments will be tied to each generator. Each command (`PBXProjectPrefix.run` in this case) will work with exactly one generator, so no need for an interface. Also, the command and generator will be the only ones in their respective module/tool.

So yes, there will be multiple commands and generators, but that's just because there will be multiple tools. Each tool could technically do whatever it wants though, it's just that we will be following the pattern set here as a common architecture.

tools/generators/pbxproject_prefix/src/Generator/CalculatePBXProjectPrefix.swift
brentleyjones: Went with `calculate()` for this and `write()` for the other function. Since they are the primary functions of the generator, this kinda makes it clearer (by removing a confusing second part of the name).

tools/generators/pbxproject_prefix/src/Generator/ProjectDir.swift
```diff
@@ -0,0 +1,14 @@
...
+
+extension Generator {
+    /// Calculates the `PBXProject.projectDir` string.
+    ///
+    /// - Parameters:
+    ///   - executionRoot: The absolute path to the Bazel execution root.
+    static func projectDir(executionRoot: String) -> String {
+        guard !executionRoot.hasPrefix("/private/") else {
+            return String(executionRoot.dropFirst(8))
+        }
```
mattrobmattrob, line 11: Mind adding a comment about why you're removing `/private` in this case?

tools/generators/pbxproject_prefix/src/Generator/Calculate.swift
```diff
@@ -0,0 +1,41 @@
...
+        developmentRegion: String,
+        organizationName: String?,
+        projectDir: String,
+        workspace: String
+    ) -> String {
+        let organizationNameAttribute: String
+        if let organizationName = organizationName {
+            organizationNameAttribute = #"""
+				ORGANIZATIONNAME = \#(organizationName.pbxProjEscaped);
+
```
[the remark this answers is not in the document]
mattrobmattrob: Seemingly, it's just so it always has a new line at the bottom?

mattrobmattrob (approved):
IMO, this is going to drastically improve the understandability of the repo. 👍

tools/generators/README.md
```diff
@@ -0,0 +1,70 @@
...
+  - Creates automatic `.xcscheme`s
+- `files_and_groups`:
+  - `PBXFileReference`s
+  - `PBXGroup`s
+  - Closes the `PBXProj` element
+
+## Xcode schemes
+
+[Automatic schemes](docs/bazel.md#xcodeproj-scheme_autogeneration_mode) are
+generated by one of the `PBXProj` partial generators.
```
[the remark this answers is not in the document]
brentleyjones: Yes.

tools/generators/lib/PBXProj/src/Identifiers.swift
```diff
@@ -0,0 +1,13 @@
+/// Helps set identifiers for `PBXProj` elements.
+///
+/// Identifiers are unique 12 byte numbers encoded as 24 character hex strings.
+/// Since the various `PBXProj` partial generators need to work independently
+/// from each other, they need help to be able to generate unique identifiers.
+/// `Identifiers` is used to generate these identifiers.
```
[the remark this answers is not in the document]
brentleyjones: I'll be sure to do that as I add more.

tools/generators/lib/PBXProj/src/String+Extensions.swift
```diff
@@ -0,0 +1,60 @@
...
+            charactersIn: UnicodeScalar("A") ... UnicodeScalar("Z")
+        )
+        invalidSet.insert(
+            charactersIn: UnicodeScalar("a") ... UnicodeScalar("z")
+        )
+        invalidSet.invert()
+        return invalidSet
+    }()
+
+    private static var specialCheckCharacters = CharacterSet(charactersIn: "_/")
```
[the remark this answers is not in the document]
brentleyjones: Lifted from `XcodeProj`, so I'm not sure tbh.

tools/generators/pbxproject_prefix/BUILD
```diff
@@ -0,0 +1,102 @@
...
+
+# Tests
+
+config_setting(
+    name = "release_build",
+    values = {
+        "compilation_mode": "opt",
+    },
+)
+
```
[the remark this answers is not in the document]
brentleyjones: Not anymore

tools/generators/pbxproject_prefix/src/Generator/Arguments.swift
```diff
@@ -0,0 +1,39 @@
...
+        )
+        var outputPath: URL
+
+        @Argument(help: "Absolute path to the Bazel workspace.")
+        var workspace: String
+
+        @Argument(
+            help: """
+Path to a file that contains the absolute path to the Bazel execution root.
+""",
```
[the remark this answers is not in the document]
brentleyjones: It's because the path is generated during the execution phase, so we can't pass the value itself as an argument. This is because we can't get absolute paths in the analysis phase.

tools/generators/pbxproject_prefix/src/Generator/Calculate.swift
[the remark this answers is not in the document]
brentleyjones: Needs the newline at the bottom, because of how the final `cat` will work. All the partials need trailing newlines (like real unix text files).

tools/generators/README.md
```diff
@@ -0,0 +1,70 @@
+# `.xcodeproj` generators
+
+The `.xcodeproj` file bundle has multiple files inside of it. The generators
+in this package are responsible for creating those various files.
+
+## `PBXProj`
+
+The most important file in the `.xcodeproj` file bundle is the `project.pbxproj`
+file. This file details the project itself, and controls almost of all Xcode’s
```
luispadron, line 9:
```suggestion
file. This file details the project itself, and controls almost all of Xcode’s
```

tools/generators/README.md
luispadron, line 1: Really helpful!

luispadron (approved):
Sweet!

tools/generators/pbxproject_prefix/src/Generator/Calculate.swift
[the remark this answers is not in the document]
brentleyjones: I'll add that as a comment.

test/fixtures/generator/BUILD
```diff
@@ -1,5 +1,5 @@
 load(
-    "//tools/generator:xcodeproj_targets.bzl",
```
brentleyjones, line 2: I'm going to land the legacy rename as its own PR.

PR #2156, opened 2023-05-05
Format some code

brentleyjones:
Cleaned up while I was in the area for another reason.

### DIFF
--- a/tools/generator/src/Generator/SemanticVersion.swift
+++ b/tools/generator/src/Generator/SemanticVersion.swift
@@ -47,13 +47,7 @@ public struct SemanticVersion: Equatable, Hashable {
 
 extension SemanticVersion: Comparable {
     public static func < (lhs: SemanticVersion, rhs: SemanticVersion) -> Bool {
-        guard lhs.major == rhs.major else {
-            return lhs.major < rhs.major
-        }
-        guard lhs.minor == rhs.minor else {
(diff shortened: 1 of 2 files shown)


tools/generator/src/Generator/SemanticVersion.swift
```diff
@@ -45,21 +47,25 @@ struct SemanticVersion: Equatable, Hashable {
 
 extension SemanticVersion: Comparable {
     public static func < (lhs: SemanticVersion, rhs: SemanticVersion) -> Bool {
-        return (lhs.major < rhs.major)
-            || (lhs.major == rhs.major && lhs.minor < rhs.minor)
-            || (lhs.major == rhs.major && lhs.minor == rhs.minor && lhs.patch < rhs.patch)
+        guard lhs.major == rhs.major else {
```
jpsim, line 50: This can be done more simply:

```swift
(lhs.major, lhs.minor, lhs.patch) < (rhs.major, rhs.minor, rhs.patch)
```

jpsim (approved):

tools/generator/src/Generator/SemanticVersion.swift
brentleyjones: Me and my old pre-SE-0283 ways! Thanks 😄.

PR #2158, opened 2023-05-05
Update FAQ reg: Xcode project not showing up simulators

sanju-naik:
Updates FAQ document reg: Project not showing up any simulators when generated using rules_xcodeproj.


luispadron (approved):
Thanks!

docs/faq.md
```diff
@@ -158,3 +158,10 @@ fallback profiles, or if you use specify a profile in the workspace.
 
 Copying the profile to `~/Library/MobileDevice/Provisioning Profiles` will
 resolve the error.
+
+## Why I do not see any Simulators after generating project using rules_xcodeproj?
+
+This can happen if you have opened Xcode using `rosetta`, the solution is to get native arm simulator support for your app and disable rosetta.
+
+![Alt text](/docs/xcode-rosetta.png?raw=true "Xcode without Rosetta")
```
mattrobmattrob, line 166: I'd prefer we not include the 0.5 MB image just for the FAQ as long as that's okay with you. I think the text description is enough since it's not necessarily just as simple as disabling the checkbox, right?

brentleyjones: The way we’ve handled images elsewhere is to upload them to GitHub, and use that link instead.

brentleyjones: (Just in case you still want to use the image.)

docs/faq.md
```diff
@@ -158,3 +158,10 @@ fallback profiles, or if you use specify a profile in the workspace.
 
 Copying the profile to `~/Library/MobileDevice/Provisioning Profiles` will
 resolve the error.
+
+## Why I do not see any Simulators after generating project using rules_xcodeproj?
```
brentleyjones, line 162:
```suggestion
## Why I do not see any simulators after generating project using rules_xcodeproj?
```

brentleyjones: Can you move this section above `Why do I get an error like “Provisioning profile "PROFILE_NAME" is Xcode managed, but signing settings require a manually managed profile”?`

brentleyjones (approved):

brentleyjones:
You'll need to sign your commits to pass the DCO check.

docs/faq.md
sanju-naik: Done

docs/faq.md
sanju-naik: Makes sense 👍 , removed the image.

mattrobmattrob (approved):

Issue #1748, opened 2023-02-03
Bug: Resource groups appear in a project with a target focusing


CognitiveDisson:
## Describe the bug

In BwB mode, resource group files are included in the project even if the target is unfocused. There is no way to unfocus them.

## Example

We have a swift library:
```
swift_library(
  name = "SomeSwiftLibrary",
  data = ["//Some:SomeSwiftLibrary_Resources"]
)
```
and a resource group:
```
apple_resource_group(
  name = "SomeSwiftLibrary_Resources",
  resources = [
    "/Some:Resources/asset.xcassets/Contents.json",
    "/Some:Resources/asset.xcassets/Contents.json",
    "/Some:Resources/asset.xcassets/image.imageset/image.png",
    "/Some:Resources/asset.xcassets/image.imageset/image@2x.png", 
    "//Some:Resources/asset.xcassets/image.imageset/image@3x.png"
  ],
)
```

Even if we add `//Some:SomeSwiftLibrary_Resources` and `//Some:SomeSwiftLibrary` to `unfocused_targets`, the `asset.xcassets` will still be present in the project.

```
xcodeproj(
    name = "xcodeproj",
    project_name = "Project",
    build_mode = "bazel",
    top_level_targets = ["//iOSApp"],
    focused_targets = [...], // do not contain `SomeSwiftLibrary`
    unfocused_targets = ["/Some:SomeSwiftLibrary", "//Some:SomeSwiftLibrary_Resources"],
)
```

## Expected behavior

If a target is not focused, all of its resource files should not appear in the project.

brentleyjones:
Seems we will need to associate `owner` to `resources` and `extra_files` in `internal/resources.bzl`, instead of just the `file_path`s. Then when we process those in `xcodeproj_rule.bzl`, at least for top-level bundles that aren't resource bundles, we can filter those files to only ones in `focused_labels`.

narlei:
I've tried to put the resources into the `unfocused_targets` declaration, but it didn't work too. The focus mode is amazing and this project is what makes Bazel productive for our project. Please let me know if I can help with something, this bugfix is very important to us. Again, thank you for this project!

brentleyjones:
Hi @narlei. My above comment still holds for how I would go about trying to fix this, if you want to take a stab at it. Though tight now you can unfocus resource bundles, just not resource groups.

PR #2159, opened 2023-05-08
Yongjin/fix unfocusing resource filegroups

yongjincho92:
**Background**
Addresses https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/1748

**Change**
Per https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/1748:
- associate owner to resources in `internal/resources.bzl`, instead of just the file_paths. 
- process those in `xcodeproj_rule.bzl`, filter those files to only ones in focused_labels.


xcodeproj/internal/resources.bzl
```diff
@@ -343,6 +343,19 @@ def collect_resources(
         parent.dependency_paths.append(child_bundle_path)
 
     frozen_bundles = []
+    resource_to_owners = {}
+    for (resource, owner) in resource_info.owners.to_list():
+        # AppleResourceInfo treats resource files inside xcassets separately, whereas
+        # rules_xcodeproj only cares about .xcassets folder
+        if ".xcassets" in resource:
+            resource = resource.split(".xcassets")[0] + ".xcassets"
```
brentleyjones, line 351: You can use the `max_splits` argument here. Or maybe `partition` (like the code that normalizes it elsewhere, maybe just use `normalize_file_path`?).

xcodeproj/internal/input_files.bzl
```diff
@@ -775,7 +775,7 @@ def _from_resource_bundle(bundle):
         non_arc_srcs = EMPTY_LIST,
         pch = None,
         resource_bundle_dependencies = bundle.dependencies,
-        resources = depset(bundle.resources),
+        resources = depset([(resource, None) for resource in bundle.resources]) if bundle.resources else EMPTY_DEPSET,
```
brentleyjones, line 778: Should this be `None`?

brentleyjones, line 778: I think you will need to do `folder_resources` as well, which are just structured resources.

xcodeproj/internal/resources.bzl
```diff
@@ -343,6 +343,19 @@ def collect_resources(
...
+
+        # a resource can have multiple owners
+        if resource in resource_to_owners and owner not in resource_to_owners[resource]:
```
brentleyjones, line 354: Instead of `owner not in resource_to_owners[resource]`, you could make the value of the `dict` a set (a `dict` with keys as the values storing, and `None` as the value`), and then just insert without condition. Also, instead of `if resource in resource_to_owners` you can use `setdefault()`.

xcodeproj/internal/resources.bzl
```diff
@@ -374,7 +395,7 @@ def collect_resources(
             ]
             if bundle
         ],
-        resources = root_bundle.resources,
+        resources = resource_owner_pairs,
         folder_resources = root_bundle.folder_resources,
```
brentleyjones, line 399: Should do `folder_resources` as well.

xcodeproj/internal/xcode_targets.bzl
```diff
@@ -796,11 +798,13 @@ def _build_settings_to_dto(
 
     return build_settings
 
-def _inputs_to_dto(inputs):
+def _inputs_to_dto(inputs, focused_labels, unfocused_labels):
```
brentleyjones, line 801:
```suggestion
def _inputs_to_dto(inputs, *, focused_labels, unfocused_labels):
```

xcodeproj/internal/resources.bzl
yongjincho92: sounds good!

xcodeproj/internal/input_files.bzl
yongjincho92: not sure if I follow the question. Do you mean just set `None` instead of `EMPTY_DEPSET`?

brentleyjones: I mean, I don't think it should use `None`. Either this should already be a list of tuples, or we should associate a label here. Feels like it should already be a list of tuples though.

yongjincho92: ah makes sense

xcodeproj/internal/resources.bzl
yongjincho92: good point!

xcodeproj/internal/input_files.bzl
yongjincho92: I just remembered the reason I had this was because of these lines  
https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/main/xcodeproj/internal/resources.bzl#L110-L128

Right now, I am post processing `root_bundle.resources` and `root_bundle.folder_resources` after the attributes are set, so `bundle` inside the `_from_resource_bundle` method doesn't have `resource` of type List of Tuples, but rather List of Strings (as expected). 

I think it makes sense for us to actually change `bundle.resources` and `bundle.folder_resources` to have types List of tuples to match. What do you think of that?

brentleyjones: I currently don't have an opinion on where it should be calculated. It should just be the most efficient as possible, and it should correctly set the label.

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -462,10 +462,10 @@ targets.
             # Don't create targets for resource bundles in BwB mode, but still
             # include their files if they aren't unfocused
             focused_targets_extra_files.append(
-                (xcode_target.label, xcode_target.inputs.resources),
+                (xcode_target.label, depset([r[0] for r in xcode_target.inputs.resources.to_list()])),
```
brentleyjones: You should find a way to not use `.to_list()`. Also, this isn't correct, is it? It would be changing the owner of these files from whatever we set to `xcode_target.label`. Should this instead be something like this?
```suggestion
                xcode_target.inputs.resources,
```

And for that to work, the resources in the tuple would have to be a `depset()`.

yongjincho92: > You should find a way to not use .to_list()

Yeah I was iffy about it that I had to make it to list and then to depset again. I agree on trying to find a way not to use it. 

> Also, this isn't correct, is it? It would be changing the owner of these files from whatever we set to `xcode_target.label`

Yes, you're right.

yongjincho92:
According to the errors, it looks like I might need to change more. 

```
focused_targets_extra_files.append(xcode_target.inputs.resources) 
```
type of `(xcode_target.inputs.resources)` = 'depset(tuple(`owner`,`resource`))'

and from 
```
        for file, owner_label in owned_extra_files.items():
            if label_str == owner_label:
                focused_targets_extra_files.append(
                    (
                        label,
                        depset([file.path for file in file.files.to_list()]),
                    ),
                )
```

we are adding entries of type `tuple(Label, depset(String))`

I think it would be more aligned if we make `xcode_target.inputs.resources` to the same type `tuple(Label, depset(String))`

yongjincho92:
@narlei can you check if my change works on your end? I have checked on our end that it's working fine.

xcodeproj/internal/input_files.bzl
```diff
@@ -63,6 +63,21 @@ def _process_cc_info_headers(headers, *, exclude_headers, generated):
...
+        resources: A `list` of tuples (`owner`, `file_path`) of resources that should be
+            added to the target's bundle.
+
+    Returns:
+        A `list` of tuples (`owner`, depset([`resource`]))
+    """
+    label_to_resources_depset = {}
+    for (owner, resource) in resources:
+        label_to_resources_depset.setdefault(owner, {})[resource] = None
+    return [(owner, depset(label_to_resources_depset[owner].keys())) for owner in label_to_resources_depset]
```
brentleyjones, line 79:
```suggestion
    return [
        (owner, depset(resources.keys()))
        for resources, owner in label_to_resources_depset.items()
    ]
```

xcodeproj/internal/input_files.bzl
```diff
@@ -421,9 +436,9 @@ def _collect_input_files(
         if resources_result.dependencies:
             resource_bundle_dependencies = resources_result.dependencies
         if resources_result.resources:
-            resources = depset(resources_result.resources)
+            resources = depset(_transform_into_label_to_resources(resources_result.resources))
```
brentleyjones, line 439:
```suggestion
            resources = depset(
                _transform_into_label_to_resources(resources_result.resources),
            )
```

xcodeproj/internal/input_files.bzl
```diff
@@ -421,9 +436,9 @@ def _collect_input_files(
...
         if resources_result.folder_resources:
-            folder_resources = depset(resources_result.folder_resources)
+            folder_resources = depset(_transform_into_label_to_resources(resources_result.folder_resources))
```
brentleyjones, line 441:
```suggestion
            folder_resources = depset(
                _transform_into_label_to_resources(
                    resources_result.folder_resources,
                ),
            )
```

xcodeproj/internal/input_files.bzl
```diff
@@ -764,7 +779,7 @@ def _from_resource_bundle(bundle):
     return struct(
         compiling_output_group_name = None,
         entitlements = None,
-        folder_resources = depset(bundle.folder_resources),
+        folder_resources = depset(_transform_into_label_to_resources(bundle.folder_resources)),
```
brentleyjones, line 782:
```suggestion
        folder_resources = depset(
            _transform_into_label_to_resources(bundle.folder_resources),
        ),
```

xcodeproj/internal/resources.bzl
```diff
@@ -14,6 +14,15 @@ load(":target_id.bzl", "get_id")
...
 # Utility
 
+def _normalize_resource_path(resource):
+    for extension in RESOURCES_FOLDER_TYPE_EXTENSIONS:
+        if extension not in resource:
+            continue
+        prefix, ext, _ = resource.partition(extension)
+        resource = prefix + ext
+        break
+    return resource
```
brentleyjones, line 24:
```suggestion
        return prefix + ext

    return resource
```

xcodeproj/internal/resources.bzl
```diff
@@ -261,7 +299,7 @@ def collect_resources(
             `process_resource_bundles`.
         *   `dependencies`: A `list` of `id`s of resource bundle targets that
             this target depends on.
-        *   `resources`: A `list` of `file_path`s of resources that should be
+        *   `resources`: A `list` of tuples (`owner, `file_path`) of resources that should be
             added to the target's bundle.
```
brentleyjones, line 303:
```suggestion
        *   `resources`: A `list` of tuples (`owner, `file_path`) of resources
            that should be added to the target's bundle.
```

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -461,12 +461,8 @@ targets.
...
             # Don't create targets for resource bundles in BwB mode, but still
             # include their files if they aren't unfocused
-            focused_targets_extra_files.append(
-                (xcode_target.label, xcode_target.inputs.resources),
-            )
-            focused_targets_extra_folders.append(
-                (xcode_target.label, xcode_target.inputs.folder_resources),
-            )
+            focused_targets_extra_files.extend(xcode_target.inputs.resources.to_list())
+            focused_targets_extra_folders.extend(xcode_target.inputs.folder_resources.to_list())
```
brentleyjones, line 465:
```suggestion
            focused_targets_extra_files.extend(
                xcode_target.inputs.resources.to_list(),
            )
            focused_targets_extra_folders.extend(
                xcode_target.inputs.folder_resources.to_list(),
            )
```

I don't like the `to_list()` here. I think we can avoid it?

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -732,6 +728,8 @@ targets.
         ) = xcode_targets.to_dto(
             xcode_target,
             actions = actions,
+            focused_labels = focused_labels,
+            unfocused_labels = unfocused_labels,
```
brentleyjones, line 732: Can you alphabetize?

narlei:
> @narlei can you check if my change works on your end? I have checked on our end that it's working fine.

Hi, I tested with your branch, and it worked as well.
Using hyperfine, I did a performance test and I didn't notice significant changes.

xcodeproj/internal/xcodeproj_rule.bzl
yongjincho92: you mean `focused_targets_extra_folders.extend(xcode_target.inputs.folder_resources)`?

xcodeproj/internal/resources.bzl
```diff
@@ -122,20 +125,12 @@ def _add_structured_resources_to_bundle(
             generated.append(file)
 
         if not inner_dir:
-            owners = resource_to_owners.get(file.path, {})
-            if not owners:
-                bundle.resources.append((None, file.path))
-            for owner in owners:
-                bundle.resources.append((owner, file.path))
+            _update_bundle_owner_resource_tuples(file.path, bundle.resources, resource_to_owners)
```
brentleyjones, line 128:
```suggestion
            _update_bundle_owner_resource_tuples(
                resource = file.path,
                owner_resource_tuples = bundle.resources,
                resource_to_owners = resource_to_owners,
            )
```

xcodeproj/internal/resources.bzl
```diff
@@ -23,6 +23,13 @@ def _normalize_resource_path(resource):
 
     return resource
 
+def _update_bundle_owner_resource_tuples(resource, owner_resource_tuples, resource_to_owners):
```
brentleyjones, line 26:
```suggestion
def _update_bundle_owner_resource_tuples(
        *,
        resource,
        owner_resource_tuples,
        resource_to_owners):
```

xcodeproj/internal/resources.bzl
```diff
@@ -91,11 +98,7 @@ def _add_resources_to_bundle(
             xccurrentversions = xccurrentversions,
         )
         if fp:
-            resource_owners = resource_to_owners.get(fp, {})
-            if not resource_owners:
-                bundle.resources.append((None, fp))
-            for resource_owner in resource_owners:
-                bundle.resources.append((resource_owner, fp))
+            _update_bundle_owner_resource_tuples(fp, bundle.resources, resource_to_owners)
```
brentleyjones, line 101:
```suggestion
            _update_bundle_owner_resource_tuples(
                resource = fp,
                owner_resource_tuples = bundle.resources,
                resource_to_owners = resource_to_owners,
            )
```

xcodeproj/internal/resources.bzl
```diff
@@ -122,20 +125,12 @@ def _add_structured_resources_to_bundle(
...
             continue
 
         # Special case for localized
         if inner_dir.endswith(".lproj"):
-            owners = resource_to_owners.get(file.path, {})
-            if not owners:
-                bundle.resources.append((None, file.path))
-            for owner in owners:
-                bundle.resources.append((owner, file.path))
+            _update_bundle_owner_resource_tuples(file.path, bundle.resources, resource_to_owners)
```
brentleyjones, line 133:
```suggestion
            _update_bundle_owner_resource_tuples(
                resource = file.path,
                owner_resource_tuples = bundle.resources,
                resource_to_owners = resource_to_owners,
            )
```

xcodeproj/internal/resources.bzl
```diff
@@ -147,13 +142,7 @@ def _add_structured_resources_to_bundle(
...
 
         folder_resource = paths.join(dir[:-(1 + len(nested_path))], inner_dir)
-        owners = resource_to_owners.get(folder_resource, {})
-        if not owners:
-            bundle.folder_resources.append((None, folder_resource))
-        for owner in owners:
-            bundle.folder_resources.append(
-                (owner, folder_resource),
-            )
+        _update_bundle_owner_resource_tuples(folder_resource, bundle.folder_resources, resource_to_owners)
```
brentleyjones, line 145:
```suggestion
        _update_bundle_owner_resource_tuples(
            resource = folder_resource,
            owner_resource_tuples = bundle.folder_resources,
            resource_to_owners = resource_to_owners,
        )
```

brentleyjones (approved):

luispadron (approved):

PR #2173, opened 2023-05-10
Add `BazelDependencies` `PBXProj` partial generator

brentleyjones:


brentleyjones:
@MobileNativeFoundation/rules_xcodeproj-maintainers Would it be better if this change _also_ included the starlark, similar to #2176? They are reviewable and testable on their own, but I realize they have some connection, and it might make it easier to review to have both of them in the same PR. Up to y'all.

brentleyjones:
Kept it separate for now: #2197. LMK if you think should be combined in the future.

brentleyjones:
Since the generator is rarely run, I can probably change it to make it easier to read (at the expense of a little performance). Any thoughts on how you would like it changed?

brentleyjones:
I'm going to assume the conditional adding of stuff, and how it's formatted around newlines, is one of the main issues. I can minimally build up an array of fragments, and join them at the end.

mattrobmattrob:
Yeah, I think it's the intra-string `\(...)` that obfuscate the actual structure a bit.

BalestraPatrick (approved):
Looks good to me, no particular comments!

brentleyjones:
@mattrobmattrob I think `calculate()` should be a lot more readable now, can you let me know what you think?

mattrobmattrob:
Looks better to me. Thanks, @brentleyjones !

tools/BUILD
```diff
@@ -22,6 +27,40 @@ _SCHEME_DIAGNOSTICS = xcode_schemes.diagnostics(
...
+                "Release",
+                # defaultXcodeConfiguration
+                "--default-xcode-configuration",
+                "Release",
+                # platforms
+                "--platforms",
+                "macosx",
+                "iphonesimulator",
+                # preBuildScript
+                # postBuildScript
```
thiagohmcruz, line 53: I assume this is not a leftover right? It's here to show the reader that these options are also available

tools/generators/bazel_dependencies/src/Generator/BuildSettings.swift
```diff
@@ -0,0 +1,59 @@
...
+        indexImport: String,
+        platforms: [Platform],
+        targetIdsFile: String
+    ) -> String {
+        let sortedPlatforms = Set(platforms).sorted()
+
+        // We have to support only a single platform during Index Build to
+        // prevent issues duplicated outputs, but it also has to be a platform
+        // that one of the targets uses, otherwise it's not invoked at all.
+        // Index Build is so weird...
```
thiagohmcruz, line 23: +1

tools/generators/bazel_dependencies/src/Generator/Calculate.swift
```diff
@@ -0,0 +1,166 @@
...
+            ))
+        }
+
+        buildPhases.append((
+            id: Identifiers.BazelDependencies.bazelBuild,
+            element: #"""
+		\#(Identifiers.BazelDependencies.bazelBuild) = {
+			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
+			buildActionMask = 2147483647;
```
thiagohmcruz, line 32: I assume this is always the same and that's why we can hard code it like this right?

thiagohmcruz (approved):
Looks great, the process of adding partials is very clear to me and this is a good example PR for future similar changes. Only small questions for my own learning.

tools/BUILD
brentleyjones: Yeah

tools/generators/bazel_dependencies/src/Generator/Calculate.swift
brentleyjones: Yeah: https://github.com/tuist/XcodeProj/blob/f570155209af12643309ac4e758b875c63dcbf50/Sources/XcodeProj/Objects/BuildPhase/PBXBuildPhase.swift#L6

tools/generators/bazel_dependencies/src/Generator/RunScriptBuildPhase.swift
```diff
@@ -0,0 +1,33 @@
...
+    ///   - script: The text of the script.
+    static func runScriptBuildPhase(name: String, script: String?) -> String? {
+        guard let script = script else {
+            return nil
+        }
+
+        return #"""
+{
+			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
```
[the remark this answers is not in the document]
brentleyjones: We could change that in the future. Current generator though doesn't have it as an option.

PR #2176, opened 2023-05-10
Add `pbxproj_partials.write_pbxproject_prefix`

brentleyjones:


test/internal/pbxproj_partials/BUILD
```diff
@@ -0,0 +1,15 @@
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load(
+    ":write_pbxproject_prefix_tests.bzl",
+    "write_pbxproject_prefix_test_suite",
+)
+
+write_pbxproject_prefix_test_suite(name = "write_pbxproject_prefix")
+
+test_suite(name = "pbxproj_partials")
```
thiagohmcruz, line 9: I don't have experience with writing `@bazel_skylib//lib:unittest.bzl` tests tbh, mind clarifying why this `test_suite` rule is necessary here if it's not receiving any actual tests?

xcodeproj/internal/pbxproj_partials.bzl
```diff
@@ -0,0 +1,78 @@
...
+    # colorize
+    if colorize:
+        args.add("--colorize")
+
+    actions.run(
+        arguments = [args],
+        executable = tool,
+        inputs = [execution_root_file],
+        outputs = [output],
+        mnemonic = "WriteXcodeProjPBXProjectPrefixPBXProjPartial",
```
thiagohmcruz, line 71: Nit: This mnemonic seems to hold a lot of info? Do you think it's important to add all these? What about `WritePBXProjPrefix` since `PBXProj` already implies everything written in [tools/generators/README.md](https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/main/tools/generators/README.md)?

xcodeproj/internal/pbxproj_partials.bzl
```diff
@@ -0,0 +1,78 @@
...
+
+    # executionRootFile
+    args.add(execution_root_file)
+
+    # minimumXcodeVersion
+    args.add(minimum_xcode_version)
+
+    # developmentRegion
+    development_region = project_options.get("development_region")
+    if development_region and development_region != "en":
```
thiagohmcruz, line 54: Remind me why special case `en` here? It's the default?

thiagohmcruz (approved):
It was very easy to follow the way these tests are structure and def a good pattern to follow for other partials, no major issues here for me. 🚀

thiagohmcruz (approved):
It was very easy to follow the way these tests are structure and def a good pattern to follow for other partials, no major issues here for me. 🚀

test/internal/pbxproj_partials/BUILD
brentleyjones: If you don’t specify tests, it grabs all of them defined in the package. So it’s receiving the test suite listed above it.

xcodeproj/internal/pbxproj_partials.bzl
brentleyjones: Yeah, I can shorten it.

luispadron (approved):

xcodeproj/internal/pbxproj_partials.bzl
brentleyjones: Went with `WritePBXProjPBXProjectPrefix`. All of the partials will use the `WritePBXProj` prefix, so they can get modified with `--modify_execution_info` easily. Then they will have a suffix that matches their generator name.

PR #2184, opened 2023-05-11
Add support for native test_suite rules as top level targets

thiagohmcruz:
Work towards #1956

Add support for native `test_suite` as top level target. Next is to support `ios_*_test_suite` rules (#1968)

**Note on Custom schemes**: 

Native `test_suite` rules will just wrap other tests and be completely skipped during project generation. There's no test bundle in its deps like `ios_*_test_suite` rules so we won't collect a label that can be replaced later, this effectively means that AFAICT  there's no easy way to allow adding `test_suite` targets in custom schemes? Something like this passed to the generator currently fails:
```sh
schemes = [
    xcode_schemes.scheme(
        name = "My-Custom-Scheme",
        test_action = xcode_schemes.test_action(
            targets = [
                "//foo/bar:SomeNativeTestSuite",
            ],
        ),
    ),
],
```
with an err similar to this:
```sh
Target @//foo/bar:SomeNativeTestSuite was not found in the transitive dependencies of @//foo/bar:my_xcodeproj `top_level_targets` attribute.
```

Looking for suggestions on how to support custom schemes if we think it's a must.


brentleyjones:
We should add support for `alias_labels` which would expand a test_suite label into an array of other labels, to allow use in custom schemes.

That can be a follow up though.

brentleyjones (approved):

thiagohmcruz:
> alias_labels

Cool, will do!

luispadron:
Trying this out now, unless im doing something wrong, added a test_suite target to `top_level_targets`. Seeing this when generating:

```sh
Target @//Code/DesignSystem/MooncakeUI:MooncakeUI-Unit-SnapshotTests was not found in the transitive dependencies of @//.cashcli/Projects:CashBazelProject's `top_level_targets` attribute. Did you reference an alias (only actual target labels are supported in Scheme definitions)? Check that @//Code/DesignSystem/MooncakeUI:MooncakeUI-Unit-SnapshotTests is spelled correctly, and if it is, add it or a target that depends on it to @//.cashcli/Projects:CashBazelProject's `top_level_targets` attribute.
```

luispadron:
> Is that a raw `test_suite` or one of the `rules_ios`'s rules? 
Native Bazel `test_suite`

> Also are the deps of the suite `rules_ios` tests or raw `rules_apple` tests? 

The deps are 4 `ios_unit_test` targets from `rules_ios` which all depend on a single `dep_middleman` from `rules_ios`.

> Also if you're setting that target in `schemes` it won't work until the follow up lands (the `alias_labels` thing above), this PR addresses adding it only to `top_level_targets` directly

Ah, that was it (at least to fix generation). The macro was pulling snapshots into a scheme, removing that logic generates the project and the 4 tests get pulled in via just the test_suite label.

However, building now fails (it passes in `main`) with this error:

```sh
Multiple commands produce '/Users/lpadron/Library/Developer/Xcode/DerivedData/CashBazelProject-gtfybecdnveqrzbftrflmnsfapjm/Build/Intermediates.noindex/CashBazelProject.build/bazel-out/applebin_ios-ios_sim_arm64-dbg-ST-ddc4449e8e9a/bin/Code/DesignSystem/MooncakeUI/MooncakeUI-Unit-SnapshotTests_swift/Objects-normal/arm64/TileViewSnapshotTests.o'
```

But maybe this is what you mean by supporting in the next PR.

thiagohmcruz:
@luispadron if you're ok with that I'd appreciate a stamp here!

luispadron (approved):

PR #2188, opened 2023-05-11
Ensure label in get_id is normalized

thiagohmcruz:
We might want to use the label contained in the `id` from `get_id` to compare and/or help calculate things (e.g. #1968). In order to do that it would be helpful to ensure it's normalized like other labels.

Effectively this PR makes an `id` that looks like this today
```sh
//foo/bar:baz some-config-here
```
look like this instead
```sh
@//foo/bar:baz some-config-here
```

Making it "normalized" so it can be compared with actual labels.


thiagohmcruz:
(should've made a draft, forgot that fixtures would break here)

luispadron (approved):

thiagohmcruz:
(I'm glad `./test/update_all_fixtures.sh` exists)

brentleyjones:
We shouldn’t normalize the label in the target id. It should be an opaque type, and we really shouldn’t do any logic based on it. See #2085 for progress towards that.

thiagohmcruz:
> We shouldn’t normalize the label in the target id. It should be an opaque type, and we really shouldn’t do any logic based on it. See #2085 for progress towards that. I have another change locally that hashes the value to make it a true opaque type.

Ok! Put up this to get some feedback, will close and go with a different approach to what I intended to do do enable the test suite stuff. Thx for the context and example!

PR #2196, opened 2023-05-12
Add support for ios_*_test_suite rules as top level targets

### LINKED ISSUE #1956: Bug: Test suite targets not supported as top level targets

thiagohmcruz:
Work towards #1956 (will close #1968 in favour of the changes in this PR).

#2184 enables `test_suite` targets to be used as top level targets, the changes in this PR do the same for `ios_*_test_suite` targets.

* Track the `skip_type` during target generation.
* Use the value of `skip_type` to conditionally create the replacement labels.
    * Specifically, only do work if `skip_type` == `apple_test_bundle` and in that case removes the `rules_apple` internal suffix `.__internal__.__test_bundle` from the replacement label. This allows the custom scheme logic to match the provided `ios_*_test_suite` target label with the values in `replacement_labels` and do the correct replacement when creating the schemes in Xcode.


thiagohmcruz:
@luispadron I'm adding tests but I'd appreciate if you could test this one and let me know how it goes 🙇

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -88,20 +97,26 @@ def _should_skip_target(*, ctx, target):
         target: The `Target` to check.
 
     Returns:
-        `True` if `target` should be skipped for target generation.
+        A `tuple` with two elements:
+
+        -   A `bool` indicating if `target` should be skipped for target generation.
+        -   The `skip_type` for this target.
```
brentleyjones, line 103: No need for the tuple. Can just return the `skip_type` if it should be skipped, or `None` otherwise.

luispadron:
@thiagohmcruz nice! im out until Tuesday but maybe @jszumski has time before then.

xcodeproj/internal/xcodeprojinfo.bzl
thiagohmcruz: Right, I guess if there is a skip type it should indeed be what implies a skip is needed, ok I'll update this

thiagohmcruz: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2196/commits/892ecded45235c59f6c316480d3ca733e3b1242e

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -266,6 +278,20 @@ def _skip_target(
         if attr in deps_attrs and info.xcode_target
     ]
 
+    def _target_replacement_label(info):
+        if not info.xcode_target:
+            return target.label
+        if _APPLE_INTERNAL_TEST_BUNDLE_SUFFIX not in info.xcode_target.label.name:
```
thiagohmcruz, line 284: We also discussed log a warning if the suffix is not present, so I'll probably move the skip type check up and add the warning here

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -97,27 +97,22 @@ def _get_skip_type(*, ctx, target):
...
-        targets.is_test_bundle(
-            target = target,
-            deps = getattr(ctx.rule.attr, "deps", None),
-        ),
-        skip_type.apple_test_bundle,
+    is_test_bundle = targets.is_test_bundle(
+        target = target,
+        deps = getattr(ctx.rule.attr, "deps", None),
     )
+    return skip_type.apple_test_bundle if is_test_bundle else None
```
brentleyjones:
```suggestion
    if targets.is_test_bundle(
        target = target,
        deps = getattr(ctx.rule.attr, "deps", None),
    ):
        return skip_type.apple_test_bundle
    
    return None
```

brentleyjones (approved):

luispadron:
Had a chance to test, I'm seeing the previous issue resolved (where running one of the targets in the test suite would run on all the devices defied in `runner`) it now correctly runs on just the Xcode selected device, even if it doesn't match the runner 👍🏼 .

Some things we should follow-up on:

- `test_suite` in a custom scheme breaks generation which is different than the other `top-level` target behavior. If it's not much more work I'd consider adding support for schemes in this PR/before next release so it's a smoother change.
- From a user point of view, I'd like the targets to be merged into a single test target if they only differ by `runner`, in our project these all test the same thing: 
  
  Screenshot 2023-05-16 at 12 00 17 PM

luispadron (approved):

brentleyjones:
It shouldn't show as all those various versions anymore.

thiagohmcruz:
@luispadron 

>test_suite in a custom scheme breaks generation which is different than the other top-level target behavior. If it's not much more work I'd consider adding support for schemes in this PR/before next release so it's a smoother change.

`test_suite` in custom schemes is currently not working indeed, there is a [follow up](test_suite in a custom scheme breaks generation which is different than the other top-level target behavior. If it's not much more work I'd consider adding support for schemes in this PR/before next release so it's a smoother change.) already that I'm planning to work on next (and yep we should have that before cutting a release)

>From a user point of view, I'd like the targets to be merged into a single test target if they only differ by runner, in our project these all test the same thing:

These should've been "merged", let me see if I can repro

brentleyjones:
Ahh, then those won't be renamed.

xcodeproj/internal/xcodeprojinfo.bzl
thiagohmcruz: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2196/commits/bced9a71e84979eb3d7af74eca1490921aca6b1d

xcodeproj/internal/xcodeprojinfo.bzl
thiagohmcruz: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2196/commits/2044fa8c9e816ed30e0f0aed1637b2a1755079b9

thiagohmcruz:
We might want to squash this 😅

brentleyjones:
We always squash on merge 😄.

examples/integration/iOSApp/Test/SwiftUnitTests/BUILD
```diff
@@ -1,4 +1,5 @@
-load("@build_bazel_rules_apple//apple:ios.bzl", "ios_unit_test")
+load("@build_bazel_rules_apple//apple:ios.bzl", "ios_unit_test", "ios_unit_test_suite")
+load("@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl", "ios_test_runner")
```
brentleyjones, line 2:
```suggestion
load(
    "@build_bazel_rules_apple//apple:ios.bzl",
    "ios_unit_test",
    "ios_unit_test_suite",
)
load(
    "@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl",
    "ios_test_runner",
)
```

examples/integration/iOSApp/Test/UITests/BUILD
```diff
@@ -1,5 +1,6 @@
-load("@build_bazel_rules_apple//apple:ios.bzl", "ios_ui_test")
+load("@build_bazel_rules_apple//apple:ios.bzl", "ios_ui_test", "ios_ui_test_suite")
 load("@build_bazel_rules_swift//swift:swift.bzl", "swift_library")
+load("@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl", "ios_test_runner")
```
brentleyjones, line 3:
```suggestion
load(
    "@build_bazel_rules_apple//apple:ios.bzl",
    "ios_ui_test",
    "ios_ui_test_suite",
)
load(
    "@build_bazel_rules_swift//swift:swift.bzl",
    "swift_library",
)
load(
    "@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl",
    "ios_test_runner",
)
```

examples/integration/test/fixtures/bwb.xcodeproj/rules_xcodeproj/external.xcfilelist
```diff
@@ -18,4 +18,7 @@ $(BAZEL_EXTERNAL)/com_google_google_maps/GoogleMapsCore.xcframework/ios-arm64_x8
 $(BAZEL_EXTERNAL)/examples_command_line_external/ExternalFramework.framework
 $(BAZEL_EXTERNAL)/examples_command_line_external/ImportableLibrary/Library.h
 $(BAZEL_EXTERNAL)/examples_command_line_external/ImportableLibrary/libImportableLibrary.a
+$(BAZEL_EXTERNAL)/examples_ios_app_external/Assets.xcassets
```
brentleyjones, line 21: Interesting that this appears now.

examples/rules_ios/iOSApp/Test/ObjCUnitTests/BUILD
```diff
@@ -2,6 +2,7 @@ load(
     "@build_bazel_rules_ios//rules:test.bzl",
     rules_ios_ios_unit_test = "ios_unit_test",
 )
+load("@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl", "ios_test_runner")
```
brentleyjones, line 5:
```suggestion
load(
    "@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl",
    "ios_test_runner",
)
```

examples/rules_ios/iOSApp/Test/SwiftUnitTests/BUILD
```diff
@@ -1,4 +1,5 @@
 load("@build_bazel_rules_ios//rules:test.bzl", rules_ios_ios_unit_test = "ios_unit_test")
+load("@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl", "ios_test_runner")
```
brentleyjones, line 2:
```suggestion
load(
    "@build_bazel_rules_ios//rules:test.bzl",
    rules_ios_ios_unit_test = "ios_unit_test",
)
load(
    "@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl",
    "ios_test_runner",
)
```

examples/rules_ios/iOSApp/Test/UITests/BUILD
```diff
@@ -1,4 +1,5 @@
 load("@build_bazel_rules_ios//rules:test.bzl", rules_ios_ios_ui_test = "ios_ui_test")
+load("@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl", "ios_test_runner")
```
brentleyjones, line 2:
```suggestion
load(
    "@build_bazel_rules_ios//rules:test.bzl",
    rules_ios_ios_ui_test = "ios_ui_test",
)
load(
    "@build_bazel_rules_apple//apple/testing/default_runner:ios_test_runner.bzl",
    "ios_test_runner",
)
```

examples/integration/test/fixtures/bwb.xcodeproj/rules_xcodeproj/external.xcfilelist
thiagohmcruz: Indeed, not idea if this is surfacing something we should investigate?

brentleyjones: Now that I think about it, it's probably just you adding `"//iOSApp/Test/UITests:iOSAppUITests"` to the targets. Can you do that as a stand alone PR that merges before this, to keep the noise down?

thiagohmcruz: Yep, definitely, will put up that PR in a few. Btw noticed that `//iOSApp/Test/UITests:iOSAppUITests` was also not set as top level target in the `rules_ios` example, should be put both in a separate PR then? Or one for each?

thiagohmcruz: I'll put both in separate commits, PR coming soon

thiagohmcruz: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2202

luispadron:
> Ahh, then those won't be renamed.

Understand this change now more since it's based on `rules_apple` naming conventions, makes sense. We'll probably go that route unless ya'll have thoughts on how to support this better in Bazel:

- Running snapshot tests for all supported devices that are tested via a single label like: `bazel test //Foo:Snapshots` (runs 4 test targets with different runner)
- Running tests for one supported device via a single label like: `bazel test //Foo:SnapshotTests --config=iphone_13` (runs 1 test target with specific run, skips rest)
  - `runners` doesn't support `select` so we do this using the native `test_suite` rule along with `target_compatible_with`

We're probably going to give up allowing running on a single device and move to using the apple test suite rule instead. Folks can run via Xcode to target a device.

examples/integration/xcodeproj_targets.bzl
```diff
@@ -153,4 +156,23 @@ def get_xcode_schemes():
...
+                },
+                targets = [
+                    "//iOSApp/Test/ObjCUnitTests:iOSAppObjCUnitTestSuite",
+                    "//iOSApp/Test/SwiftUnitTests:iOSAppSwiftUnitTestSuite",
+                ],
+                post_actions = [
+                    xcode_schemes.pre_post_action(
+                        name = "Run After Tests",
+                        script = "echo \"Hi\"",
+                        expand_variables_based_on = "//iOSApp/Test/SwiftUnitTests:iOSAppSwiftUnitTests",
```
thiagohmcruz, line 173: Oops, this should probably be one of the `targets` passed above and not this. 

ps: a custom scheme for raw `test_suite` will be added as well but in a follow up (some remaining things to implement there in order to support it)

thiagohmcruz: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2196/commits/77f8c029649d0f4ebe037dc443d6595c40798df7

thiagohmcruz:
Will rebase and make tests `manual` in this PR once #2205 lands.

thiagohmcruz:
> * target_compatible_with

@luispadron The way I see it the set of PRs I'm landing mostly address two use cases:

* Raw `test_suite` => maps all targets 1-1 to Xcode
* `rules_apple`'s `ios_*_test_suite` => "merge" into a single target scheme allowing Xcode to be the runner (this is possible due to known naming pattern and impl details in `rules_apple` specifically)

We should def discuss if your use case is something we want to support but I'd say it's out of scope for this PR. Specifically:

>Running snapshot tests for all supported devices that are tested via a single label like: bazel test //Foo:Snapshots (runs 4 test targets with different runner)

* `test_suite` with any tests + "some configuration upfront?" => run all tests from a single label (should be achievable [via alias_labels](https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2184#issuecomment-1544025699) which would allow `test_suite` to be set as a custom scheme? Not sure yet, need to think more about this)

>Running tests for one supported device via a single label like: bazel test //Foo:SnapshotTests --config=iphone_13 (runs 1 test target with specific run, skips rest)

I think this would be a special case of the item above, same thing but ignoring all runners when `test_suite` is a custom scheme and instead of running all underlying tests "merge" them so Xcode can be the runner for you to select say `iPhone 13` 

Just thinking out loud, I don't have a good answer to this atm. Will iterate on `alias_labels` next and see if the above applies or not.

PR #2199, opened 2023-05-15
Add support for `--experimental_sibling_repository_layout`

brentleyjones:


brentleyjones:
I don't want to merge this yet, as it's a perf regression. Ideally we can determine globally in Starlark when the flag is enabled, and only check for that form of the prefix.

PR #2211, opened 2023-05-18
Sort application schemes before other auto generated schemes

David-Lam:
This pr improve developer experience by making Xcode select the app scheme by default when generating xcodeproj for multiple top level targets.


tools/generators/legacy/src/Generator/CreateAutogeneratedXCSchemes.swift
```diff
@@ -120,7 +121,8 @@ extension Generator {
                         buildMode: buildMode,
                         schemeInfo: schemeInfo
                     ),
-                    isTopLevel: pbxTarget.isTopLevel
+                    isTopLevel: pbxTarget.isTopLevel,
+                    isApplication: pbxTarget.productType == PBXProductType.application
```
luispadron, line 125: nit: Make extension like `pbxTarget.isTopLevel`

luispadron (approved):

yongjincho92 (approved):

David-Lam:
Good call! I think in this case we don't need to check `isTopLevel` since the product type order should already imply top level product type before non top levels.

tools/generators/legacy/src/Generator/CreateAutogeneratedXCSchemes.swift
David-Lam: Thanks for the suggestion! [Moved logic to extension file.](https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2211/files#diff-176ddd00ccbdfa85509edd61a4e12ad0ca0d6d51e6b16acb57f2753a3cea2618R284-R328)

luispadron (approved):

brentleyjones:
I wasn't clear, but I don't think a strict ordering like you did is best. I think it should be groups like I suggested, where types in the same group would be ordered by name. So a command line tool could be listed before an app depending on its name.

David-Lam:
Sounds good! I misunderstand the previous suggestion. [Updated](https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2211/files#diff-176ddd00ccbdfa85509edd61a4e12ad0ca0d6d51e6b16acb57f2753a3cea2618R286-R317) to set product type ordering as group.

tools/generators/legacy/src/Extensions/PBXProductType+Extensions.swift
```diff
@@ -281,4 +281,40 @@ extension PBXProductType {
...
+        switch self {
+        // Applications
+        case .application,
+            .onDemandInstallCapableApplication,
+            .watchApp,
+            .watch2App,
+            .watch2AppContainer,
+            .commandLineTool,
+            .messagesApplication,
+            .xpcService:
```
brentleyjones, line 295: nit:
```suggestion
        case .application,
            .commandLineTool,
            .messagesApplication,
            .onDemandInstallCapableApplication,
            .watchApp,
            .watch2App,
            .watch2AppContainer,
            .xpcService:
```

tools/generators/legacy/src/Extensions/PBXProductType+Extensions.swift
```diff
@@ -281,4 +281,40 @@ extension PBXProductType {
...
+            .driverExtension,
+            .systemExtension,
+            .messagesExtension,
+            .stickerPack,
+            .xcodeExtension:
+            return 1
+        // Tests
+        case .unitTestBundle,
+            PBXProductType.uiTestBundle,
+            PBXProductType.ocUnitTestBundle:
```
brentleyjones, line 313: nit:
```suggestion
        case .ocUnitTestBundle,
            .uiTestBundle,
            .unitTestBundle:
```

brentleyjones (approved):

PR #2219, opened 2023-05-19
Fix formatting of `Identifiers.Project.buildConfiguration()`

brentleyjones:
Was missing leading zeros.


brentleyjones:
A future change will actually use this and be a proper verifier.

PR #2230, opened 2023-05-22
Add `Identifiers.FilesAndGroups.element()`

brentleyjones:


tools/generators/lib/PBXProj/src/Identifiers.swift
```diff
@@ -49,13 +51,79 @@ public enum Identifiers {
...
+        ///   - type: The type of path being identified.
+        ///   - hashCache: A cache that will be used to guarantee that the
+        ///     identifier returned is unique.
+        public static func element(
+            _ path: String,
+            type: ElementType,
+            hashCache: inout Set<String>
+        ) -> String {
+            let hash = elementHash(path + type.rawValue, hashCache: &hashCache)
+            return #"FF\#(hash) /* \#(path) */"#
```
brentleyjones, line 90: Currently this allows us to have 2^80 files. Which is more than enough. We might even cut away more in the future, once we determine how we want to identify targets.

luispadron (approved):

tools/generators/lib/PBXProj/src/Identifiers.swift
```diff
@@ -49,13 +51,79 @@ public enum Identifiers {
...
+            let content: String
+            if retryCount == 0 {
+                content = path
+            } else {
+                content = "\(path)\0\(retryCount)"
+            }
+
+            let digest = Insecure.MD5.hash(data: content.data(using: .utf8)!)
+            return digest
+                // We can only use the first 22 characters
```
thiagohmcruz, line 122: IIRC only the first 22 chars are relevant in an MD5 and the rest is padding correct? 

That said, why we have to drop the last 10 characters here for the purpose of generating an identifier later on? Wouldn't it work the same way without dropping even if those characters are not "relevant"?

thiagohmcruz (approved):
The test cases answered most of my Qs! Two small clarification Qs still but LGTM

tools/generators/lib/PBXProj/src/Identifiers.swift
brentleyjones: So Xcode identifiers are 24 characters. We are using 2 characters at the front for `"FF"`. That leaves 22 characters that we can use. MD5 generates a digest that is 32 characters long. So we need to truncate it to fit within the remaining 22 characters. We choose the front 22 because are the most unique.

brentleyjones: I'll add this as a comment.

thiagohmcruz: >So Xcode identifiers are 24 characters

Oh ok so this is the bit I was missing, yep a short comment on this would be great 👍

tools/generators/lib/PBXProj/src/Identifiers.swift
```diff
@@ -49,13 +51,79 @@ public enum Identifiers {
...
+        public static func element(
+            _ path: String,
+            type: ElementType,
+            hashCache: inout Set<String>
+        ) -> String {
+            let hash = elementHash(path + type.rawValue, hashCache: &hashCache)
+            return #"FF\#(hash) /* \#(path) */"#
+        }
+
+        private static func elementHash(
```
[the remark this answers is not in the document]
brentleyjones: I added some more comments around this. Let me know if it explains it well enough 🙏.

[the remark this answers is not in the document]
thiagohmcruz: Looks great, thank you for adding that!

PR #2245, opened 2023-05-25
Stop setting `SWIFT_INCLUDE_PATHS` when not needed in BwB mode

brentleyjones:
We were setting it before to support SwiftUI Previews compiling. With the change to use `swift.compile.params`, we can now set it a more efficient way, which fixes a performance degradation. Also, before this change all modes would have double includes.


xcodeproj/internal/xcode_targets.bzl
```diff
@@ -513,41 +513,51 @@ _PREVIEWS_ENABLED_PRODUCT_TYPES = {
...
+        def _handle_swiftmodule_path(file):
+            path = file.path
+            bs_path = xcode_generated_paths.get(path)
+            if bs_path:
+                include_path = paths.dirname(bs_path)
+            else:
+                include_path = file.dirname
+
+            if include_path.find(" ") != -1:
+                include_path = '"{}"'.format(include_path)
```
thiagohmcruz, line 533: Why not always `.format`? Is this a perf related concern?

thiagohmcruz (approved):

xcodeproj/internal/xcode_targets.bzl
brentleyjones: I'm not changing behavior here. But it's not wrapping in quotes when not needed to keep the size of the generated project down.

thiagohmcruz: Just noted similar code that existed previously as well, my bad as I thought this was net new. Regardless, got it, not change in behaviour and it makes sense to keep proj size down. Thx for clarifying!

brentleyjones:
This regresses Xcode's SourceKit `source.request.editor.open.interface` request, which will be fixed in #2256.

mattrobmattrob (approved):

PR #2256, opened 2023-05-30
Set `-I`, `-explicit-swift-module-map-file`, and `-vfsoverlay` in `OTHER_SWIFT_FLAGS`

brentleyjones:
Xcode's SourceKit `source.request.editor.open.interface` request filters a lot of arguments, and the params file in particular. In order to fix this request we need to include `-I` and `-vfsoverlay` arguments in the `OTHER_SWIFT_FLAGS` build setting. We also filter these same arguments in `swift_compiler_params_processor.py` to prevent duplication.

Since `-working-directory` is also filtered for the request, we ensure that relative paths use `$(CURRENT_EXECUTION_ROOT)`.


brentleyjones:
If the vfs overlays have relative paths (will will be the case for almost all of them produced by Bazel), this will still be broken for them. That's always been the case though. I'm looking for a fix.

thiagohmcruz (approved):
Couldn't find any major breakage or net net issue from this, LGTM and if there's more to be fixed IMO it can be addressed in a follow up.

luispadron (approved):

chiragramani (approved):

PR #2264, opened 2023-06-01
Move all `-Xcc` flags into `OTHER_SWIFT_FLAGS`

brentleyjones:


brentleyjones:
I don't plan on merging this, at least not yet. This reverts too much of the optimizations made around memory and cpu. After the incremental generator exists, that one can do this sort of thing much more efficiently.

chiragramani:
This fixes the Jump to symbol definition issue for us as `isystem` is now present in the SourceKit request, successfully completing the lookup done by the module loader.

yongjincho92 (approved):
This change fixes Swift Sourcekit issue on our end with a small effect on Xcode open to editable time. Thanks so much for the fix!

PR #2268, opened 2023-06-02
Add support for raw test_suite rules in custom schemes

thiagohmcruz:
Keeping as a draft for now as I'm looking for some initial feedback (will add comments with questions). Will add a more descriptive PR description and update fixtures after that.

The intent is to expand raw `test_suite` targets to its `tests` actual tests when creating custom schemes, this way one can set such targets in `schemes => xcode_schemes.test_action`


tools/generators/legacy/src/DTO/XcodeScheme.swift
```diff
@@ -36,15 +36,22 @@ extension XcodeScheme {
...
-        get throws {
-            var buildTargets = [BazelLabel: XcodeScheme.BuildTarget]()
-
-            func enableBuildForValue(
-                _ label: BazelLabel,
-                _ keyPath: WritableKeyPath<XcodeScheme.BuildFor, XcodeScheme.BuildFor.Value>,
-                _ mode: EnableBuildForValueMode = .merge
-            ) throws {
+    func withDefaults(
+        aliasLabels: [BazelLabel: [BazelLabel]] = [:]
```
thiagohmcruz, line 40: Making this a `func` that can take `aliasLabels` was the easiest way I've found to expand labels in a single place so the scheme action creations that happen next work as expected

tools/generators/legacy/src/DTO/XcodeScheme.swift
```diff
@@ -61,62 +68,79 @@ disabled, but the target is referenced in the scheme's \(keyPath.actionType) act
...
+        // Create a new build action which includes all of the referenced labels as build targets
+        // We must do this after processing all of the other actions.
+        let newBuildAction = try XcodeScheme.BuildAction(
+            targets: buildTargets.values,
+            preActions: buildAction?.preActions ?? [],
+            postActions: buildAction?.postActions ?? []
+        )
+
+        // Default TestAction
+        let newTestAction: XcodeScheme.TestAction?
```
thiagohmcruz, line 121: This is effectively the new thing in this scope, `testAction?.targets` needs to be expanded if `aliasLabels` is not empty

Compared `buildTargets.values.map { $0.label }` with `testAction?.targets + expansions from aliasLabels` and it's the same list so I assume it's safe to override `targets` with values from `buildTargets` directly here since that's where the expansion happened above in `enableBuildForValue`.

tools/generators/legacy/src/Extensions/XcodeScheme+Extensions.swift
```diff
@@ -63,39 +65,48 @@ add it or a target that depends on it to \(runnerLabel)'s `top_level_targets` at
...
-                    missingLabel: label
-                )
-            )
-            guard labelTargetInfo.isTopLevel else {
-                continue
+
+        for targetLabel in allBazelLabels {
+            var labelsToProcess: [BazelLabel] = [targetLabel]
+            if let actualLabels = aliasLabels[targetLabel], actualLabels.count > 0 {
+                labelsToProcess = actualLabels
```
thiagohmcruz, line 81: Same as the other similar code path with similar logic, we need to expand and run this for each actual label

tools/generators/legacy/src/Extensions/XcodeScheme+Extensions.swift
```diff
@@ -63,39 +65,48 @@ add it or a target that depends on it to \(runnerLabel)'s `top_level_targets` at
...
+
+                    let targetID = best.id
+                    topLevelTargetIDs.insert(targetID)
+                    resolvedTargetIDs[label] = targetID
+                }
             }
         }
 
-        let otherLabels = allBazelLabels.subtracting(topLevelLabels)
+        let otherLabels = allBazelLabels.subtracting(topLevelLabels).subtracting(aliasLabels.keys)
```
thiagohmcruz, line 109: Subtracting labels that were expanded above since those are handled already

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -297,7 +301,33 @@ def _skip_target(
             "{}:{}".format(package_label_str, ctx.rule.attr.generator_name),
         )
 
+    # Set `alias_labels` only for raw `test_suite` rules by checking for the
+    # existence of the `generator_name` attribute. This attribute will be set
+    # in `ios_*_test_suite` rules.
+    if ctx.rule.kind == "test_suite" and not getattr(ctx.rule.attr, "generator_name"):
```
thiagohmcruz, line 307: This is very specific. We only want this for `test_suite` and `generator_name` will be set for `ios_*_test_suite` rules and we don't need to alias labels in that scenario but the rule kind is still `test_suite`. So need a way to differentiate raw `test_suite` and `ios_*_test_suite` here.

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -297,7 +301,33 @@ def _skip_target(
...
+        alias_labels = [
+            struct(
+                alias = "%s" % target.label,
+                labels = depset(
+                    [
+                        "%s" % info.label
+                        for attr, info in transitive_infos
+                        if attr == "tests"
+                    ]
```
thiagohmcruz, line 316: Explicitly finding the labels in `tests` since this is `test_suite` specific, the original intent was was to collect `info.xcode_target.label` from `deps_transitive_infos` but `info.xcode_target` is `None` for those since the targets in `tests` were skipped

xcodeproj/internal/xcodeproj_rule.bzl
```diff
@@ -1166,6 +1166,11 @@ def _write_spec(
             continue
         target_hosts.setdefault(s.hosted, []).append(s.host)
 
+    alias_labels = {
```
thiagohmcruz, line 1169: @brentleyjones we discussed using this in other places e.g. `owned_extra_files`, I can push that next but curious to know if you know of other places we could expand? Or a method to find what else needs to be expanded maybe?

tools/generators/legacy/src/DTO/XcodeScheme.swift
```diff
@@ -36,15 +36,22 @@ extension XcodeScheme {
...
+
+        func enableBuildForValue(
+            _ targetLabel: BazelLabel,
+            _ keyPath: WritableKeyPath<XcodeScheme.BuildFor, XcodeScheme.BuildFor.Value>,
+            _ mode: EnableBuildForValueMode = .merge
+        ) throws {
+            var labelsToProcess: [BazelLabel] = [targetLabel]
+            if let actualLabels = aliasLabels[targetLabel], actualLabels.count > 0 {
+                labelsToProcess = actualLabels
+            }
```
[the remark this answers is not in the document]
brentleyjones: Can `actualLabels.count` ever be 0?

xcodeproj/internal/xcodeproj_rule.bzl
brentleyjones: I think it might only be `owned_extra_files`.

tools/generators/legacy/src/DTO/XcodeScheme.swift
brentleyjones, line 52:
```suggestion
            let labelsToProcess: [BazelLabel]
            if let actualLabels = aliasLabels[targetLabel], actualLabels.count > 0 {
                labelsToProcess = actualLabels
            } else {
                labelsToProcess = [targetLabel]
            }
```

tools/generators/legacy/src/DTO/XcodeScheme.swift
brentleyjones: You'll need to move the `try testAction?.targets.forEach { try enableBuildForValue($0, \.testing) }` line to be after this, so it does it based on `newTestAction`.

tools/generators/legacy/src/DTO/XcodeScheme.swift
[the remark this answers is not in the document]
brentleyjones: (I guess with my suggestion we would want this check anyway.)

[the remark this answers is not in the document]
thiagohmcruz: If the way we collect it works as expected no. This is just me being a bit paranoid maybe. I guess we could improve readability here by removing it since if this ever gets here empty it should be a visible failure in the generator.

tools/generators/legacy/src/Extensions/XcodeScheme+Extensions.swift
```diff
@@ -63,39 +65,48 @@ add it or a target that depends on it to \(runnerLabel)'s `top_level_targets` at
...
-                )
-            )
-            guard labelTargetInfo.isTopLevel else {
-                continue
+
+        for targetLabel in allBazelLabels {
+            var labelsToProcess: [BazelLabel] = [targetLabel]
+            if let actualLabels = aliasLabels[targetLabel], actualLabels.count > 0 {
+                labelsToProcess = actualLabels
             }
```
brentleyjones, line 82:
```suggestion
            let labelsToProcess: [BazelLabel]
            if let actualLabels = aliasLabels[targetLabel], !actualLabels.isEmpty {
                labelsToProcess = actualLabels
            } else {
                labelsToProcess = [targetLabel]
            }
```

xcodeproj/internal/xcodeprojinfo.bzl
brentleyjones: Why do we care if we do this for `ios_*_test_suite`? Is this just an optimization?

tools/generators/legacy/src/DTO/XcodeScheme.swift
thiagohmcruz: Ok, so will be explicit here and use `testAction?.targets` but expanding what is in `aliasLabels` only

xcodeproj/internal/xcodeprojinfo.bzl
thiagohmcruz: Applying the `alias_labels` logic for `ios_*_test_suite` targets used in custom schemes breaks things, since it will try to expand to the runners and for `ios_*_test_suite` what we actually want is to keep the parent `test_suite` as the scheme in Xcode building the test bundle and letting Xcode be the runner.

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -297,7 +301,33 @@ def _skip_target(
...
+    # Set `alias_labels` only for raw `test_suite` rules by checking for the
+    # existence of the `generator_name` attribute. This attribute will be set
+    # in `ios_*_test_suite` rules.
+    if ctx.rule.kind == "test_suite" and not getattr(ctx.rule.attr, "generator_name"):
+        alias_labels = [
+            struct(
+                alias = "%s" % target.label,
+                labels = depset(
+                    [
+                        "%s" % info.label
```
brentleyjones, line 313:
```suggestion
                        str(info.label)
```

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -297,7 +301,33 @@ def _skip_target(
             "{}:{}".format(package_label_str, ctx.rule.attr.generator_name),
         )
 
+    # Set `alias_labels` only for raw `test_suite` rules by checking for the
+    # existence of the `generator_name` attribute. This attribute will be set
+    # in `ios_*_test_suite` rules.
+    if ctx.rule.kind == "test_suite" and not getattr(ctx.rule.attr, "generator_name"):
+        alias_labels = [
+            struct(
+                alias = "%s" % target.label,
```
brentleyjones, line 310:
```suggestion
                alias = str(target.label),
```

tools/generators/legacy/src/DTO/XcodeScheme.swift
thiagohmcruz: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2268/commits/fa2e32834ce3c604564e91ddadc2cc3aea7808df

tools/generators/legacy/src/Extensions/XcodeScheme+Extensions.swift
thiagohmcruz: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2268/commits/fa2e32834ce3c604564e91ddadc2cc3aea7808df

tools/generators/legacy/src/DTO/XcodeScheme.swift
```diff
@@ -36,15 +36,24 @@ extension XcodeScheme {
...
+            _ targetLabel: BazelLabel,
+            _ keyPath: WritableKeyPath<XcodeScheme.BuildFor, XcodeScheme.BuildFor.Value>,
+            _ mode: EnableBuildForValueMode = .merge
+        ) throws {
+            let labelsToProcess: [BazelLabel]
+            if let actualLabels = aliasLabels[targetLabel] {
+                labelsToProcess = actualLabels
+            } else {
+                labelsToProcess = [targetLabel]
+            }
```
luispadron, line 54: nit: 
```suggestion
            let labelsToProcess = aliasLabels[targetLabel, default: [targetLabel]]
```

tools/generators/legacy/src/DTO/XcodeScheme.swift
```diff
@@ -61,62 +70,79 @@ disabled, but the target is referenced in the scheme's \(keyPath.actionType) act
...
+        // We must do this after processing all of the other actions.
+        let newBuildAction = try XcodeScheme.BuildAction(
+            targets: buildTargets.values,
+            preActions: buildAction?.preActions ?? [],
+            postActions: buildAction?.postActions ?? []
+        )
+
+        // Default TestAction
+        let newTestAction: XcodeScheme.TestAction?
+        if aliasLabels.keys.count > 0 {
```
luispadron, line 124: nit: 
```suggestion
        if !aliasLabels.keys.isEmpty {
```

brentleyjones:
@thiagohmcruz Not sure if this is still needed with incremental generation, but if it is, can you open a new PR?

thiagohmcruz:
> @thiagohmcruz Not sure if this is still needed with incremental generation, but if it is, can you open a new PR?

Also not sure, will take a look and open a PR if still needed!

PR #2273, opened 2023-06-05
Properly error out when using an `.xcworkspace`

brentleyjones:
Currently not supported, but building previously would silently fail instead of erroring out.

### DIFF
--- a/xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
+++ b/xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
@@ -167,9 +167,9 @@ https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/new?template=bu
         print(
             f"""\
 error: Failed to determine labels and targets. Note, currently `.xcworkspace`s \
-currently aren't supported. Please make sure you are opening the generated \
+aren't supported. Please make sure you are opening the generated \
 `.xcodeproj` file bundle directly. If you are, and you still get this error, \
-then lease file a bug report here: \
+then please file a bug report here: \
 https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/new?template=bug.md""",
             file = sys.stderr,
         )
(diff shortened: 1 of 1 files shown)


mattrobmattrob (approved):
We've been doing this for ages. 😆

Is this to fix a particular issue?

xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
```diff
@@ -160,8 +163,20 @@ def _calculate_label_and_target_ids(
         )
         return scheme_labels_and_target_ids
 
+    if not labels_and_target_ids:
+        print(
+            f"""\
+error: Failed to determine labels and targets. Note, currently `.xcworkspace`s \
+currently aren't supported. Please make sure you are opening the generated \
```
luispadron, line 170:
```suggestion
aren't supported. Please make sure you are opening the generated \
```

thiagohmcruz, line 170:
```suggestion
error: Failed to determine labels and targets. Note, currently `.xcworkspace`s \
aren't supported. Please make sure you are opening the generated \
```

luispadron (approved):

xcodeproj/internal/bazel_integration_files/calculate_output_groups.py
```diff
@@ -160,8 +163,20 @@ def _calculate_label_and_target_ids(
...
+`.xcodeproj` file bundle directly. If you are, and you still get this error, \
+then lease file a bug report here: \
```
thiagohmcruz, line 172:
```suggestion
then please file a bug report here: \
```

brentleyjones:
> Is this to fix a particular issue?

Fixes an issue if you are using Cocoapods, and Xcode updates the Pods project PIF file last, causing it to get picked up by this script, resulting in the wrong (no) information found.

PR #2274, opened 2023-06-05
Use build configuration from other actions in build before "default"

### LINKED ISSUE #2271: Bug: Builds multiple versions of the target(Debug/Release) when using Release scheme
### Description

When using custom build xcode_configurations and the project cannot create a single target due to difference `deps` based on build_configurations i.e
```
deps  = select({
  ":Debug": ["//Foo"],
  ":Release": ["//Foo", "//Bar"],
})
```

It lists both the targets under Build Action for the `Release` scheme and tries to build both
Screenshot 2023-06-02 at 9 22 10 PM

And since it tris to build `Debug` scheme during release mode so build fails because other `selects` would try to apply `:Release` value but the build is happening for `Debug` Schme. 

### Reproduction steps

Clone this repository - https://github.com/sanju-naik/bazel-demo-github and check out `firebase-sessions-crash-in-opt` branch. 

Generate project by running ` bazel run //:BazelDemo_gen` and open project you would notice build_action has both targets listed.

To simulate build failure uncomment `provisioning_profile` attribute and provide provisioning profile only for the `Release` config. 

### Expected behavior

`BazelDemo-Release` scheme should only list `BazelDemo(Release)` target and it should build only Release config. 

i.e I think We should probably accept `build_configuration` in `xcode_schemes.build_action` and based on this pass targets to build action based on build_config.

### rules_xcodeproj version

1.6.0

### Xcode version

14.2

### Bazel version

5.3.0

### rules_apple version

_No response_

### rules_swift version

_No response_

### Additional information

_No response_

brentleyjones:
Fixes #2271.

This prevents multiple versions of a target from appearing in the build action.


sanju-naik:
Tested this by checking out this branch in my local fork, and can confirm works as expected. Thank you 👍

tools/generators/legacy/test/SetTargetConfigurationsTests.swift
```diff
@@ -57,7 +57,6 @@ final class SetTargetConfigurationsTests: XCTestCase {
         // Assert
 
         XCTAssertNoDifference(pbxTargets, expectedPBXTargets)
-        XCTAssertNoDifference(pbxProj, expectedPBXProj)
```
mattrobmattrob, line 60: This removed because it's now different? Since we're not asserting it anymore, do we need the `try expectedPBXProj.fixReferences()` above?

mattrobmattrob (approved):

tools/generators/legacy/test/SetTargetConfigurationsTests.swift
brentleyjones: I had a comment about this, but it disappeared. For some reason this assert is causing a runtime crash now. Not because they are different, but because of typical odd swiftc miscompiles.

I'm fine with removoing this for now, as I will replace it with something better in the incremental generator. I didn't have time to debug this more now.

PR #2275, opened 2023-06-05
macOS 13.4: Using /usr/bin/swift

chiragramani:
macOS 13.4 swiftc invocations fail when Xcode requests for -v. 
Repro:
```
./App.xcodeproj/rules_xcodeproj/bazel/swiftc -v

Set elements successfully appended to file.
Swift/ErrorType.swift:200: Fatal error: Error raised at top level: Error Domain=NSCocoaErrorDomain Code=4 "The file “swiftc” doesn’t exist." UserInfo={NSFilePath=swiftc}
zsh: trace trap  ./App.xcodeproj/rules_xcodeproj/bazel/swiftc -v
```

### DIFF
--- a/tools/swiftc_stub/main.swift
+++ b/tools/swiftc_stub/main.swift
@@ -103,7 +103,7 @@ func handleSwiftUIPreviewThunk(_ args: [String]) throws {
     else {
         print("warning: No such argument '-sdk'")
         try exit(runSubProcess(
-            executable: "swiftc",
+            executable: "/usr/bin/swiftc",
             args: Array(args.dropFirst())
         ))
     }
@@ -119,7 +119,7 @@ func handleSwiftUIPreviewThunk(_ args: [String]) throws {
     else {
         print("warning: Failed to parse DEVELOPER_DIR from '-sdk'")
         try exit(runSubProcess(
-            executable: "swiftc",
+            executable: "/usr/bin/swiftc",
             args: Array(args.dropFirst())
         ))
     }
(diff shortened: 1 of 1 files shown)


BalestraPatrick:
@chiragramani Is this an Xcode 15 bug or macOS 13.4? I didn't see this issue on Xcode 14.2 on macOS 13.4, but I'm seeing this in Xcode 15 on macOS 13.4, and this fixes it.

brentleyjones:
So it seems that we broke this when translating the stub to Swift. It just happened to never be called until now. When I change this to use `/bin/bash`, so that it can resolve `swiftc` with `PATH`, it resolves to `/usr/bin/swiftc`. `/usr/bin/swiftc` is an `xcrun` shim, so it will consider `DEVELOPER_DIR`... but that isn't set when this is called. So we are instead getting the system `swiftc`, which isn't right.

For example:

```console
$ /usr/bin/swiftc -v
Apple Swift version 5.8 (swiftlang-5.8.0.124.2 clang-1403.0.22.11.100)
Target: arm64-apple-macosx13.0
```

but, it really should be this:

```console
$ /Applications/Xcode-15.0.0-Beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/swiftc -v
Apple Swift version 5.9 (swiftlang-5.9.0.114.6 clang-1500.0.27.1)
Target: arm64-apple-macosx13.0
```

---

I'm looking at how to make it grab the right `swiftc` now.

brentleyjones:
https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2276

PR #2283, opened 2023-06-07
Import indexes in Index Builds as well

brentleyjones:
BwB Index Builds also generate unit files, because we have them use the same build as a normal build in order to get cache hits. If a project benefits from index importing (focused projects in particular), then those generated unit files can be imported. This can be beneficial to getting indexing bootstrapped quicker, since it doesn't require a normal build of every top level target.

A downside to importing during Index Build is that it adds blocking time to the end of those builds. We can't do the normal async importing because then the followup Index Builds will cancel the import of the previous ones.

So far people think this additional time isn't a hinderance, so we are now going to do the additional import.


brentleyjones:
If we go with this change, I have to update the new incremental generators as well.

thiagohmcruz (approved):

luispadron (approved):

mattrobmattrob (approved):
I didn't test on any big project like requested in the Bazel Slack but it LGTM.

Mind adding a bit of context about why you're making this change in the PR description/final commit message?

yongjincho92 (approved):
Measured the time of running synchronous index-import on our repo, and it was adding less than 0.1 seconds per each index build.

PR #2286, opened 2023-06-09
Format error in Xcode when there are no diagnostics parsed

BalestraPatrick:
We sometimes receive complains from our developers due to the fact that a tool invoked through a custom rule or such fails without a descriptive error. Most of the times, this can be controlled by prefixing all error messages with `error: ` when being executed from inside Xcode, but in some cases it's not. For those cases, the current output in Xcode is pretty cryptic:
Screenshot 2023-06-09 at 1 02 46 PM

Many engineers don't seem to be aware that it's possible to reveal more information in the report navigator. For this reason, I propose adding an error message that suggests to do if no error was parsed in the log. The result is the following:
Screenshot 2023-06-09 at 1 02 25 PM

In case there are errors parsed instead (such as a compiler error), we don't do anything.
Screenshot 2023-06-09 at 1 04 26 PM

I'm happy to hear what people think about this change, and if others have encountered a similar problem.

### DIFF
--- a/xcodeproj/internal/bazel_integration_files/process_bazel_build_log.py
+++ b/xcodeproj/internal/bazel_integration_files/process_bazel_build_log.py
@@ -36,6 +36,7 @@ def _main(command: List[str]) -> None:
     relative_diagnostic = re.compile(
         r"^.+?:\d+(:\d+)?: (error|warning): ."
     )
+    has_relative_diagnostic = False
 
     def _replacement(match: re.Match) -> str:
         message = match.group(0)
@@ -75,6 +76,11 @@ def _main(command: List[str]) -> None:
             return
 
         output_line = relative_diagnostic.sub(_replacement, input_line)
+        # Record if we have performed a relative diagnostic substitution.
+        if output_line != input_line:
+            nonlocal has_relative_diagnostic
+            has_relative_diagnostic = True
+
         print(output_line, flush=True)
 
     while process.poll() is None:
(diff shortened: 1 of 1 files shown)


xcodeproj/internal/bazel_integration_files/process_bazel_build_log.py
```diff
@@ -75,6 +76,9 @@ def _process_log_line(line: str):
             return
 
         output_line = relative_diagnostic.sub(_replacement, input_line)
+        if not has_formatted_error and " error: " in output_line:
+            has_formatted_error = True
```
brentleyjones, line 80: Can't we use `relative_diagnostic` in some way to do the substitution and know if we did it? So if we did the substitution, we know that there was an error or warning. Maybe check `input_line != output_line` (if there isn't a better way)?

BalestraPatrick: I thought about it, but the regex also matches warnings, and we only want to show this error if there are no errors at all (regardless if there are warnings or not).

brentleyjones: Hmm. If there are warnings, does it show the weird exit code thing?

BalestraPatrick: I see what you mean, and yes you're right. We can simply check the exit code and see if we have previously done a substitution for a relative diagnostic to see if we need to print the default error message.

brentleyjones (approved):

brentleyjones:
I love little QoL improvements like this :heart:.

thiagohmcruz (approved):

PR #2291, opened 2023-06-14
Handle replacement label when using wrapper macro

### LINKED ISSUE #2289: Bug: Unable to find the `Key` for the `TargetID` when top_level_targets contain targets generated by same macro in same package
### Description

We were upgrading our project to use 1.7.0 but hit the following error:
```
Unable to find the `Key` for the `TargetID`, "//MyPackage:MyTests.__internal__.__test_bundle applebin_ios-ios_x86_64-dbg-ST-4abc478c7114". (function: init(referencedContainer:targets:targetHosts:extensionPointIdentifiers:consolidatedTargetKeys:pbxTargets:), file: tools/generators/legacy/src/Generator/TargetResolver.swift, line: 51)
```

I did some digging and found that the regression is introduced by [Fix tests in custom scheme when ios_unit_test.bundle_name is used](https://github.com/MobileNativeFoundation/rules_xcodeproj/commit/f335ff2dd12c3ce71f32d20920a4e5c427a04efa) . This [change](https://github.com/MobileNativeFoundation/rules_xcodeproj/commit/f335ff2dd12c3ce71f32d20920a4e5c427a04efa#diff-31ddaae3d278246b432650a6d8c1b1a4e6940ce58ae5df7fae0c16fb00ebf758R290-R293) switch to using `generator_name` to construct target replacement label. However, some of our test targets are expanded under the same macro (have same `generator_name`), which cause those targets to have identical target replacement label. This cause the error to be thrown when we generate project that contains multiple test targets with same `generator_name` and package in `top_level_targets`.

### Reproduction steps

See repro diff at https://github.com/David-Lam/rules_xcodeproj/pull/1/files.

Checkout [dlam-reproKeyError](https://github.com/David-Lam/rules_xcodeproj/tree/dlam-reproKeyError) at https://github.com/David-Lam/rules_xcodeproj then run:
```
cd examples/rules_ios/
bazel run //:xcodeproj
```

### Expected behavior

`bazel run //:xcodeproj` should complete successfully. 

### rules_xcodeproj version

1.7.0

### Xcode version

14.1

### Bazel version

6.2.0

### rules_apple version

_No response_

### rules_swift version

_No response_

### Additional information

_No response_

thiagohmcruz:
Fixes #2289

- [x] TODO: Add a test to exercise this (wrap one test in a macro)

See comment in code. Mostly based on the `rules_apple` logic here: https://github.com/bazelbuild/rules_apple/blob/c628b0e4b3261b4627bcd4e9f91e75b4e5eb0e1a/apple/internal/testing/apple_test_assembler.bzl#L100-L125


xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -293,8 +293,26 @@ def _skip_target(
...
+        # are two options:
+        #
+        # 1. `ctx.rule.attr.runner` is rules_apple's default and we can use
+        # `ctx.rule.attr.name` as is
+        # 2. `ctx.rule.attr.runner` is custom and `ctx.rule.attr.name` has
+        # the pattern "{target_name}_{runner_name}"
+        #
+        # In both scenarios if we remove the "_{runner_name}" suffix we
+        # get the label_name that we need.
+        if ctx.rule.attr.generator_function != "ios_unit_test":
```
thiagohmcruz, line 310: Need to check if this is generic enough, will test with other test specific rule kinds

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -293,8 +293,26 @@ def _skip_target(
         # `bundle_name` can be used to name the bundle instead of the
         # target name. Because of that we use `ctx.rule.attr.generator_name`
         # here to ensure this is always a real target label.
+        label_name = ctx.rule.attr.generator_name
+
+        # If `generator_function` is not `ios_unit_test` it means this target
+        # is wrapped in a macro and `generator_name` does not hold the value
```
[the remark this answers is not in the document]
thiagohmcruz: only tests due to this above
```py
        if target_skip_type != skip_type.apple_test_bundle:
            return target.label
```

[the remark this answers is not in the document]
mattrobmattrob: How did I miss that. 🤦

xcodeproj/internal/xcodeprojinfo.bzl
thiagohmcruz: `runners` is at the `ios_*_test_suite` level when this code is running we're always dealing with the underlying targets which contain `rules_apple`'s internal test bundle as a dep and `runner` set (to default or a custom one defined by the user).

When a test is wrapped in macro called `macro_foo` then `generator_function` value will be `macro_foo` and not `ios_unit_test`. Which makes me think what we want here is 
```py
ctx.rule.attr.generator_function in ['ios_unit_test', ... <other known test kinds>]
```

Does that addresses your Q?

mattrobmattrob: And that's mostly my round about suggestion, being explicit here about the `rules_apple` things we're trying to handle vs. a `!=` test since it'll be quite generous. And mostly thinking about things like `ios_build_test` and so on like you mention.

examples/rules_ios/iOSApp/Test/ObjCUnitTests/macros.bzl
```diff
@@ -0,0 +1,23 @@
+"""Exercise targets wrapped in macros"""
+
+load(
+    "@build_bazel_rules_ios//rules:test.bzl",
+    rules_ios_ios_unit_test = "ios_unit_test",
+)
+
+def macro_1(name):
```
luispadron, line 8: Nit: 
```suggestion
def rules_ios_unit_test_macro(name):
```

thiagohmcruz: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2291/commits/87002db8a04b95285286dfc9e5778880fcb7b938

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -289,12 +289,40 @@ def _skip_target(
...
+        # 2. Use `ctx.rule.attr.generator_name`
+        #
+        # In many scenarios `generator_name` would be enough since it holds the
+        # name of the test rule generating it but if the test
+        # targets are wrapped in macros then `generator_name` can be anything
+        #
+        # Example:
+        #
+        # `iOSAppObjCUnitTestSuite_iPhone-13-Pro__16.2` => `iOSAppObjCUnitTestSuite`
+        #
```
thiagohmcruz, line 320: @mattrobmattrob @luispadron @brentleyjones this summarizes the approach and my reasoning, let me know if it makes sense and if it addresses the other review comments or not

xcodeproj/internal/xcodeprojinfo.bzl
thiagohmcruz: Let's continue on the other thread: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2291/files#r1232615504

(ps: `ios_build_test` and similar are not hit here since they're skipped earlier while targets are being processed, so this should be safely running only for `*_(unit|ui)_test` rules only as all of them respect the patterns I mentioned in the comment added to the code above)

xcodeproj/internal/xcodeprojinfo.bzl
```diff
@@ -289,12 +289,40 @@ def _skip_target(
         label_str = bazel_labels.normalize_label(info.xcode_target.label)
         package_label_str = label_str.split(":")[0]
 
+        # Important notes:
+        #
+        # 1. This relies on implementation details of
+        # https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/testing/apple_test_assembler.bzl#L100-L125
```
thiagohmcruz, line 295: I know we want columns of size `80` but at the same time I think it's important to link to the reference code path?

brentleyjones: Correct, links don't need to wrap, they just need to start on their own line if they would wrap.

luispadron (approved):

mattrobmattrob (approved):

PR #2306, opened 2023-06-23
Add support for params files to partial generators

brentleyjones:


luispadron (approved):

tools/generators/lib/GeneratorCommon/src/ParsableCommand+Extensions.swift
```diff
@@ -0,0 +1,34 @@
+import ArgumentParser
+import Foundation
+
+extension ParsableCommand {
+    public static func parseAsRootSupportingParamsFiles() async {
```
thiagohmcruz, line 5: Is there any need to exercise this and explicitly assert in a test that `parseAsRootSupportingParamsFiles` is parsing things correctly?

thiagohmcruz (approved):

tools/generators/lib/GeneratorCommon/src/ParsableCommand+Extensions.swift
brentleyjones: Since it's pretty small and understandable, and to make it testable would be a pain, and since this made my generator call that used a params file work, I think that's good enough for me.

brentleyjones: I'm fine with someone else adding a test though. I just realized we can create temp files like how we do for `ResolveSymlinksTests`, and we can create a mock `ParsableCommand`.

PR #2321, opened 2023-06-28
Add `Platform.OS` to `PBXProj`

brentleyjones:


tools/generators/lib/PBXProj/src/Platform.swift
```diff
@@ -1,6 +1,13 @@
 import ArgumentParser
 
 public enum Platform: String, ExpressibleByArgument {
+    public enum OS: Comparable {
```
thiagohmcruz, line 4: Just curious, is visionOS going to be a separate platform like this? Haven't looked into anything about it yet lol

mattrobmattrob (approved):

tools/generators/lib/PBXProj/src/Platform.swift
brentleyjones: Yes

PR #2322, opened 2023-06-28
Adjust `Identifier` number scheme

brentleyjones:
This is to make way for the next set of generators.

### DIFF
--- a/tools/generators/lib/PBXProj/src/Identifiers.swift
+++ b/tools/generators/lib/PBXProj/src/Identifiers.swift
@@ -6,45 +6,75 @@ import CryptoKit
 /// Since the various `PBXProj` partial generators need to work independently
 /// from each other, they need help to be able to generate unique identifiers.
 /// `Identifiers` is used to generate these identifiers.
+///
+/// The identifiers aren't all hashes. Grouping information is also encoded into
+/// them. The first two characters indicate that it's a global element (`FF`), a
+/// file or group element (`FE`), or a generator shard (`00`-`FD`).
+///
+/// - Global elements (which start with `FF`) are then classified as belonging
+///   to the project as a whole (`00`) or the `BazelDependencies` target
+///   (`01`). For both of them, their associated elements (e.g. configuration
+///   list, build phases, etc.) use
+///   `00000000000000000000`-`000000000000000000FF`, and their configurations
+///   (e.g. "Debug", "Release", etc.) use
+///   `00000000000000000100`-`000000000000000001FF`.
+///
+/// - Files and groups (which start with `FE`) use the remaining characters
+///   (`0000000000000000000000`-`FFFFFFFFFFFFFFFFFFFFFF`).
+///
+/// - Generator shards (which start with `00`-`FD`) use the next two
+///   characters to indicate a target and its associated elements (`00`), a
+///   target dependency (`01`), a container item proxy (`02`), or a build file
+///   (`FF`).
+///
+///   - Targets (which start with `xx00`) use the next eight characters to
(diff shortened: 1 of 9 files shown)


tools/generators/lib/PBXProj/src/Identifiers.swift
```diff
@@ -6,45 +6,74 @@ import CryptoKit
...
 /// from each other, they need help to be able to generate unique identifiers.
 /// `Identifiers` is used to generate these identifiers.
+///
+/// The identifiers aren't all hashes. Grouping information is also encoded into
+/// them. The first two characters indicate that it's a global element (`FF`), a
+/// file or group element (`FE`), or a generator shard (`00`-`FD`).
+///
+/// - Global elements (which start with `FF`) are then classified as belonging
+///   to the project as a whole (`00`) or the `BazelDependencies` target
+///   (`01`). For both of them, their associated elements (e.g. configuration
```
mattrobmattrob, line 16: I don't immediately understand this portion.

Is it `FF...01` & `FF...00`, `FF01...` & `FF00...`, or something else?

tools/generators/lib/PBXProj/src/Identifiers.swift
```diff
@@ -6,45 +6,74 @@ import CryptoKit
...
+///   list, build phases, etc.) use `000000000000000000`-`0000000000000000FF`,
+///   and their configurations (e.g. "Debug", "Release", etc.) use
+///   `000000000000000100`-`0000000000000001FF`.
```
mattrobmattrob, line 19: Are these missing two digits? Seems the partial tests identifiers are 24 characters and these with `FF` & `00`/`01` are only 22 characters.

tools/generators/lib/PBXProj/src/Identifiers.swift
```diff
@@ -6,45 +6,74 @@ import CryptoKit
...
+///     associated elements (e.g. configuration list, build phases, etc.), and
+///     `000000000100`-`0000000001FF` for its configurations (e.g. "Debug",
+///     "Release", etc.).
+///
+///   - Target dependencies (which start with `xx01`) and container item
+///     proxies (which start with `xx02`) use the next eight characters to
+///     identify the dependent target with a target sub-identifier
+///     (`00000000`-`FFFFFFFF`). Then it uses the remaining characters to
+///     identify the dependency target with the shard and target sub-identifier
+///     (`xx0000000000`-`xx00FFFFFFFF`).
```
mattrobmattrob, line 41: So:

Prefix: `xx01` or `xx02` (4 characters)
Target sub-identifier: `00000000`-`FFFFFFFF` (8 characters)
Dep target: `xx0000000000`-`xx00FFFFFFFF` (12 characters)

Combined: `{Prefix}{Target sub-identifier}{Dep target}` (24 characters)

What do the `xx` in dep target represent?

mattrobmattrob (approved):
LGTM. Just some requests to tighten up the comments in `Identifiers.swift` for maximum clarity. 👍

tools/generators/lib/PBXProj/src/Identifiers.swift
brentleyjones: `FF00...` and `FF01...`.

I'm 100% open to rewording. This applies to all of your comments.

tools/generators/lib/PBXProj/src/Identifiers.swift
brentleyjones: Correct, they are off by `00`. Will fix.

mattrobmattrob:
I'm all good with the code comments. We can revisit if they cause any confusion. 👍

luispadron (approved):

PR #2323, opened 2023-06-29
Add `Identifiers.Targets`

brentleyjones:


tools/generators/lib/PBXProj/src/BuildPhase.swift
```diff
@@ -0,0 +1,31 @@
...
+    case embedAppExtensions
+    case embedAppClips
+    
+    public var name: String {
+        switch self {
+        case .bazelIntegration: return """
+Copy Bazel Outputs / Generate Bazel Dependencies (Index Build)
+"""
+        case .createCompileDependencies: return "Create compiling dependencies"
+        case .createLinkDependencies: return "Create linking dependencies"
```
mattrobmattrob, line 20: [nit] Just to match the rest:
```suggestion
        case .createCompileDependencies: return "Create Compiling Dependencies"
        case .createLinkDependencies: return "Create Linking Dependencies"
```

mattrobmattrob (approved):

tools/generators/lib/PBXProj/src/BuildPhase.swift
brentleyjones: Not going to take this right now, as I'm copying the existing implementation. I want to take a pass on all of them later though, which I can after this PR lands.

thiagohmcruz (approved):

PR #2329, opened 2023-06-30
Upgrade rules_swift to 1.9.0

mattrobmattrob:
Ignore my poorly named branch.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -8,7 +8,7 @@ module(
 bazel_dep(name = "bazel_skylib", version = "1.3.0")
 bazel_dep(
     name = "rules_swift",
-    version = "1.6.0",
+    version = "1.9.0",
     repo_name = "build_bazel_rules_swift",
 )
 bazel_dep(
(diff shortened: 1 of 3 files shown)


MODULE.bazel
```diff
@@ -8,7 +8,7 @@ module(
 bazel_dep(name = "bazel_skylib", version = "1.3.0")
 bazel_dep(
     name = "rules_swift",
-    version = "1.6.0",
+    version = "1.9.0",
```
brentleyjones, line 11: Do we dep on something in this version? If not, we should only bump the workspace version. bzlmod deps should be the lowest version we support in order to support MSV.

mattrobmattrob: Good point. Was just doing it to include https://github.com/bazelbuild/rules_swift/commit/67eb075c8661ca2c2b7c2c0f2fa783caf67f3667 but will make that change in:

https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/de5a257717960598d3648084e1dd4a88a2153cac/xcodeproj/repositories.bzl#L144-L165

mattrobmattrob: Actually, realize it's the same version and the comment is out of date. 👍

PR #2337, opened 2023-07-07
Restrict DEVELOPER_DIR regex to the start of the line

keith:
All these patterns should only only be replaced at the start of the line since they should always be replacing absolute paths to Xcode. This theoretically improves the regex performance but more importantly should make the regex slightly more clear.


brentleyjones (approved):
I just want to make sure you tested that it still works :+1:

keith:
yep, can verify adding this works, and adding `^foo` for sure breaks it

PR #2343, opened 2023-07-07
Set Project Format to `Xcode 15` when `minimum_xcode_version` is 15+

brentleyjones:


thiagohmcruz (approved):

luispadron (approved):

tools/generators/legacy/src/Generator/CreateProject.swift
```diff
@@ -213,3 +213,13 @@ Xcode \(min(project.minimumXcodeVersion.major, 14)).0
         return pbxProj
     }
 }
+
+private extension SemanticVersion {
+    var pbxProjObjectVersion: UInt {
+        switch major {
+            case 15: return 60
+            case 14: return 56
+            default: return 55 // Xcode 13
```
brentleyjones, line 222: I'll change this to use ranges instead, so that it doesn't break on Xcode 16 🤦.

PR #2360, opened 2023-07-13
Fix order of flags in `swift_debug_settings.py`

brentleyjones:
The compiler flags for dependencies should come before dependents.


brentleyjones:
I'm going to compare something against Xcode, since it might be that just the flags need to be this way, but includes should be the current way.

thiagohmcruz (approved):
Worth doing the comparison you mentioned indeed, but overall LGTM!

luispadron (approved):

brentleyjones:
Yeah, seems the previous order was actually correct for flags, but backwards for frameworks. My next change will fix the frameworks order.

PR #2368, opened 2023-07-18
Evaluate the product basename based on `$BAZEL_OUTPUTS_PRODUCT`

gferon:
We discovered this problem when using a `select()` on the `bundle_name` of a `ios_application` where the condition comes from a `build_configuration` transition.

Without this patch, the build phase "Copy Bazel Outputs / Generate Bazel dependencies" gets a script with hardcoded value for the basename of the output, which actually _might_ vary dependending on the build configuration (when `bundle_name` or `executable_name` is dynamically evaluated).

This patch:
- Fixes the issue by using an environment variable available to the script.
- Adds three `config_setting` to actually make use of the `build_configuration` transition and use `select` with `bundle_name`.

We would greatly appreciate any feedback you may have. Thank you!

/cc @ohitsdaniel


examples/integration/BUILD
```diff
@@ -25,6 +25,27 @@ string_flag(
...
+)
+
+config_setting(
+    name = "build_b",
+    flag_values = {
+        ":flag_to_transition_on": "B",
+    },
+)
+
+config_setting(
```
brentleyjones, line 42: f

examples/integration/BUILD
```diff
@@ -77,8 +98,10 @@ string_flag(
...
         xcode_configurations = XCODE_CONFIGURATIONS,
     )
-    for simulator_cpu in [
-        "sim_arm64",
-        "x86_64",
+    for simulator_cpu, build_mode in [
+        ("sim_arm64", "xcode"),
+        ("sim_arm64", "bazel"),
+        ("x86_64", "xcode"),
+        ("x86_64", "bazel"),
```
brentleyjones, line 105: Can you make this change in its own PR to land before this?

gferon: Absolutely, I suppose maybe CI has to be changed as well?

brentleyjones (changes requested):
Good find, and nice fix!

examples/integration/BUILD
brentleyjones: No, it already does this for fixtures. the CI is blocked by you not being a regular contributor. I can unblock that.

brentleyjones:
In the root of the repository, can you run `./test/update_all_fixtures.sh` and push those changes? That will adjust the test fixtures for the changes you made.

gferon:
> In the root of the repository, can you run `./test/update_all_fixtures.sh` and push those changes? That will adjust the test fixtures for the changes you made.

Done, my MacBook ended up doing an _actual_ TestFlight to regenerate fixtures 😄

tools/generators/legacy/src/Generator/AddTargets.swift
```diff
@@ -192,27 +192,17 @@ Product for target "\(key)" not found in `products`
...
-else
-  "$BAZEL_INTEGRATION_DIR/copy_outputs.sh" \
-    "\#(Generator.bazelForcedSwiftCompilePath)" \
-    "\#(bazelBasename)" \
-    "\#(productType.rsyncExcludeFile)"
-"""#
-        } else {
-            copyOutputs = ""
-        }
-
```
brentleyjones, line 207: We still need this conditional. I think instead of the full remove like I suggested (sorry) we can change `outputs.productBasename` to `outputs.hasProductOutput` (similar to the `hasSwiftOutputs`.

gferon: I have a question though: isn't checking `output.productBasename` the same as checking whether `BAZEL_OUTPUTS_PRODUCT` is set? I guess there's a subtle difference but I couldn't find any.

brentleyjones: Hmm. Yeah. I guess we can unconditionally run the script. It was probably an optimization we don't need.

gferon: Ok, then this change should be good to go. Are any other changes needed?

brentleyjones: Not that I know of. I need about 15 mins before I can start to review this again. It’s probably going to be merged as is though.

examples/cc/test/fixtures/bwb.xcodeproj/project.pbxproj
```diff
@@ -696,7 +696,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "set -euo pipefail\n\nif [[ \"$ACTION\" == \"indexbuild\" ]]; then\n  cd \"$SRCROOT\"\n\n  \"$BAZEL_INTEGRATION_DIR/generate_index_build_bazel_dependencies.sh\"\n\nfi\n";
+			shellScript = "set -euo pipefail\n\nif [[ \"$ACTION\" == \"indexbuild\" ]]; then\n  cd \"$SRCROOT\"\n\n  \"$BAZEL_INTEGRATION_DIR/generate_index_build_bazel_dependencies.sh\"\nelse\n  \"$BAZEL_INTEGRATION_DIR/copy_outputs.sh\" \\\n    \"_BazelForcedCompile_.swift\" \\\n    \"\"\nfi\n";
```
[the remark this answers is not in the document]
brentleyjones: Per other comment: this is fine. The check in the script prevents additional work. We might want to optimize more later, but this has so little overhead, it's probably fine.

brentleyjones (approved):

brentleyjones:
Errors from CI:

```
ERROR: /Users/administrator/buildbuddy.noindex/remote_build/e8cccc0a-f136-43e0-83c4-0b60991cb5a9/repo-root/examples/integration/iOSApp/Source/BUILD:26:16: configurable attribute "bundle_name" in //iOSApp/Source:iOSApp doesn't match this configuration. Would a default condition help?
Conditions checked:
 //:build_aaaaaaa
 //:build_b
 //:build_c
```

And probably unrelated to your change, but we might have some non-determinism:

```diff
--- a/examples/integration/test/fixtures/bwb.xcodeproj/xcshareddata/xcschemes/iOSAppUnitTests_Scheme.xcscheme
+++ b/examples/integration/test/fixtures/bwb.xcodeproj/xcshareddata/xcschemes/iOSAppUnitTests_Scheme.xcscheme
@@ -109,7 +109,7 @@
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "1765D91CC168F25A5BC51603"
-               BuildableName = "iOSApp.C.app"
+               BuildableName = "iOSApp.AAAAAAA.app"
                BlueprintName = "iOSApp"
                ReferencedContainer = "container:../../../../../test/fixtures/bwb.xcodeproj">
             </BuildableReference>
```

Can you another PR that does the target based changes, so we iterate on those separately? And then have this PR just be the fix and updated fixes from the fix alone?

brentleyjones (approved):

tools/generators/legacy/src/Generator/AddTargets.swift
```diff
@@ -192,27 +192,17 @@ Product for target "\(key)" not found in `products`
...
         let shellScript = #"""
 set -euo pipefail
 
 if [[ "$ACTION" == "indexbuild" ]]; then
   cd "$SRCROOT"
 
   "$BAZEL_INTEGRATION_DIR/generate_index_build_bazel_dependencies.sh"
-\#(copyOutputs)
```
brentleyjones, line 215: Because of this change, you'll still need to run `./test/update_all_fixtures.sh` again.

brentleyjones:
> Hope you will squash this monster of a branch 👻 when mergin git.

Yeah, they are all squash commits on merge :+1:.

examples/integration/iOSApp/Source/BUILD
```diff
@@ -29,7 +29,11 @@ ios_application(
     app_clips = ["//AppClip"],
     app_icons = glob(["Assets.xcassets/AppIcon.appiconset/**"]),
     bundle_id = IOS_BUNDLE_ID,
-    bundle_name = "iOSApp",
+    bundle_name = select({
+        "@//:build_aaaaaaa": "iOSApp.AAAAAAA",
+        "@//:build_b": "iOSApp.B",
+        "@//:build_c": "iOSApp.C",
+    }),
```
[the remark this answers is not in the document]
gferon: Ok, I _think_ it should be good now, sorry for the back and forth 😅

brentleyjones (approved):

brentleyjones:
Thanks! Please open the PR with the project change as well now, minimally so we can look into why it was failing on CI.

PR #2375, opened 2023-07-18
Fix `copy_outputs.sh` for archives

brentleyjones:
Regressed with fa0f92e9dccc48da847604633fb1fdb9f0075aa7.

The names of these variables are confusing (`BAZEL_OUTPUTS_PRODUCT` probably should indicate that it can be the name of the archive). We can address that later.


test/fixtures/generator/bwb.xcodeproj/project.pbxproj
```diff
@@ -4215,6 +4215,7 @@
 				BAZEL_LABEL = "@//tools/generators/legacy/test:tests";
 				BAZEL_OUTPUTS_DSYM = "\"bazel-out/applebin_macos-darwin_x86_64-opt-STABLE-4/bin/tools/generators/legacy/test/tests.xctest.dSYM\"";
 				BAZEL_OUTPUTS_PRODUCT = "bazel-out/applebin_macos-darwin_x86_64-opt-STABLE-4/bin/tools/generators/legacy/test/tests.__internal__.__test_bundle.zip";
+				BAZEL_OUTPUTS_PRODUCT_BASENAME = tests.xctest;
```
brentleyjones, line 4218: This change shows how the old way would be `tests.__internal__.__test_bundle.zip` but now is `tests.xctest`.

PR #2387, opened 2023-07-22
Placing generated Core Data assets into a resource bundle target creates invalid references in BwX mode

wileykestner:
Moving _generated_ Core Data `.xcdatamodeld` assets into a Resource Bundle causes four (4) references to appear in the Copy Resource Bundle build phase:

Screenshot 2023-07-22 at 7 12 36 PM

None of these references appear valid.

Steps to repro:

1. `cd examples/integration`
2. `bazel run //:xcodeproj-xcode-sim_arm64`
3. Open the project and run the `iOS App (Simulator)` scheme (or run `xcodebuild build -project Integration.xcodeproj -scheme 'iOSApp (Simulator)'`)


brentleyjones:
I'll dig in soon, but I would expect 4, for each configuration. And then it would use other means to exclude them (which I think might have broke).

brentleyjones:
I found a couple bugs (which I have PRs coming for), but then I also ran into a roadblock. Seems that Xcode is ignoring `EXCLUDED_SOURCE_FILE_NAMES` for `.xcdatamodeld` files (even after I changed it to use `$(BAZEL_OUT)` and the right path). So if you have multiple configurations, even just device and simulator, this breaks.

Now, I can fix that by preventing bundles that have conditional `.xcdatamodeld` from consolidating. This would be only BwX mode, and the end result is that you would now have specific target environment and Xcode configuration flavors. So in this repro we would have 4 `ExampleResources`, `iOSApp`, etc. targets/schemes. I'm not sure that is worth the effort though.

brentleyjones:
I haven't submitted another PR, which would add `.xcdatamodel` files to `important_generated`, which is also needed, and will result in those files being generated when generating the project (as Xcode needs to see them before a build can successfully be kicked off). But regardless, you will get stuck at the same point as me if you have any other configurations.

brentleyjones:
BwX mode is going away, so closing.

PR #2406, opened 2023-07-27
Add some structured concurrency to the `files_and_groups` generator

brentleyjones:
This has a very minor effect currently, so most of the computation happens before these async tasks. A future change will add in more file reading and writing related to `PBXBuildFile` creation, which will benefit from these changes.


luispadron (approved):
neat

thiagohmcruz (approved):

PR #2407, opened 2023-07-27
Use the module name as the default target name

CognitiveDisson:
The name of the product module name should be used as a better source for the Xcode target than the target part of the label. This is because the target part of the label does not necessarily have to be unique; only the complete label serves as the unique identifier of the target.

For this reason, rules_xcodeproj should use the name of the product module for disambiguation and as the Xcode target name if it is both set and unique. If it is not unique or not set, we must fall back on the old logic of using the target name.

Here is a before and after for `examples/rules_ios`:

| Before | After |
| - | - |
| Before  | After |


tools/generators/legacy/src/Generator/ConsolidateTargets.swift
```diff
@@ -354,7 +354,7 @@ extension ConsolidatedTarget {
         self.targets = targets
         let aTarget = self.targets.first!.value
 
-        name = aTarget.name
+        self.name = aTarget.product.moduleName ?? aTarget.name
```
brentleyjones, line 357: Since each target could have different module names (because of `select()`), how should we handle that? I think we have three options:

1. They can't be consolidated
2. We fall back to the target name
3. We do like what you have, just using the first (sorted)

I don't like 1, but 3 or 2 seem fine to me?

xcodeproj/internal/library_targets.bzl
```diff
@@ -54,10 +54,11 @@ def process_library_target(
     build_settings = {}
 
     product_name = ctx.rule.attr.name
+    product_module_name = get_product_module_name(ctx = ctx, target = target)
```
brentleyjones, line 57: This will call `derive_module_name()` for Swift targets that don't set a custom module name. We don't want to send that to the generator.

How about we change `get_product_module_name` to return two values: the actual attribute, and the build setting value? We would set the build setting value like we have now, but we would use the actual attribute in `ConsolidatedTarget`.

tools/generators/legacy/src/Generator/ConsolidateTargets.swift
CognitiveDisson: I think we should use the target name as a fallback ( option 2 ). So if the module name is unique we use it, if not we use the target and if that doesn't help we use the label.

brentleyjones: Can you make that change? You would collect all of the module names in this init, and if all are non-nil and the same, use it, otherwise use the target.name like before.

CognitiveDisson: I implemented this way, based on fixtures works as expected. The implementation is a bit ugly, if you see how to improve it let me know.

xcodeproj/internal/library_targets.bzl
CognitiveDisson: I thought about creating a separate method to retrieve only the value of an attribute, but that would require executing `getattr` twice. I assume that this method has a performance cost, so I decided to follow your suggestion.

tools/generators/legacy/src/Generator/ConsolidateTargets.swift
```diff
@@ -392,6 +391,15 @@ extension ConsolidatedTarget {
                     rhsTarget.buildSettingConditional
             }
             .map { $1 }
+
+        if let moduleName = aTarget.product.moduleName,
+           sortedTargets.filter({ $0.product.moduleName == moduleName }).count == 1
```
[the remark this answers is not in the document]
CognitiveDisson: It's not the same. I checked that the product name only appears once, and your suggestion checks that there is no target with a different name. Additionally, it could behave incorrectly if some of the targets have nil as the module name.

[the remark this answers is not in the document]
CognitiveDisson: We need loop through all the elements of the array to see if there are any duplicates.

thiagohmcruz (approved):
We use `rules_ios` and project generation is 🟢 for us in this branch. I can't vet this in more detail at this time to see if the fact that the devs will see different names in Xcode is the only effective change for us or if there's any other "behaviour" change I'm not seeing here.

That said, not a blocker but maybe folks using `rules_ios` need to vet carefully their internal assumptions before bumping to this.

PR #2409, opened 2023-07-28
Add `Identifiers.BuildFiles`

brentleyjones:


tools/generators/lib/PBXProj/src/Identifiers.swift
```diff
@@ -81,6 +83,161 @@ FF01000000000000000001\#(String(format: "%02X", index)) \#
...
+            retryCount: Int
+        ) -> String {
+            let content: String
+            if retryCount == 0 {
+                content = hashable
+            } else {
+                content = "\(hashable)\0\(retryCount)"
+            }
+
+            let digest = Insecure.MD5.hash(data: content.data(using: .utf8)!)
```
jszumski, line 227: This should let you avoid the force unwrap

```suggestion
            let digest = Insecure.MD5.hash(data: Data(content.utf8))
```

tools/generators/lib/PBXProj/src/Identifiers.swift
```diff
@@ -81,6 +83,161 @@ FF01000000000000000001\#(String(format: "%02X", index)) \#
...
+            return digest
+                // Xcode identifiers are 24 characters. We are using 4
+                // characters at the front. That leaves 20 characters that we
+                // can use. MD5 digests are 16 bytes (32 characters)
+                // long. So we need to truncate it to fit within the remaining
+                // 20 characters (by dropping 6 bytes). We choose the front 20
+                // because they are the most unique.
```
jszumski, line 234: > We choose the front 20 because they are the most unique.

Definitely a nit but aren't all parts of a cryptographic hash uniformly unique?

tools/generators/lib/PBXProj/src/Identifiers.swift
```diff
@@ -386,3 +559,80 @@ extension BuildPhase {
...
+
+        try data.write(to: url)
+    }
+
+    private func encode(into data: inout Data) {
+        data.append(type.rawValue.data(using: .utf8)!)
+        data.append(shard.data(using: .utf8)!)
+        data.append(hash.data(using: .utf8)!)
+        data.append(path.isFolder ? Self.isFolder : Self.notIsFolder)
+        data.append(path.path.data(using: .utf8)!)
```
jszumski, line 588: You can use the same `Data` construction here to avoid the force unwraps

jszumski (approved):

tools/generators/lib/PBXProj/src/Identifiers.swift
brentleyjones: I thought I read somewhere that it impacted the entropy, but now I'm reading the opposite. I'll remove that verbiage.

thiagohmcruz (approved):

PR #2411, opened 2023-07-28
Use `Data.init` instead of `.data(using:)`

brentleyjones:
Prevents force unwraps.

### DIFF
--- a/tools/generators/lib/PBXProj/src/ConsolidationMapEntry.swift
+++ b/tools/generators/lib/PBXProj/src/ConsolidationMapEntry.swift
@@ -48,7 +48,7 @@ extension ConsolidationMapEntry {
     }
 
     func encode(into data: inout Data) {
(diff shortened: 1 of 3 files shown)


jszumski (approved):

tools/generators/lib/PBXProj/src/ConsolidationMapEntry.swift
```diff
@@ -48,7 +48,7 @@ extension ConsolidationMapEntry {
     }
 
     func encode(into data: inout Data) {
-        data.append(name.data(using: .utf8)!)
+        data.append(Data(name.utf8))
```
luispadron, line 51: Nice, TIL this gets rid of the optional

luispadron (approved):

mattrobmattrob (approved):

PR #2413, opened 2023-07-28
Write `PBXBuildFile`s in `files_and_groups` generator

brentleyjones:


brentleyjones:
There are some FIXME's, which I need to address, so moving to draft, but I would like reviews to start now if possible 🙏.

tools/generators/files_and_groups/src/Generator/CreateShardBuildFileElements.swift
```diff
@@ -0,0 +1,79 @@
...
+            readBuildFileSubIdentifiersFile(buildFileSubIdentifierFile)
+
+        let fileIdentifiers = try await fileIdentifiersTask.value
+
+        return try subIdentifiers.map { subIdentifier in
+            return createBuildFileElement(
+                subIdentifier: subIdentifier,
+                fileIdentifier: try fileIdentifiers.value(
+                    for: subIdentifier.path,
+                    context: "Path"
```
thiagohmcruz, line 74: Maybe a more descriptive `context`?

thiagohmcruz:
First pass LGTM, I can take another look once the `FIXME`s are in!

tools/generators/files_and_groups/src/Generator/CreateBuildFileElements.swift
```diff
@@ -0,0 +1,71 @@
...
+        /// Reads `[Identifiers.BuildFile.SubIdentifier]` from disk.
+        func callAsFunction(
+            buildFileSubIdentifierFiles: [URL],
+            fileIdentifiersTask: Task<[BazelPath: String], Error>
+        ) async throws -> [BuildFileElement] {
+            try await callable(
+                /*buildFileSubIdentifierFiles:*/ buildFileSubIdentifierFiles,
+                /*fileIdentifiersTask:*/ fileIdentifiersTask,
+                /*createShardBuildFileElements:*/ createShardBuildFileElements
+            )
```
luispadron, line 31: Q: why not use parameter names here?

luispadron (approved):

tools/generators/files_and_groups/src/Generator/CreateBuildFileElements.swift
brentleyjones: `callable` is defined as `callable: Callable`. `Callable` can't have labels, as it's a function type. It's the main reason for the concrete types and using `callAsFunction()`, so that external to this type we can use labels.

PR #2419, opened 2023-07-31
Fix another extraneous Xcode configuration disambiguation

brentleyjones:


tools/generators/legacy/src/Generator/DisambiguateTargets.swift
```diff
@@ -525,18 +525,15 @@ struct ArchitectureComponents {
...
     }
 
-    /// The set of `ConsolidatedTarget.Key`s among the targets passed to
-    /// `add(target:consolidatedKey:)`.
-    private var consolidatedKeys: Set<ConsolidatedTarget.Key> = []
-
     /// The set of xcodeConfigurations among the targets passed to
     /// `add(target:consolidatedKey:)`.
-    private var xcodeConfigurations: Set<OrderedSet<String>> = []
+    private var xcodeConfigurations: [ConsolidatedTarget.Key: Set<String>]
```
thiagohmcruz, line 530: Mind clarifying why we don't need `OrderedSet` anymore?

thiagohmcruz (approved):

tools/generators/legacy/src/Generator/DisambiguateTargets.swift
brentleyjones: I added it a couple PRs back to work around an issue with disambiguation involving multiple Xcode configurations. In particular with the `DistinguishingKey`.  A later change will remove that part all together. Here though, I only used `OrderedSet` in these intermediary types to prevent having to cast to a different type. They didn't need to be.

PR #2420, opened 2023-07-31
Only build up disambiguators for the correct type of name

brentleyjones:
This way we aren’t disambiguating both name and label form of every target. This will also allow us to add a third form without additional cost.


tools/generators/legacy/src/Generator/DisambiguateTargets.swift
```diff
@@ -32,26 +32,31 @@ extension Generator {
...
-                .insert(normalizedLabel)
-            names[normalizedName, default: .init()]
-                .add(target: target, key: key)
-            labels[normalizedLabel, default: .init()]
-                .add(target: target, key: key)
+                .insert(target.normalizedLabel)
+        }
+        for (key, target) in targets {
+            let nameAndProductType = NameAndProductType(target: target)
+            if labelsByNameAndProductType[nameAndProductType]!.count == 1 {
```
thiagohmcruz, line 41: We know this force unwrap is safe since 

`targets` not empty => `nameAndProductType` necessarily in `labelsByNameAndProductType`

correct?

thiagohmcruz (approved):

tools/generators/legacy/src/Generator/DisambiguateTargets.swift
brentleyjones: We do the same loop above, and for every target we insert into `labelsByNameAndProductType`. So it minimally has one element for every `target`, yes.

PR #2430, opened 2023-08-01
Stop including normal groups in `bazelPathAndIdentifiers`

brentleyjones:
`bazelPathAndIdentifiers` will be used to map `BazelPath`s to identifiers for `PBXBuildFile` creation. A `PBXBuildFile` won’t need to reference a group, so we don’t need to add those identifiers to `bazelPathAndIdentifiers`.

### DIFF
--- a/tools/generators/files_and_groups/src/ElementCreator/CreateGroupChild.swift
+++ b/tools/generators/files_and_groups/src/ElementCreator/CreateGroupChild.swift
@@ -189,11 +189,14 @@ extension GroupChild.ElementAndChildren {
     init(
         bazelPath: BazelPath,
         element: Element,
+        includeParentInBazelPathAndIdentifiers: Bool = true,
         resolvedRepository: ResolvedRepository?,
         children: GroupChildElements
     ) {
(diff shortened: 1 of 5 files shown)


tools/generators/files_and_groups/src/ElementCreator/CreateGroupChild.swift
```diff
@@ -189,11 +189,14 @@ extension GroupChild.ElementAndChildren {
     init(
         bazelPath: BazelPath,
         element: Element,
+        includeInBazelPathAndIdentifiers: Bool = true,
```
mattrobmattrob, line 192: Might be nice to mention that the children are included but this controls the element's identifier?

mattrobmattrob (approved):

tools/generators/files_and_groups/src/ElementCreator/CreateGroupChild.swift
brentleyjones: Ahh, yeah, good call.

PR #2461, opened 2023-08-11
Change how host output groups are set in fallback case

brentleyjones:
Since the `BAZEL_HOST_` build settings aren’t read in `calculate_output_groups.py`, it’s better to have the fallback values determined in a similar way, by having each buildable contribute to the file. We don't include `additionalBuildableReferences` because those are transitive dependencies and are only in the scheme for Swift UI Previews support.


examples/integration/test/fixtures/bwb.xcodeproj/xcshareddata/xcschemes/WidgetExtension.xcscheme
```diff
@@ -39,6 +39,22 @@
                </EnvironmentBuildable>
             </ActionContent>
          </ExecutionAction>
+         <ExecutionAction
+            ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
+            <ActionContent
+               title = "Set Bazel Build Output Groups for iOSApp"
+               scriptText = "echo &quot;$BAZEL_LABEL,$BAZEL_TARGET_ID&quot; &gt;&gt; &quot;$SCHEME_TARGET_IDS_FILE&quot;&#10;">
```
mattrobmattrob, line 46: This added `ExecutionAction` is intentional, right? Just wondering since it'll, in theory, be appending to the same `$SCHEME_TARGET_IDS_FILE`.

mattrobmattrob (approved):

examples/integration/test/fixtures/bwb.xcodeproj/xcshareddata/xcschemes/WidgetExtension.xcscheme
brentleyjones: Correct. The purpose of the PR is to add these, so the additional values in `$SCHEME_TARGET_IDS_FILE` will be used by `calculate_output_groups.py` in the fallback case.

BalestraPatrick (approved):

PR #2463, opened 2023-08-11
Add `PBXProductType.resourceBundle`

brentleyjones:
A resource bundle and a executable bundle both use the `com.apple.product-type.bundle` product identifier. In order to support encoding the distinction between the two, without adding an `isResourceBundle` attribute like before, while also taking the opportunity to compress the information into a single character, we now use a different raw value for the `enum` and translate it to an identifier where needed.


xcodeproj/internal/product.bzl
```diff
@@ -4,6 +4,41 @@ load("@bazel_skylib//lib:paths.bzl", "paths")
...
+    "com.apple.product-type.app-extension.intents-service": "i",
+    "com.apple.product-type.app-extension.messages": "m",
+    "com.apple.product-type.app-extension.messages-sticker-pack": "s",
+    "com.apple.product-type.tv-app-extension": "t",
+    "com.apple.product-type.extensionkit-extension": "E",
+    "com.apple.product-type.watchkit-extension": "1",
+    "com.apple.product-type.watchkit2-extension": "W",
+    "com.apple.product-type.xcode-extension": "2",
+    "com.apple.product-type.bundle": "B",
+    # Resource bundles are "b"
```
thiagohmcruz, line 25: Sorry, not sure if I understand how this comment relates to what comes next?

thiagohmcruz (approved):

xcodeproj/internal/product.bzl
brentleyjones: It doesn't, it's saying that no other value can use `"b"`, because resource bundles, which are also `com.apple.product-type.bundle`, will use it. I'll move it to be above `com.apple.product-type.bundle`.

thiagohmcruz: SGTM!

luispadron (approved):

PR #2465, opened 2023-08-14
Fix handling of unfocused `.xcdatamodel` bundles

### LINKED ISSUE #2464: Bug: `XCVersionGroup` not found in `elements`
### Description

Regressed here https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2457

* I hit this internally and for some reason this only happens if an `apple_framework` is the **only** top level target. You can see that in the branch linked below used for repro steps, if any other target is passed as top level plus the framework then the issue is gone.
* Seems to be related to the fact that the same `.xccurrentversion` file is being propagated in both direct and transitive here https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/main/xcodeproj/internal/input_files.bzl#L762-L766? If I manually remove an `.xccurrentversion` already present in the direct list from the transitive list the issue is also gone.

### Reproduction steps

* Checkout branch [thiago/xcversion-group-not-found-repro](https://github.com/MobileNativeFoundation/rules_xcodeproj/compare/main...thiagohmcruz:rules_xcodeproj:thiago/xcversion-group-not-found-repro)
* Then
```sh
cd examples/rules_ios

bazel run //:xcodeproj
```

You'll hit this err:
```sh
external/rules_xcodeproj/tools/generators/legacy/src/Generator/CreateFilesAndGroups.swift:795: "iOSApp/Source/CoreUtilsObjC/Model.xcdatamodeld" `XCVersionGroup` not found in `elements`
Please file a bug report at https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/new?template=bug.md
```

### Expected behavior

`bazel run //:xcodeproj` should generate the project with no errors.

### rules_xcodeproj version

51a4fece4de4b5198707c734c9a46106354f566d

### Xcode version

14.2

### Bazel version

6.3.2

### rules_apple version

repros in this repo so same as rules_xcodeproj at 51a4fece4de4b5198707c734c9a46106354f566d

### rules_swift version

repros in this repo so same as rules_xcodeproj at 51a4fece4de4b5198707c734c9a46106354f566d

### Additional information

_No response_

brentleyjones:
Fixes #2464.

“Regressed” with b75ca4b2f9ce0ef8401a09e94c902499b8fd72d9.

### DIFF
--- a/tools/generators/legacy/src/Generator/CreateFilesAndGroups.swift
+++ b/tools/generators/legacy/src/Generator/CreateFilesAndGroups.swift
@@ -795,7 +795,7 @@ extension Generator {
                 // We can get `.xccurrentversion` files for `.xcdatamodel`
                 // bundles that are uncategorized (e.g.
                 // `rules_ios_apple_framework.resource_bundles`). If no
-                // downstream target is focused, we will won't have an
+                // downstream target is focused, we won't have an
                 // `xcVersionGroups` for this `xccurrentversion`. We can safely
                 // ignore these.
                 continue


thiagohmcruz (approved):
Thx for the fix! Also yeah not a regression really but rather an edge case I guess, bad choice of word!

PR #2471, opened 2023-08-15
Filter BwB dependencies in Starlark

brentleyjones:
- Less information sent to generator
- Less memory Starlark memory retained
- Better target consolidation (since it was taking into account removed dependencies)


thiagohmcruz (approved):

mattrobmattrob (approved):

brentleyjones:
I just realized that this might have broken some people's custom schemes. The custom scheme logic needs to iterate through dependencies to find the correct version of some labels, if they are listed in `build_action`.

No one has reported anything for this though, so it's probably an edge case. But just calling it out.

Issue #1837, opened 2023-03-10
Bug: Files generated via swift_proto_library are not accessible on Xcode file tree


ra1028:
## Describe the bug

Maybe, this should be classified as a feature request rather than a bug.
Files generated via swift_proto_library are not appeared in the `Bazel Generated Files` section of the Xcode file tree.
Therefore, jump-to-definition and autocomplete work well, but file search and syntax highlighting do not.

Probably, this is related to https://github.com/buildbuddy-io/rules_xcodeproj/issues/1744

## Reproduction steps

Define swift_proto_library target like below, and add it to the `xcodeproj`.

```python
proto_library(
    name = "example_proto",
    srcs = glob([
        "*.proto",
    ]),
)

swift_proto_library(
    name = "example_proto_swift",
    deps = [":example_proto"],
) 

swift_library(
    name = "example_lib",
    ...
    deps = [":example_proto_swift"]  ,   
)

ios_unit_test(
    name = "example_test",
    deps = [":example_lib"],
)

xcodeproj(
   name = "xcodeproj",
   ...
   top_level_targets = [
       top_level_target(":example_test", target_environments = ["simulator"]),
   ],
)
```

## Expected behavior

When an Xcode project is generated, rules_xcodeproj places the files to be generated in `xxx/bin/external/yyyy` under `Bazel Generated Files` as missing files, and then to be available after you build the target.
All the following Xcode features are enabled for the generated files and symbols.

- Jump to definition
- Auto completion
- Syntax highlighting
- Xcode file search
- Xcode text search

## Versions
- Bazel: 6.0.0-pre.20220804.3
- rules_xcodeproj: 1.2.0
- rules_apple: d04edf4136b4818662d7126582ec257aee372339 (not a release tag for some reason)
- rules_swift: 6153a848f747e90248a8673869c49631f1323ff3 (not a release tag for some reason)

## Additional context

We also have a target defined using [rules_proto_grpc](https://github.com/rules-proto-grpc/rules_proto_grpc), and the files generated by it appear on the Xcode file tree. It has a rule just to compile the protos and then the files are consumed by the official `swift_library` rule.

thii:
Addressed by https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2484.

PR #2473, opened 2023-08-16
Ensure generated `.swift` files are downloaded with BwtB

thii:
Indexing only works if the generated files present on disk, but files
generated by `swift_proto_library` aren't currently collected via any
output group. Before this change, if you build with BwtB, they wouldn't
get downloaded.

This is a temporary workaround until the following issue is properly
handled:
https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/1837


brentleyjones:
I'm surprised this is needed. If a generated file is in `srcs` it should get collected. Do you have a repro (is it in that other issue)?

thii:
It's not in `srcs` (`swift_proto_library` doesn't have a `srcs`, but takes a `proto_library` in `deps`).

brentleyjones:
But that doesn’t create a target, and thus isn’t indexed, right? We only need inputs for Xcode targets to fix indexing.

thii:
> Do you have a repro (is it in that other issue)?

Can you try with the proto example in rules_swift? https://github.com/bazelbuild/rules_swift/blob/master/examples/xplatform/proto/BUILD

brentleyjones:
Or are you saying that you just wish the file to exist so that you can jump to it? That's a different issue.

thii:
Ah, I mean it's not live indexed, but imported from a cached build. And when you click on a symbol but the file doesn't exist, Xcode just shows a question mark. If the file exist, Xcode correctly jumps to that. I'll provide an example, probably tomorrow.

brentleyjones:
So yes, this would fix the issue of referenced swift files needing to be downloaded. No need for a repro for that.

Do we need a similar thing for Objective-C? I guess not until we have index importing for it.

brentleyjones (approved):

brentleyjones:
> What issue are you referring to?

When you said:

> Indexing only works if the generated files present on disk

I thought you meant live indexing, not jumping to an already indexed source.

luispadron (approved):

mattrobmattrob (approved):

PR #2482, opened 2023-08-18
Simplify/improve `non_xcode_targets` `mergable_xcode_library_targets` calculation

brentleyjones:
This now allows multiple layers of non-Xcode targets to exist between the top-level target and the library. The `product_path` was no longer used, so I removed that as well.

### DIFF
--- a/xcodeproj/internal/non_xcode_targets.bzl
+++ b/xcodeproj/internal/non_xcode_targets.bzl
@@ -11,6 +11,7 @@ load(":configuration.bzl", "calculate_configuration")
 load(":input_files.bzl", "input_files")
 load(":linker_input_files.bzl", "linker_input_files")
 load(":lldb_contexts.bzl", "lldb_contexts")
+load(":memory_efficiency.bzl", "memory_efficient_depset")
 load(":output_files.bzl", "output_files")
 load(":processed_target.bzl", "processed_target")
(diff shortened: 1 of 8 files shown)


xcodeproj/internal/non_xcode_targets.bzl
```diff
@@ -98,17 +99,12 @@ rules_xcodeproj requires {} to have `{}` set.
...
-        for target, providers in [
-            (info.xcode_target, info.compilation_providers)
+    mergable_xcode_library_targets = memory_efficient_depset(
+        transitive = [
+            info.mergable_xcode_library_targets
             for info in transitive_infos
         ]
-        if providers._is_xcode_library_target
-    ]
+    )
```
brentleyjones, line 107: @mattrobmattrob Does this change make sense to you?

brentleyjones:
@luispadron @thiagohmcruz Can y'all test this doesn't regress anything for your rules_ios use case?

luispadron (approved):
Tested in our project, didn't notice any issues not in 1.9.1

xcodeproj/internal/non_xcode_targets.bzl
mattrobmattrob: Seems reasonable (and better) to me especially given no fixture changes WRT merging. 👍

Issue #1744, opened 2023-02-02
Feature Request: Add enhanced support for `swift_{grpc,proto}_library`


brentleyjones:
- [ ] A target needs to be created for each `proto_library` that is a transitive dependency of the `swift_proto_library`. A target shouldn't be made for `swift_proto_library` itself.
- [ ] An external target for `@com_github_apple_swift_protobuf//:SwiftProtobuf` need to be the dependency of these.
- [ ] The `X.pb.swift` generated files need to be sources of these `proto_library` targets.

brentleyjones:
> How to know if that target is a Swift target or a C++ target?

Why do you need to know? But we mainly check for `SwiftInfo` or look at the type of source files.

thii:
Given a swift_proto_library target and a cc_proto_library target both depend on a proto_library target, if an Xcode target is created for each proto_library, which type of source files should we collect for it?

brentleyjones:
I don't think that will matter. You would collect info up to the `swift_proto_library` target, then from there you would create multiple targets for each `proto_library`. You aren't going to create targets at the `proto_library` stage (similar to how we create resource bundles after getting to the bundling targets, because we need to let the aspects do their thing).

PR #2484, opened 2023-08-19
Add initial support for `swift_proto_library`

thii:
This only supports the use case where there's only one
`swift_proto_library` target for each `proto_library` target for now.

This will create a target for each `swift_proto_library`, with the
generated `.pb.swift` files being sources, and show them in Xcode.

Addresses https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/1837
Part of https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/1744


brentleyjones (approved):

mattrobmattrob (approved):
👏

PR #2485, opened 2023-08-21
Add support for Bazel 7 `ObjcProvider` changes

brentleyjones:
Info isn’t being populated into it like before, and eventually the provider is going away. So if we don’t find any libraries from it we fallback to the `CcInfo`.

### DIFF
--- a/xcodeproj/internal/compilation_providers.bzl
+++ b/xcodeproj/internal/compilation_providers.bzl
@@ -162,6 +162,7 @@ def _merge_compilation_providers(
             ],
             order = "topological",
         )
+
         # Works around an issue with `*_dynamic_framework`
         cc_info = None
     else:
--- a/xcodeproj/internal/compilation_providers.bzl
+++ b/xcodeproj/internal/compilation_providers.bzl
@@ -119,7 +119,6 @@ def _collect_compilation_providers(
             _framework_files = EMPTY_DEPSET,
             _is_top_level = False,
             _is_xcode_library_target = is_xcode_library_target,
-            _propagated_cc_info = cc_info,
             _propagated_framework_files = EMPTY_DEPSET,
             _propagated_objc = objc,
             cc_info = cc_info,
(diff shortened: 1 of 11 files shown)


brentleyjones:
I'm going to check if I need to stuff in `_extract_top_level_values` as well.

thiagohmcruz (approved):

brentleyjones:
This requires a bit more. Once I have Bazel 7 test running locally, I'll update this, merge it, then merge the Bazel 7 CI (I'll have this change in that PR as proof that it all works end-to-end).

xcodeproj/internal/compilation_providers.bzl
```diff
@@ -133,37 +148,61 @@ def _merge_compilation_providers(
...
+        propagated_framework_files = memory_efficient_depset(
+            transitive = [
+                apple_dynamic_framework_info.framework_files,
+                framework_files,
+            ],
+            order = "topological",
+        )
+    else:
+        propagated_framework_files = framework_files
+
```
brentleyjones, line 177: I'm not a fan of this. It all works, but it feels to heavy handed. I'm looking to see if there is a better way.

brentleyjones: Seems BwX framework detection is broken with the new linking scheme. We get symlinked inputs in `link.params`:

```
bazel-out/applebin_ios-ios_sim_arm64-dbg-ST-6c3cd4e1007a/bin/_solib_ios_sim_arm64/_U_S_SLib_CLibFramework.iOS___ULib/LibFramework.iOS_lipobin.dylib
bazel-out/ios-sim_arm64-min15.0-applebin_ios-ios_sim_arm64-dbg-ST-fda82787d112/bin/_solib_ios_sim_arm64/_U@_Umain~non_Umodule_Udeps~com_Ugithub_Ukrzyzanowskim_Ucryptoswift_S_S_CCryptoSwift___UCryptoSwift.xcframework_Sios-arm64_Ui386_Ux86_U64-simulator_SCryptoSwift.framework/CryptoSwift
```

This is actually better for BwB mode (as long as we can make sure these symlinks and the targets they link to are downloaded correctly with BwtB), because we can do away with `$PREVIEW_FRAMEWORK_PATHS`. For BwX mode though, I'm not sure I can find a way to map these symlinks to the dylib that Xcode produces. Besides for breaking SwiftUI Previews, it breaks linking all together. So I either have to fix it or BwX mode dies.

brentleyjones: Well, `$PREVIEW_FRAMEWORK_PATHS` is still needed, because of course the `@rpath`s and `LC_ID_DYLIB`s are the same (in order to support bundling).

brentleyjones: Disabling BwX mode for now.

yongjincho92 (approved):

PR #2491, opened 2023-08-21
Use Bzlmod with fixtures

brentleyjones:
Blzmod will be enabled by default with Bazel 7, so it makes sense to switch our default in our fixtures as well. This all also allow us to validate Bazel 7 fixtures in a future change.


brentleyjones:
`examples/sanitizers` and `examples/rules_ios` aren't setup to use blzmod yet, so they only have the Bazel 6 change of the leading `@` on labels.

thiagohmcruz (approved):

PR #2504, opened 2023-08-22
Test Bazel 7

brentleyjones:


buildbuddy.yaml
```diff
@@ -173,6 +195,16 @@ actions:
...
 
+  ## Uncomment when rules_ios performs ObjcProvider -> CcInfo linking migration
+  # - name: Integration Test - "examples/rules_ios" - Bazel 7
+  #   <<: *bazel_7
+  #   <<: *action_base
+  #   <<: *normal_resources
+  #   <<: *examples_rules_ios_workspace
+  #   bazel_commands:
+  #     - *validate_integration
+  #     - *build_all
```
brentleyjones, line 206: FYI @thiagohmcruz @luispadron

If you run

```
USE_BAZEL_VERSION=7.0.0-pre.20230724.1 bazel run --config=cache //test/fixtures:update
```

You'll see the fixture changes that shouldn't happen. I believe after rules_ios does the migration (you can use https://github.com/bazel-contrib/bazel_features/blob/69605e758b7ab93bbf3121027f032ac95e859d59/features.bzl#L16 if needed to support both Bazel 5 and 6+, or if you only support Bazel 6+ I think you can unconditionally make changes like rules_apple did), then the fixtures should show no changes under that version of Bazel 7.

mattrobmattrob (approved):

buildbuddy.yaml
luispadron: https://github.com/bazel-ios/rules_ios/issues/757

luispadron (approved):

PR #2506, opened 2023-08-23
Add `swift_{grpc,proto}_library`'s source files to a requested output group

thii:
Part of https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/1744


brentleyjones (commented):
I wonder if we should keep the regex expression. Just thinking about other rules like this that we may not support. I don't think it's bad to keep it, should only be a couple extra files downloaded in the worst case?

thii:
SGTM. I'll revert that change.

mattrobmattrob (approved):

Issue #1929, opened 2023-03-29
Bug: Fails to build rules_ios library target in BwB


luispadron:
### What happened?

Generated a `rules_ios` based project with `BwB` and the `all` generation scheme mode.
When building the "yellow" library targets in Xcode that are generated we received a run script error:

```
CompileC /Users/lpadron/Library/Developer/Xcode/DerivedData/Project-dxupostqumucsycroidgkcqwpevd/Build/Intermediates.noindex/Project.build/bazel-out/ios-sim_arm64-min14.0-applebin_ios-ios_sim_arm64-dbg-ST-839cdaff2330/bin/SwiftFramework/SwiftFramework/Objects-normal/arm64/_CompileStub_.o /Users/lpadron/Library/Developer/Xcode/DerivedData/Project-dxupostqumucsycroidgkcqwpevd/Build/Intermediates.noindex/Project.build/bazel-out/ios-sim_arm64-min14.0-applebin_ios-ios_sim_arm64-dbg-ST-839cdaff2330/bin/SwiftFramework/SwiftFramework/DerivedSources/_CompileStub_.m normal arm64 objective-c com.apple.compilers.llvm.clang.1_0.compiler (in target 'SwiftFramework' from project 'Project')
    cd /var/tmp/_bazel_lpadron/e26c4188e56823b3c8e12b2563cf675b/rules_xcodeproj/build_output_base/execroot/rules_xcodeproj_playground
[...]

PR #2511, opened 2023-08-24
Stop generating static framework targets in Xcode

luispadron:
These targets aren't buildable because of: https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/105 
We shouldn't show them in Xcode.

| Before  | After |
| ------------- | ------------- |
| Screenshot 2023-08-24 at 12 10 53 PM | Screenshot 2023-08-24 at 12 43 25 PM |

### DIFF
--- a/xcodeproj/internal/automatic_target_info.bzl
+++ b/xcodeproj/internal/automatic_target_info.bzl
@@ -49,6 +49,12 @@ def _is_test_target(target):
         return False
     return target[AppleBundleInfo].product_type in _TEST_TARGET_PRODUCT_TYPES
 
+def _is_static_framework(target):
+    """Returns whether the given target is a static framework or not."""
+    if AppleBundleInfo not in target:
+        return False
+    return target[AppleBundleInfo].product_type == "com.apple.product-type.framework.static"
+
 ## Aspects
 
 # These are declared as constants to cause starlark to reuse the same instances
@@ -211,6 +217,13 @@ def calculate_automatic_target_info(ctx, build_mode, target):
         is_top_level = True
         provisioning_profile = "provisioning_profile"
         xcode_targets = _TEST_BUNDLE_XCODE_TARGETS
+    elif _is_static_framework(target):
+        # TODO: https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/105
+        # Static frameworks are not supported yet as Xcode targets.
+        is_supported = False
+        is_top_level = True
+        collect_uncategorized_files = rule_kind != "apple_bundle_import"
+        xcode_targets = _DEFAULT_XCODE_TARGETS[this_target_type]
     elif AppleBundleInfo in target and target[AppleBundleInfo].binary:
         # Checking for `binary` being set is to work around a rules_ios issue
         alternate_icons = "alternate_icons"
--- a/xcodeproj/internal/unsupported_targets.bzl
+++ b/xcodeproj/internal/unsupported_targets.bzl
@@ -110,6 +110,7 @@ def process_unsupported_target(
         compilation_providers = compilation_providers,
         dependencies = dependencies,
         inputs = provider_inputs,
+        is_top_level = automatic_target_info.is_top_level,
         lldb_context = lldb_contexts.collect(
             build_mode = build_mode,
             id = None,


luispadron:
Seeing this error if folks have ideas: 

```sh
cd examples/rules_ios && bazel run //:xcodeproj

Error in fail:
'@//Lib:LibDynamic' is not a top-level target, but was listed in `top_level_targets`. Only list top-level targets (e.g. binaries, apps, tests, or distributable frameworks) in `top_level_targets`. Schemes and `focused_targets`/`unfocused_targets` can refer to dependencies of targets listed in `top_level_targets`, and don't need to be listed in `top_level_targets` themselves.
````

Looks like not generating the top level static framework causes the underlying buildable library to not be included as top level

File: xcodeproj/internal/non_xcode_targets.bzl [hunk 1]
```diff
@@ -45,6 +46,12 @@ def process_non_xcode_target(
     cc_info = target[CcInfo] if CcInfo in target else None
     objc = target[apple_common.Objc] if apple_common.Objc in target else None
     swift_info = target[SwiftInfo] if SwiftInfo in target else None
+    is_top_level_target = False
+
+    # TODO: https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/105
+    # Static framework targets are not supported as Xcode targets yet but are still top level targets.
+    if AppleBundleInfo in target and target[AppleBundleInfo].product_type == "com.apple.product-type.framework.static":
```
luispadron, line 53: Is it worth reusing `_is_static_framework` here?

xcodeproj/internal/automatic_target_info.bzl
```diff
@@ -196,6 +202,12 @@ def calculate_automatic_target_info(ctx, build_mode, target):
         infoplists = _INFOPLISTS_ATTRS
         provisioning_profile = "provisioning_profile"
         xcode_targets = _TEST_BUNDLE_XCODE_TARGETS
+    elif _is_static_framework(target):
```
luispadron, line 205: Should this only be done for rules_ios targets?

brentleyjones:
Side comment: let's land https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2512 first, since your stuff is in the same area.

luispadron:
Will rebase and refactor this to work with your changes

xcodeproj/internal/automatic_target_info.bzl
```diff
@@ -200,6 +206,13 @@ def calculate_automatic_target_info(ctx, build_mode, target):
         is_top_level = True
         provisioning_profile = "provisioning_profile"
         xcode_targets = _TEST_BUNDLE_XCODE_TARGETS
+    elif _is_static_framework(target):
+        # TODO: https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/105
+        # Static frameworks are not supported yet as Xcode targets.
+        is_supported = False
+        is_top_level = True
+        collect_uncategorized_files = rule_kind != "apple_bundle_import"
+        xcode_targets = _DEFAULT_XCODE_TARGETS[this_target_type]
```
luispadron, line 215: Are these the appropriate fields to forward?

luispadron:
Tested generating the Cash App project and seeing a lot nicer scheme selector without unbuildable static framework targets: 
Screenshot 2023-08-24 at 2 59 30 PM

@thiagohmcruz Mind giving this a test in Square repo and reporting back when you have time ❤️

luispadron:
Seeing some targets like `LibDynamic` are missing though so will investigate that more

mattrobmattrob:
And this is less a problem with `rules_ios` and more an issue with the fact that `apple_framework` ends up creating the "static framework" under the hood to wrap everything which is what this removes from Xcode since it isn't supported, correct?

luispadron:
> And this is less a problem with `rules_ios` and more an issue with the fact that `apple_framework` ends up creating the "static framework" under the hood to wrap everything which is what this removes from Xcode since it isn't supported, correct?

Yeah this is removing from Xcode all things with the `.framework.static` product type, I thought this would be fine because I assumed it created target like:

```
foo
  - foo (static framework)
  - foo_swift (swift_library)
  - foo_objc (objc_library)
```

luispadron:
So looking at the BUILD file it looks like `LibDynamic` is excluded because it's the top-level `static framework` (even though its marked `link_dynamic = True`, probably a rules_ios issue) and it depends on `Lib` which is still shown in the project so think that makes sense. 

Essentially this change will exclude the top level `framework.static` targets but attempt to keep their library deps included in the project.

Any thoughts @mattrobmattrob @thiagohmcruz around this change?

xcodeproj/internal/automatic_target_info.bzl
thiagohmcruz: https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/105 is not specific to `rules_ios` IIUC correct? If that's the case I wouldn't do this only for `rules_ios`

xcodeproj/internal/automatic_target_info.bzl
thiagohmcruz: Good Q and also not sure, maybe @brentleyjones could chime in?

thiagohmcruz (approved):
Overall not against the changes, tested and things seem to work fine for us.

xcodeproj/internal/automatic_target_info.bzl
brentleyjones: It's a good start. We will know if we need more fields.

luispadron:
@brentleyjones Is this something I should look at updating more or do we not want this? I'm kind of on the fence here, if all of these targets are strictly unbuildable currently then hiding them makes sense, but if under certain conditions they work I wouldn't want them to be hidden for folks who've been relying on them.

brentleyjones:
If they are hidden, then the library(s) that feed into it will still exist.

Can you remind me what happens in the broken experience? Also, how it behaves in the new incremental generation mode? That will inform me on if we should go with this or not. Personally I'm leaning towards taking this change though, since they aren't a supported target type right now.

luispadron:
> Can you remind me what happens in the broken experience?

You get this error: https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/1929. The libraries they package should be buildable though. 

I'll give the incremental generator a try and report back

brentleyjones:
With the latest incremental generation mode changes, static framework targets have proper target merging and now build.

luispadron:
Confirmed for our project the merging work and static framework targets build as expected 🎉

PR #2512, opened 2023-08-24
Refactor `non_xcode_target` handling

brentleyjones:
Renamed to `unsupported_target` to make it clear that it doesn’t have to do with generating an Xcode target (which target focusing can change).

Also moved top-level target detection into `XcodeProjAutomaticTargetProcessingInfo`.


xcodeproj/internal/automatic_target_info.bzl
```diff
@@ -155,13 +158,13 @@ def calculate_automatic_target_info(ctx, build_mode, target):
     hdrs = EMPTY_LIST
     implementation_deps = EMPTY_LIST
     infoplists = EMPTY_LIST
+    is_supported = True
+    is_top_level = False
```
luispadron, line 162: Perfect!

thiagohmcruz (approved):

PR #2530, opened 2023-08-30
Fix `_should_create_provider` for module extension targets

brentleyjones:
Regressed with 88237f24f1060ad875845a2583793ec7758af733.


brentleyjones:
@thii I get this error still though (pretty sure it was there before this regression?):
CleanShot 2023-08-30 at 08 32 18@2x

thii:
Yes, that was there before this regression. I'll need to find a way to collect implicit dependencies added by `swift_protoc_gen_aspect`.

thii (approved):

PR #2531, opened 2023-08-30
Fix inherited build settings

brentleyjones:
If we set a base value, then `$(inherited)` would pick up that value instead of the project default. Now if we use `$(inherited)` we don’t set the base value at all.


examples/integration/test/fixtures/bwb.xcodeproj/project.pbxproj
```diff
@@ -21450,10 +21450,8 @@
 				COMPILE_TARGET_NAME = iOSApp.library;
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
 				DEPLOYMENT_LOCATION = NO;
-				DEVELOPMENT_TEAM = "$(inherited)";
 				"DEVELOPMENT_TEAM[sdk=iphoneos*]" = V82V4GQZXM;
-				EXECUTABLE_NAME = iOSApp_ExecutableName;
-				"EXECUTABLE_NAME[sdk=iphoneos*]" = "$(inherited)";
+				"EXECUTABLE_NAME[sdk=iphonesimulator*]" = iOSApp_ExecutableName;
```
brentleyjones, line 21454: Before this change, the `$(inherited)` for the device build would use the same value as the simulator build.

examples/integration/test/fixtures/bwb.xcodeproj/project.pbxproj
```diff
@@ -21683,10 +21681,8 @@
 				COMPILE_TARGET_NAME = iOSApp.library;
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
 				DEPLOYMENT_LOCATION = NO;
-				DEVELOPMENT_TEAM = "$(inherited)";
```
brentleyjones, line 21686: This was superfluous.

tools/generators/legacy/src/Generator/SetTargetConfigurations.swift
```diff
@@ -812,11 +812,7 @@ where Key == BuildSettingConditional, Value == [String: BuildSetting] {
         // TODO: If we ever add support for Universal targets we need to
         //   consolidate "ARCHS" to an `.any` conditional
 
-        let baseConditions = self.keys
-            .sorted()
-            .filter { $0 != .any }
-
-        let baseCondition = baseConditions.first!
+        let baseConditions = Set(keys.sorted().filter { $0 != .any })
```
luispadron, line 815:
```suggestion
        let baseConditions = Set(keys).sorted().filter { $0 != .any })
```

I don't think `Set` guarantees order

luispadron (approved):

tools/generators/legacy/src/Generator/SetTargetConfigurations.swift
brentleyjones: Removed the sort since we don't care about the order anymore.

luispadron: Makes sense

PR #2542, opened 2023-09-05
Switch to `examples/rules_ios` for Bazel 5 testing

brentleyjones:
This will allow merging `examples/cc` into `examples/integration`, since it doesn’t support Bazel 5.

### DIFF
--- a/buildbuddy.yaml
+++ b/buildbuddy.yaml
@@ -160,14 +160,6 @@ actions:
     bazel_commands:
       - *nobzlmod_generate_integration
       - *nobzlmod_build_all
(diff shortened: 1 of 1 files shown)


brentleyjones:
I guess this won't work, because we use `bundle_name`, and that requires a newer rules_apple, which only works with Bazel 6+.

luispadron (approved):
I'd say it's fine to disable Bazel 5 jobs. We're going to bump minimum support to 6 in rules_ios

thiagohmcruz (approved):

brentleyjones:
I'll drop this one and remove all the Bazel 5 jobs instead.

PR #2546, opened 2023-09-06
Change sorting of build action in schemes

### LINKED ISSUE #2543: Bug: Static Library Icons are shown as Framework Icons in scheme selector
### Description

When you have SwiftUI Previews enabled `adjust_schemes_for_swiftui_previews  = True` and have a dynamic framework name starting with Alphabet `A`, rules_xcodeproj lists dynamic framework first in the scheme's Build Action [Sorted Alphabetically] and that causes the Scheme's Icon to show up as Framework even though it's a static library.

More Info - https://bazelbuild.slack.com/archives/C0429BN0F8V/p1693932846753139?thread_ts=1693931358.895869&cid=C0429BN0F8V 

### Reproduction steps

- Build An `ios_application` , and have an application target.
- Have this application depends on a static `swift_library` whose name starts with any letter other than `a`.
- Have a dynamic framework dependency whose name starts with `a` 

Then when you look for static library scheme in the generated project, The Icon shows up as Framework instead of Library.

### Expected behavior

- The scheme Icon for Static library should show up correctly.

### rules_xcodeproj version

1.7.0

### Xcode version

14.3.0

### Bazel version

5.4.3

### rules_apple version

_No response_

### rules_swift version

_No response_

### Additional information

_No response_

brentleyjones:
Fixes #2543.

This is to get the icon of the scheme to match, in order:
- Launch target
- Test target
- Other targets
- (never) Additional targets for Swift UI Previews

This still isn’t 100%, because it doesn’t use info from the other actions, but we can make it better in the incremental generators.

### DIFF
--- a/examples/integration/test/fixtures/bwb.xcodeproj/xcshareddata/xcschemes/UIFramework.tvOS.xcscheme
+++ b/examples/integration/test/fixtures/bwb.xcodeproj/xcshareddata/xcschemes/UIFramework.tvOS.xcscheme
@@ -48,9 +48,9 @@
             buildForAnalyzing = "YES">
             <BuildableReference
                BuildableIdentifier = "primary"
-               BlueprintIdentifier = "C03F5C0CA2104493AE9E52C9"
-               BuildableName = "LibFramework.tvOS.framework"
-               BlueprintName = "LibFramework.tvOS"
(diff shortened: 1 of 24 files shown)


luispadron (approved):
Nice!

PR #2556, opened 2023-09-08
Fix format of example macOS imported framework

brentleyjones:
https://developer.apple.com/documentation/bundleresources/placing_content_in_a_bundle

This is needed to work with rules_apple 3.0.


brentleyjones:
Looks like this requires rules_apple 3.0.

mattrobmattrob (approved):

PR #2571, opened 2023-09-11
Include all transitive swiftmodules in `bc` output group when needed

### LINKED ISSUE #2563: Bug: `swift_test` plugin dependency `.swiftmodule`s are not generated during Index Build
### Description

Target builds and runs all great, which is really cool. I saw an indexing issue when exploring rules_swift examples:
Screenshot 2023-09-08 at 10 58 07 AM

I see that the indexing invocations `-I` dirs don't have `StringifyMacroPlugin.swiftmodule`. When I modify the project by explicitly adding an -I pointing to a directory that has`StringifyMacroPlugin.swiftmodule`, indexing is all green.

### Reproduction steps

1. rules-swift 1.11 release.
2. Generate a rules_xcodeproj that has top-level targets: "stringify_macro", "stringify_client", "stringify_macro_test"
3. In the generated BwB project, go to StringifyMacroTests.swift.
4. Build the project. 
5. Project builds all great leaving an indexing issue to be fixed.

### Expected behavior

Indexing ✅

### rules_xcodeproj version

858c3967d4e2c7f8997e585769639da9ce4742f4

### Xcode version

Xcode 15 Beta 8 (15A5229m)

### Bazel version

6.2.0

### rules_apple version

_No response_

### rules_swift version

1.11

### Additional information

_No response_

brentleyjones:
Fixes #2563.

Top-level swift targets, such as `swift_test` don’t set a value for `SwiftInfo.direct_modules[0].swift`. This makes sense, as downstream targets shouldn’t get any transitive swiftmodules from it. We previously worked around this by manually including transitive direct swiftmodules. This worked fine until compiler plugins, which introduced a new way that transitive swiftmodules could be collected. Thankfully the information we need is in the `compilation_context`.

As part of this change, we now only add the `.swiftmodule` files to `generated`. Previously we would add `.swiftdoc` and the generated header as well. We can’t do that with the information in `compilation_context`. If the swift generated header is used as an input for another target, it will get picked up that way.

### DIFF
--- a/xcodeproj/internal/input_files.bzl
+++ b/xcodeproj/internal/input_files.bzl
@@ -553,10 +553,10 @@ def _collect_input_files(
                 ],
             )
         for module in non_target_swift_info_modules.to_list():
-            compiled, indexstore = swift_to_outputs(
+            swiftmodules, indexstore = swift_to_outputs(
                 parse_swift_info_module(module),
             )
-            generated.extend(compiled)
+            generated.extend(swiftmodules)
             if should_produce_output_groups and indexstore:
                 indexstores.append(indexstore)
 
@@ -568,16 +568,11 @@ def _collect_input_files(
         unfocused_generated_compiling = []
         unfocused_generated_indexstores = []
         for module in unfocused_swift_info_modules.to_list():
-            compiled, indexstore = swift_to_outputs(
+            swiftmodules, indexstore = swift_to_outputs(
                 parse_swift_info_module(module),
             )
 
-            if compiled:
-                # We only need the single swiftmodule in order to download
-                # everything from the remote cache (because of
-                # `--experimental_remote_download_regex`). Reducing the number
-                # of items in an output group keeps the BEP small.
-                unfocused_generated_compiling.append(compiled[0])
+            unfocused_generated_compiling.extend(swiftmodules)
             if indexstore:
                 unfocused_generated_indexstores.append(indexstore)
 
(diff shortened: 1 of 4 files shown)


xcodeproj/internal/input_files.bzl
```diff
@@ -643,6 +638,10 @@ def _collect_input_files(
     # represented in the BEP.
     modulemaps = modulemaps_depset.to_list()
 
+    a = [e for e in extra_files if "archive.objlist" in e]
+    if a:
+        print("WWWWWWWWTFFFFFFFFFFFF 1?", id, a)
+
```
brentleyjones, line 644: Will remove

xcodeproj/internal/output_files.bzl
```diff
@@ -39,8 +39,7 @@ def _get_outputs(*, debug_outputs, id, product, swift_info, output_group_info):
         # TODO: Actually handle more than one module?
         for module in swift_info.direct_modules:
             swift = parse_swift_info_module(module)
-            if swift:
-                break
+            break
```
brentleyjones, line 42: This change is because we want the first direct module always, to potentially get the dependency swiftmodules.

chiragramani (approved):

chiragramani:
Thank you @brentleyjones! indexing(https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/2563) is all green after this change.

PR #2590, opened 2023-09-15
Allow to use labels as Xcode target names

CognitiveDisson:
I've added a `target_name_mode` argument (maybe not the best name) to the rule that allows you to specify what will be used as the base name for Xcode targets.
The default is "auto" - the same behavior as now. Another option "label" - the full target label will be used as a name.

We want to use labels as the target title because we think it will improve ergonomics by allowing you to copy/paste a label from Xcode into the terminal, search for label usage, and help navigate to the target's source code.


brentleyjones:
> We want to use labels as the target title because we think it will improve ergonomics by allowing you to copy/paste a label from Xcode into the terminal, search for label usage, and help navigate to the target's source code.

How does this help navigate to the source code? Target names don't impact that in any way. Same for scheme names (which replace the `/` and `:` with `_` so they are mangled already).

I'm okay with something like this (I might want to tweak it a bit before landing), because it's just pushing a version that already exists conditionally. Just trying to fully understand the use case.

CognitiveDisson:
> How does this help navigate to the source code?

I just meant that the package name in the label matches the directory structure, which make it clear how to `cd` to this target.
Additionally, I have an alias that gets the path from the bazel label (via `bazelisk query * --output=build`) which I combine with fzf 😅

The idea is to make Bazel labels and syntax more visible to iOS developers who primarily work in Xcode. The **assumption** is that in its current state, the Xcode project kinda hide the use of Bazel, and that displaying labels in Xcode project will make it more transparent for developers.

>Target names don't impact that in any way. Same for scheme names (which replace the `/` and `:` with `_` so they are mangled already).
Unfortunately, this is true for scheme names, but you can copy the target name from the target list or from the "Report Navigator", "Test Navigator". Also in build logs - "Build target ..."

>I'm okay with something like this (I might want to tweak it a bit before landing), because it's just pushing a version that already exists conditionally. 

Great. Let me know where I should fix it.

tools/generators/legacy/src/Generator/DisambiguateTargets.swift
```diff
@@ -76,12 +76,19 @@ extension Generator {
...
+                case .label:
+                    baseName = "\(target.label)"
+                }
                 uniqueValues[key] = DisambiguatedTarget(
                     name: names[moduleNameAndProductType.normalizedModuleName]!
                         .uniqueName(
                             for: target,
                             key: key,
-                            baseName: target.moduleName
+                            baseName: baseName
```
brentleyjones, line 91: I think instead of this, we can have lines 74-104 (in the original source code) only happen when `consolidatedTargets.targetNameMode == .auto`. If you do that, it will reduce the amount of work, and do 106-113, which is what you are trying to do anyway.

CognitiveDisson: I used a switch instead because it will force to update it if a new option is added, but the downside of this is that some small part is duplicated. I hope you don't mind.

I also populate the label dictionary for each target (previously it was skipped due to `continue` inside guard statements above), which is perhaps even more appropriate, but has some memory penalty ( not big, I think ). This was necessary because the `target.normalizedLabel` did not exist in `labels` dictionary if the label option was selected.

tools/generators/legacy/src/Generator/ConsolidateTargets.swift
```diff
@@ -272,6 +274,7 @@ extension PlatformAndConfiguration: Comparable {
 struct ConsolidatedTargets: Equatable {
     let keys: [TargetID: ConsolidatedTarget.Key]
     let targets: [ConsolidatedTarget.Key: ConsolidatedTarget]
+    let targetNameMode: TargetNameMode
```
brentleyjones, line 277: I don't think we need this on this type? We can instead pass the value to `Generator.disambiguateTargets()`.

tools/generators/legacy/src/Generator/Environment.swift
```diff
@@ -24,6 +24,7 @@ struct Environment {
     let consolidateTargets: (
         _ targets: [TargetID: Target],
         _ xcodeGeneratedFiles: [FilePath: FilePath],
+        _ targetNameMode: TargetNameMode,
```
brentleyjones, line 27: After the above change, we should be able to remove it from here.

tools/generators/legacy/src/Generator/ConsolidateTargets.swift
CognitiveDisson: Fixed 👍

tools/generators/legacy/src/Generator/Environment.swift
CognitiveDisson: Fixed

xcodeproj/internal/fixtures.bzl
```diff
@@ -177,6 +178,8 @@ def xcodeproj_fixture(
         post_build: Maps to `xcodeproj.post_build`.
         pre_build: Maps to `xcodeproj.pre_build`.
         project_options: Maps to `xcodeproj.project_options`.
+        target_name_mode:  Maps to
+            `xcodeproj.target_name_mode`.
```
brentleyjones, line 182:
```suggestion
        target_name_mode:  Maps to `xcodeproj.target_name_mode`.
```

brentleyjones (approved):

brentleyjones (approved):

PR #2596, opened 2023-09-15
Simplify type of `targetIndex` argument of `slicedBy()`

brentleyjones:


tools/generators/pbxtargetdependencies/src/Generator/ConsolidationMapsArguments.swift
```diff
@@ -379,7 +379,7 @@ extension ConsolidationMapsArguments {
 
 private extension Array {
     func slicedBy<CountsCollection>(
-        targetIndex: Array<TargetID>.Index,
+        targetIndex: Int,
```
thiagohmcruz, line 382: Does this work without having to update the call sites?

thiagohmcruz (approved):

tools/generators/pbxtargetdependencies/src/Generator/ConsolidationMapsArguments.swift
brentleyjones: Yeah. All `Array` indexes are `Int`.

luispadron (approved):

PR #2607, opened 2023-09-20
Document how to manually build the release archive

CognitiveDisson:
I've added information on how to manually build the release archive (`bazel build //distribution:release`). As I see this hasn't been mentioned anywhere, so I think README.md is the right place.


brentleyjones (approved):
Thanks! This is a good start. I want to make a full developer focused document, but when I do I can move this there.

PR #2610, opened 2023-09-20
Test 6.4.0 RCs in CI

brentleyjones:
We would have had this more recently, but it was incorrectly returning an RC build instead of the latest release.


brentleyjones:
https://github.com/bazelbuild/bazelisk/issues/499

PR #2620, opened 2023-09-25
Fix parsing of `Optional` arguments

brentleyjones:
Returning `nil` means a parsing failure, we want `.some(.none)` instead.

### DIFF
--- a/tools/generators/lib/PBXProj/src/Optional+Extensions.swift
+++ b/tools/generators/lib/PBXProj/src/Optional+Extensions.swift
@@ -4,9 +4,10 @@ import ArgumentParser
 
 extension Optional: ExpressibleByArgument where Wrapped: ExpressibleByArgument {
     public init?(argument: String) {
-        guard !argument.isEmpty else {
-            return nil
+        if argument.isEmpty {
+            self = .none
+        } else {
+            self = Wrapped(argument: argument)
         }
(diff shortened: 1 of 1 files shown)


thiagohmcruz (approved):

tools/generators/lib/PBXProj/src/Optional+Extensions.swift
```diff
@@ -4,9 +4,10 @@ import ArgumentParser
 
 extension Optional: ExpressibleByArgument where Wrapped: ExpressibleByArgument {
     public init?(argument: String) {
-        guard !argument.isEmpty else {
-            return nil
+        if argument.isEmpty {
+            self = .none
```
thiagohmcruz, line 8: Any existing test already exercising this? If not worth adding one?

thiagohmcruz (approved):

tools/generators/lib/PBXProj/src/Optional+Extensions.swift
brentleyjones: No, but the future generators use, and they will error loudly if this fails. Now that it's fixed, not worth the effort imo.

thiagohmcruz: SGTM!

PR #2649, opened 2023-10-01
Favor shorthand form when using labels for Xcode target names

bubski:
### What has changed
Here's a proposal to favor the shorthand form of labels when they are used for Xcode target names,
i.e. `//foo/bar/lib` instead of `//foo/bar/lib:lib`.

That'll in particular make a positive difference when `target_name_mode = "label"` is used, and remove redundancies from names.

Before | After
--- | ---
label_before | label_after

### Why this was changed

For context, our project has targets that have long labels.
To give a pseudo-example, it would be more convenient to display in Xcode:
`//apple/Extraterrestrial/RegistrationSummaryTransition`
instead of 
`//apple/Extraterrestrial/RegistrationSummaryTransition:RegistrationSummaryTransition`

### How can a reviewer test this change

Since the shorthand form is synonymous with the full form, I don't expect this to break anything,
and haven't witnessed any issues during testing,
but please correct me if I'm shortsighted or if this change should be done elsewhere.

This has also been applied to the label fallback logic in `target_name_mode = "auto"`, and if the fallback label happens to have a shorthand form, it will be favored.

### DIFF
--- a/tools/generators/legacy/src/DTO/BazelLabel.swift
+++ b/tools/generators/legacy/src/DTO/BazelLabel.swift
@@ -1,9 +1,10 @@
 import Foundation
 
-struct BazelLabel: Equatable, Hashable {
+struct BazelLabel: Equatable, Hashable, CustomStringConvertible {
     let repository: String
     let package: String
     let name: String
+    let description: String
 }
 
 extension BazelLabel {
@@ -66,9 +67,11 @@ extension BazelLabel {
 
     init(_ value: String) throws {
         let parts = try Self.parse(value)
-        repository = parts.repository
-        package = parts.package
-        name = parts.name
+        self.init(
+            repository: parts.repository,
+            package: parts.package,
+            name: parts.name
+        )
     }
 
     init?(nilIfInvalid value: String) {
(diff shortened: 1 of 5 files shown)


tools/generators/legacy/src/DTO/BazelLabel.swift
```diff
@@ -108,6 +108,17 @@ extension BazelLabel: Decodable {
     }
 }
 
+extension BazelLabel {
+    /// The string representation of the label, favoring the shorthand form.
+    ///
+    /// E.g. for `@repo//foo/bar/wiz:wiz` it returns `@repo//foo/bar/wiz`,
+    /// while for `@repo//foo/bar/wiz:baz` it returns `@repo//foo/bar/wiz:baz`.
+    var shortDescription: String {
```
brentleyjones, line 116: By doing this only here, sorting can end up a little weird. I think we should modify the struct to include a bool (calculated once at parsing) that says if it should use short form or not. Then in `description` we can do this. That way sorting works based on the same form that is shown to users.

bubski: Thanks Brentley!

I moved the logic to `description`, but did not add a bool calculated during parsing.
Couldn't see that being needed, and instead `canOmitName` is calculated inline.
Wouldn't that be ok? Perhaps I misunderstood and you had something else in mind.

Also added and updated some unit tests for `BazelLabel`,
at the same time fixing handling of cases where `name` contains `/` to adhere to what's specified in [the docs](https://bazel.build/concepts/labels#target-names).
(though I think `rules_swift` doesn't handle targets named this way anyway)

brentleyjones: `description` is called lots of times, in particular because it's called in sorting. Because of this, I would prefer it to be calculated in `init` once. And actually, because it's called so much, let's go a step further and just pre-calculate the description and store it (instead of `canOmitName`).

bubski: I see, thank you for clarifying!
Pushed that change.

brentleyjones (approved):

bubski:
👍 squashed and signed-off (sorry, wasn't aware of that)

brentleyjones (approved):

brentleyjones:
Thank you!

brentleyjones:
You'll also need to update the fixtures: `./test/update_all_fixtures.sh`

luispadron (approved):

PR #2671, opened 2023-10-06
Convert `import_indexstores.sh` into a Swift binary

brentleyjones:
This is mainly to be able to do more complex stuff in the near future for incremental generators. It has the side benefit of being faster.


brentleyjones:
@MobileNativeFoundation/rules_xcodeproj-maintainers PTAL

thiagohmcruz (approved):

PR #2678, opened 2023-10-10
Add `xcschemes` generator

brentleyjones:


tools/generators/xcschemes/src/Generator/TargetArgsAndEnv.swift
```diff
@@ -0,0 +1,59 @@
+import ArgumentParser
+import Foundation
+import PBXProj
+import ToolCommon
+import XCScheme
+
+// FIXME: Extract to function struct
```
brentleyjones, line 7: I'll do this FIXME before merging.

tools/generators/xcschemes/src/Generator/TransitivePreviewReferences.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+import PBXProj
+import XCScheme
+
+// FIXME: Extract to function struct
```
brentleyjones, line 5: I'll do this FIXME before merging.

tools/generators/xcschemes/src/Generator/CreateScheme.swift
```diff
@@ -0,0 +1,459 @@
...
+        var buildPostActions: [ExecutionAction] = []
+        var buildPreActions: [ExecutionAction] = []
+        var launchPostActions: [ExecutionAction] = []
+        var launchPreActions: [ExecutionAction] = []
+        var profilePostActions: [ExecutionAction] = []
+        var profilePreActions: [ExecutionAction] = []
+        var testPostActions: [ExecutionAction] = []
+        var testPreActions: [ExecutionAction] = []
+
+        // FIXME: Order these based on `.order`
```
brentleyjones, line 115: I'll do this FIXME before merging.

brentleyjones (commented):
@MobileNativeFoundation/rules_xcodeproj-maintainers I have a couple FIXME to address, and lots of tests to add, but this is ready to be reviewed.

tools/generators/xcschemes/README.md
```diff
@@ -0,0 +1,516 @@
+# `XCScheme`s generator
+
+The `xschemes` generator generates `.xcscheme` files for a project.
```
thiagohmcruz, line 3:
```suggestion
The `xcschemes` generator generates `.xcscheme` files for a project.
```

tools/generators/xcschemes/src/Generator/CreateCustomSchemeInfos.swift
```diff
@@ -0,0 +1,619 @@
...
+            }
+            return try targetsByID.value(
+                for: extensionHostID,
+                context: "\(context()) extension host"
+            )
+        }
+
+        // Only set from-rule args and env if the custom scheme doesn't
+        // declare any
+        // TODO: Distinguish between `None` and `[]` in `xcshemes`?
```
thiagohmcruz, line 267: What would be the advantage or doing that instead of treating `None` and `[]` as the same "empty" input?

thiagohmcruz (approved):

tools/generators/xcschemes/src/Generator/CreateCustomSchemeInfos.swift
brentleyjones: Allows saying in the scheme that you don't want any arguments, regardless if the rule would provide some. Right now that's not possible, both `None` and `[]` would mean no custom ones, but you will get whatever the rule provides.

brentleyjones:
https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2686 shows the Starlark side of it.

tools/generators/xcschemes/src/Generator/CreateCustomSchemeInfos.swift
brentleyjones: I plan to have a special `inherit` value, which will also be the default. So if you set a falsey value, it won't inherit.

thiagohmcruz: SGTM, thx for clarifying!

brentleyjones: This is now in. `nil` means "inherit", and non-`nil` means that was explicitly specified by the user, so use those values. On the Starlark side `inherit` eventually becomes `nil` (because of a `-1` count), and `None`/`[]` both mean "don't inherit any args, I mean literally have no args for this action).

luispadron (approved):

brentleyjones:
https://github.com/MobileNativeFoundation/rules_xcodeproj/issues/2814

PR #2680, opened 2023-10-10
Allow remapping indexstores to multiple target names

brentleyjones:
This isn’t exercised yet. Incremental generation will use it.


tools/import_indexstores/Remaps.swift
```diff
@@ -25,7 +26,7 @@ func remapArgs(
         // another pattern:
         // https://github.com/bazelbuild/bazel/blob/c4a1ab8b6577c4376aaaa5c3c2d4ef07d524175c/src/main/java/com/google/devtools/build/lib/rules/cpp/CcCompilationHelper.java#L1358
         "-remap",
-        #"^(?:\#(executionRootRegex)/|\./)?(bazel-out/[^/]+/bin/)(?:_swift_incremental/)?(.*?)([^/]+)_objs/.*?([^/]+?)(?:\.swift)?\.o$=\#(objectFilePrefix)/$1$2$3/Objects-normal/\#(arch)/$4.o"#,
+        #"^(?:\#(executionRootRegex)/|\./)?(bazel-out/[^/]+/bin/)(?:_swift_incremental/)?(.*?)([^/]+)_objs/.*?([^/]+?)(?:\.swift)?\.o$=\#(objectFilePrefix)/$1$2\#(targetNameOverride ?? "$3")/Objects-normal/\#(arch)/$4.o"#,
```
brentleyjones, line 29: This is the main change. If we have a `targetNameOverride`, use that instead of actual (e.g. use `iOSApp` instead of `iOSApp.library`.

brentleyjones: I just realized we need to override more than just the name, because the library could be in a different package. I'll adjust this to provide the values for `$1$2$3` instead of just `$3`.

thiagohmcruz (approved):

PR #2681, opened 2023-10-10
Include `indexstores` as inputs to `indexstore_filelist`

brentleyjones:
This ensures they are generated/downloaded when using `--features=swift.split_derived_files_generation`.


chiragramani (approved):
Thank you, this makes the index-builds all green!

PR #2698, opened 2023-10-27
Make mutable copy of extensions when processing top-level target

markvasiv:
### What has changed
- Wraps `getattr` result with `list()` for when getting `extensions`.

### Why it was changed
- Logic below assumes `extension_targets` is mutable.
- When both `extensions` and `extension` are present, it fails with `attempting to change a frozen list`.


brentleyjones (approved):

brentleyjones:
You'll need to add a sign-off to the commit (amend it) to pass the DCO requirement.

luispadron (approved):

PR #2711, opened 2023-11-03
Add FIXME for the future

brentleyjones:


tools/lib/ToolCommon/src/URL+Extensions.swift
```diff
@@ -23,6 +23,7 @@ private extension URL.AsyncBytes {
             let bytesTask = Task {
                 var accumulator: [UInt8] = []
                 var iterator = makeAsyncIterator()
+                // FIXME: Handle when file doesn't exist (currently hangs)
```
brentleyjones, line 26: I'm not really sure how to address this right now. The FIXME is more as a note for anyone that runs into this issue in the future. It shouldn't ever happen, I only ran into it when there was something incorrectly setup on the Starlark size.

PR #2714, opened 2023-11-03
Improve handling of source-less library targets

brentleyjones:
This will allow for improved target merging in incremental generation mode.


luispadron (approved):

thiagohmcruz (approved):
Confirmed that this fixes the issue I was seeing before in the incremental gen. Thx!!

PR #2732, opened 2023-11-17
Fix framework Xcode previews for rules_ios

brentleyjones:
This makes the `ObjCProvider` merging happen in unsupported targets as well. I’m pretty sure this will introduce extra frameworks somehow, but it seems like the best fix at this time.


brentleyjones:
This can only really be fixed in the incremental generation, so closing.

PR #2768, opened 2023-12-04
Pass `rule_attr` and `rule_kind` instead of accessing `ctx`

brentleyjones:
End goal is to remove use of `ctx`.


thiagohmcruz (approved):
What is the advantage if removing `ctx`?

brentleyjones:
§2 in https://docs.google.com/document/d/1uuX1Gz6Kvivp-oOqGja342VrVO534PLF0m_zFtFIBcc

Issue #2641, opened 2023-09-29
Bug: BwB Xcode Previews fail to find transitive frameworks when using `--define=apple.experimental.tree_artifact_outputs=0`


brentleyjones:
### Description

We don't unarchive transitive frameworks when copying them over to `SwiftUIPreviewsFrameworks`. This is an issue when using `--define=apple.experimental.tree_artifact_outputs=0`.

### Reproduction steps

In `examples/integration` (which currently sets `--define=apple.experimental.tree_artifact_outputs=0`) try to use the `UI` Xcode Preview.

### Expected behavior

We shouldn't unarchive artifacts in `copy_outputs.sh`. Instead we should generate an unarchiving action, and depend on that instead in the output groups. Then, regardless of what `apple.experimental.tree_artifact_outputs` is set to, we will always have a tree artifact.

### rules_xcodeproj version

1.11.0

### Xcode version

15.0

### Bazel version

6.4.0rc1

brentleyjones:
As of #2779:

> The last thing needed is adding these products to an output group for Xcode previews, otherwise unfocused frameworks aren’t extracted.

PR #2779, opened 2023-12-05
Extract bundles in Bazel instead of `copy_outputs.sh`

brentleyjones:
Part of #2641.

This gets us most of the way to fixing Xcode Previews when not using tree artifacts. The last thing needed is adding these products to an output group for Xcode previews, otherwise unfocused frameworks aren’t extracted.

### DIFF
--- a/xcodeproj/internal/product.bzl
+++ b/xcodeproj/internal/product.bzl
@@ -37,6 +37,11 @@ PRODUCT_TYPE_ENCODED = {
     "com.apple.product-type.xpc-service": "X",
 }
 
+_ARCHIVE_EXTENSIONS = {
+    "ipa": None,
+    "zip": None,
+}
+
 def _codesign_executable(*, actions, executable):
     executable_path = "{}_codesigned".format(
         executable.basename,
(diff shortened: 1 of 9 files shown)


xcodeproj/internal/product.bzl
```diff
@@ -98,14 +103,51 @@ exit ${PIPESTATUS[0]}
...
+fi
+
+expanded_dir=$(mktemp -d)
+trap 'rm -rf "$expanded_dir"' EXIT
+
+unzip -q -DD "$archive" -d "$expanded_dir"
+mv "$expanded_dir$suffix/${output##*/}" "${output%/*}"
+""",
+        arguments = [args],
+        mnemonic = "XcodeProjExtractBundleArchive",
```
mattrobmattrob, line 138: This mnemonic should likely be added to here, right? Do we need to call it out in any docs for users?

https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/61dd744ea6806937b80330d81cac8fffdac1f9d6/shared.bazelrc#L53

mattrobmattrob (approved):

xcodeproj/internal/product.bzl
brentleyjones: I was wondering if I should mirror one of the existing bundle ones. I did at first,but I thought it might make debugging stuff more difficult.

Maybe we just add the correct execution info here. Though I'm not sure we want to force it a certain way by default (though also, I think with modify_execution_info you can override the defaults?).

mattrobmattrob: I personally like it being separate. Just want it called out as a potential cache inflator that's not necessarily expected when onboarding/updating. And I'm not sure how `modify_execution_info` plays with defaults.

PR #2829, opened 2023-12-20
Keep scheme sorting info in rsync

yongjincho92:
Issue : 
Xcode schemes were not sorted as expected 

Cause :
Scheme sorting hint file was included under xcuserdata dir and was not being rsycn due to existing --exaclude pattern. 

Change : 
Make the existing exclude a little bit more specific.

### DIFF
--- a/xcodeproj/internal/templates/legacy_installer.sh
+++ b/xcodeproj/internal/templates/legacy_installer.sh
@@ -134,7 +134,7 @@ rsync \
   --chmod=u+w,F-x \
   --exclude=project.xcworkspace \
   --exclude=rules_xcodeproj/bazel \
-  --exclude=xcuserdata \
+  --exclude="*.xcbkptlist" \
   --delete \
   "$src/" "$dest/"
 


xcodeproj/internal/templates/legacy_installer.sh
```diff
@@ -134,7 +134,7 @@ rsync \
   --chmod=u+w,F-x \
   --exclude=project.xcworkspace \
   --exclude=rules_xcodeproj/bazel \
-  --exclude=xcuserdata \
+  --exclude="*.xcbkptlist" \
```
brentleyjones, line 137: User created schemes should be left alone as well. Probably better to leave this exclude as is, and copy over the schememanagement file, similar to the incremental installer?

https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/4f6860856f808e2e044637380daadc2d2090bc05/xcodeproj/internal/templates/incremental_installer.sh#L138-L142

yongjincho92: Sounds good, let me update this PR

brentleyjones: @yongjincho92 Any progress on this?

brentleyjones:
@yongjincho92 This still seems like a good idea.

erneestoc:
+1

PR #2851, opened 2024-01-04
Track generated source files in incremental generation mode

brentleyjones:
Source files are tracked as build files by Xcode, so building targets that directly use generated source files will fail the first time they are built if we don't track them.


yongjincho92 (approved):
Fixes the issue in our project

PR #2853, opened 2024-01-05
Use Bzlmod lockfile locally to support offline builds

brentleyjones:


bazel_6.bazelrc
```diff
@@ -7,6 +7,3 @@ build --enable_bzlmod
 build --apple_crosstool_top=@local_config_apple_cc//:toolchain
 build --crosstool_top=@local_config_apple_cc//:toolchain
 build --host_crosstool_top=@local_config_apple_cc//:toolchain
-
-# Disabled until bugs are fixed
```
mattrobmattrob, line 11: The bugs were fixed, I guess?

.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 
 /bazel-*
+MODULE.bazel.lock
```
mattrobmattrob, line 4: This will include the `examples/*/MODULE.bazel.lock` as well, correct?

mattrobmattrob (approved):

bazel_6.bazelrc
brentleyjones: It's disabled by default in 6.4.0, so no need to call it out here. Didn't want to see if they actually were fixed and enable it.

.gitignore
brentleyjones: Yep!

PR #2859, opened 2024-01-05
Depend on `--experimental_remote_download_regex` for Index Build input collection in incremental generation mode

### LINKED ISSUE #2852: [Incremental] Bug: Index failed: 'fatal error: virtual filesystem overlay file not found'
### Description

Index failed when run examples/rules_ios in `incremental mode`. Howerver it works in `legacy mode`.
Detail error shown as below:
`fatal error: virtual filesystem overlay file '/*/rules_xcodeproj/examples/rules_ios/bazel-output-base/rules_xcodeproj.noindex/indexbuild_output_base/execroot/_main/bazel-out/ios-sim_arm64-min15.0-applebin_ios-ios_sim_arm64-dbg-ST-a4a9813474a3/bin/iOSApp/Test/MixedUnitTests/iOSAppMixedUnitTests_vfs.yaml' not found`

### Reproduction steps

1. checkout to main branch (currently at commit 46f7a1f47c02e8bec3398913c5d018dd3cc71da5) of 'rules_xcodeproj'
2. cd examples/rules_ios
3. run `bazel run //:xcodeproj-incremental`
4. open rules_ios.xcodeproj
5. wait for Index

### Expected behavior

Index works in incremental mode.

### rules_xcodeproj version

46f7a1f47c02e8bec3398913c5d018dd3cc71da5

### Xcode version

15.1

### Bazel version

7.0.0

### rules_apple version

_No response_

### rules_swift version

_No response_

### Additional information

_No response_

brentleyjones:
Fixes #2852.

This is a pretty big shift away from how we used to do it.

With the baggage of BwX mode gone we don’t have to track generated files in the same way. After removing that tracking we can depend on building the product for each target (i.e. the `bp` output group) to transitive generate all files needed to compile that target.

In a BwtB world though we need to make sure we still download all of those inputs, which is where `--experimental_remote_download_regex` comes in. I’ve added all of the file types that I currently know of to the regex, but it’s brittle and may require more from the user. The flag supports multiple uses, so a user can fix this on their own as well.

There are four main benefits with this new way (which is good, because the above is a pretty big downside):

1. We now generate/download previously tricky to track files like vfs overlays and hmaps
2. Less Starlark CPU usage collecting all of the generated files
3. Less Starlark memory usage retaining these generated files
4. Much smaller BEP, since the output groups are smaller, resulting in less memory usage and faster builds

### DIFF
--- a/xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
+++ b/xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
@@ -22,12 +22,14 @@ else
     # Inputs for compiling, inputs for linking, and index store data
     readonly output_group_prefixes="xc,xl,xi"
   elif [[ "${ENABLE_PREVIEWS:-}" == "YES" ]]; then
-    # Compiled outputs (i.e. swiftmodules) and generated inputs, products (i.e.
-    # bundles), index store data, and link params
+    # Compile params, products (i.e. bundles) and index store data, and link
+    # params
+    # TODO: Remove `bi` once we remove support for legacy generation mode
     readonly output_group_prefixes="bc,bp,bi,bl"
   else
-    # Compiled outputs (i.e. swiftmodules) and generated
-    # inputs, products (i.e. bundles), and index store data
+    # Products (i.e. bundles) and index store data
+    # TODO: Remove `bc` and `bi` once we remove support for legacy generation
+    # mode
     readonly output_group_prefixes="bc,bp,bi"
   fi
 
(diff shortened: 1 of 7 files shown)


xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
```diff
@@ -108,7 +110,17 @@ fi
...
+  #
+  # - .indexstore directories to allow importing indexes
+  # - .a and .swiftmodule files for lldb
+  # - primary compilation input files (.c, .C, .cc, .cl, .cpp, .cu, .cxx, .c++,
+  #   .m, .mm, .swift) for Xcode build input checking
+  #
+  # This is brittle. If different file extensions are used for compilation
+  # inputs, they will need to be added to this list. Ideally we can stop doing
+  # this once Bazel adds support for a Remote Output Service.
+  "--experimental_remote_download_regex=.*\.indexstore/.*|.*\.(a|c|C|cc|cl|cpp|cu|cxx|c++|m|mm|swift|swiftmodule)$"
```
jszumski, line 123: Was it intentional to drop `swiftdoc` and `swiftsourceinfo` here?

brentleyjones: Yes. They are only needed for indexing, and that's covered in the other `--experimental_remote_download_regex`.

Now that you ask though, I'm not sure which location Xcode uses. It might/is probably the main one instead of the index output base. I'll add them back.

jszumski (approved):

luispadron (approved):

PR #2866, opened 2024-01-09
Support pre/post actions with `launch_path`

luispadron:
Follow up to #2785 which added `launch_path`. After using this more I realized we can also support `pre`/`post` actions for this attribute.

### DIFF
--- a/tools/generators/xcschemes/src/Generator/SchemeInfo.swift
+++ b/tools/generators/xcschemes/src/Generator/SchemeInfo.swift
@@ -15,7 +15,7 @@ struct SchemeInfo: Equatable {
         let scriptText: String
         let action: Action
         let isPreAction: Bool
-        let target: Target
+        let target: Target?
         let order: Int?
     }
 
--- a/xcodeproj/internal/xcschemes/xcschemes_execution.bzl
+++ b/xcodeproj/internal/xcschemes/xcschemes_execution.bzl
@@ -256,6 +256,7 @@ def _write_schemes(
             custom_scheme_args.add(build_target.id)
             _add_execution_actions(
                 build_target,
+                target_id = build_target.id,
                 action_name = action_name,
                 scheme_name = scheme_name,
             )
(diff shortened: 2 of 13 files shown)


tools/generators/xcschemes/src/Generator/SchemeInfo.swift
```diff
@@ -15,7 +15,7 @@ struct SchemeInfo: Equatable {
         let scriptText: String
         let action: Action
         let isPreAction: Bool
-        let target: Target
+        let target: Target?
```
luispadron, line 18: I don't love this change but I'm not sure how else to represent the execution actions which have no associated build targets (i.e. `launch_path`)

brentleyjones: With your change it makes sense. And technically we could allow in the future execution actions without the macro expansion set, but the current API doesn't allow that.

xcodeproj/internal/xcschemes/xcschemes_execution.bzl
```diff
@@ -317,7 +317,7 @@ def _write_schemes(
             _add_execution_action(
                 action,
                 action_name = action_name,
-                id = target.id,
+                id = getattr(target, "id", None),
```
brentleyjones, line 320: I think checking `is_path` in an inline `if else` would be more performant.

luispadron: I dont have access to `is_path` in here without adding another argument. I moved the `getattr` to be outside the for loop though

brentleyjones: But `target` is `launch_target`, which has `.is_path`, right?

luispadron: It wont always be, it looks like this is called for test targets too which wouldn't have `is_path`. But maybe I'm missing something

brentleyjones: Ahh. Can we change it so we pass in `id` then? We already do an `if` on `is_path`, so we can assign to the variable in both paths there.

luispadron: Updated!

xcodeproj/internal/xcschemes/xcschemes_execution.bzl
```diff
@@ -317,27 +318,31 @@ def _write_schemes(
...
-                id = target.id,
+                id = target_id,
                 is_pre_action = FALSE_ARG,
                 scheme_name = scheme_name,
             )
 
     # buildifier: disable=uninitialized
     def _add_launch_target(launch_target, *, action_name, scheme_name):
         custom_scheme_args.add(launch_target.is_path)
+        target_id = None
```
brentleyjones, line 329: Nit: can you assign this in the `if`, to reduce a double assignment in the `else` path?

brentleyjones (approved):

PR #2868, opened 2024-01-16
Fix macro expansion for test schemes without launch targets

brentleyjones:


tools/generators/xcschemes/src/Generator/CreateScheme.swift
```diff
@@ -401,7 +412,7 @@ extension Generator.CreateScheme {
                 preActions: profilePreActions
                     .sorted(by: compareExecutionActions)
                     .map(\.action),
-                useLaunchSchemeArgsEnv: true,
```
brentleyjones, line 404: Also fixes this being always set.

luispadron (approved):

mattrobmattrob (approved):

PR #2878, opened 2024-01-24
Use sub-params file for incremental link params processor

polac24:
Rehash of https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/1862 for `incremental_link_params_processor.py`

### Problem

In a step for cache population for a massive application we get "Argument list too long" also from `incremental_link_params_processor`. (command: `bazel run //:name.xcodeproj -- --generator_output_groups=all_targets build`)

```
Action failed to execute: java.io.IOException: Cannot run program "...//rules_xcodeproj.noindex/build_output_base/execroot/register"): error=7, Argument list too long
...

BAZEL ERROR
ERROR: .../BUILD.bazel:373:19: Generating bazel-out/*/*-App.rules_xcodeproj.link.params failed: (Exit -1): incremental_link_params_processor failed: error executing command (from target *-App)
```

### DIFF
--- a/tools/params_processors/incremental_link_params_processor.py
+++ b/tools/params_processors/incremental_link_params_processor.py
@@ -29,19 +29,21 @@ _LD_SKIP_OPTS = {
 }
 
 
-def _parse_args(args: List[str]) -> List[str]:
-    expanded_args = []
-    # Drop first argument since it's the linker executable
-    for line in args[1:]:
-        if line.startswith("@"):
-            # Sometimes arguments might be a params file
-            with open(line[1:], encoding = "utf-8") as f:
-                expanded_args.extend(f.read().splitlines())
-        else:
-            expanded_args.append(line)
+def _parse_args(args_files: List[str]) -> List[str]:
+    args = []
+    for args_path in args_files:
+        # Each argument is a path to a file containing the actual arguments
+        with open(args_path, encoding = "utf-8") as fp:
+            lines = fp.read().splitlines()
+            if lines[0].startswith("@"):
+                # Sometimes those arguments might be also be a redirect
+                with open(lines[0][1:], encoding = "utf-8") as f:
+                    args.extend(f.read().splitlines())
+            else:
+                # First argument is the tool name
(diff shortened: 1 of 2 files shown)


brentleyjones:
46f7a1f47c02e8bec3398913c5d018dd3cc71da5 fixed this for the compile params. Ideally we can find a way to do the same here. The `args()` that we use should already turn into params files.

polac24:
@brentleyjones, thanks for referencing. I tried different features to force param files mode in linking, without success (I tested `archive_param_file`, `linker_param_file, `supports_param_files`). Let me know, if you know any other ways to workaround that.

If there is nothing available yet, can we workaround this issue with this PR?

brentleyjones:
Those features would adjust how the normal linking happens. The actions already use a params file though if they don't themselves run into this issue, so instead we are probably dealing with a native rule -> starlark thing. Let me dig in more to see if there is something else we can flip before going this route.

brentleyjones:
Yep, looks like this will always be broken the way it is until the rules are Starlarkified: https://github.com/bazelbuild/bazel/blob/48892ae099581d4c9ad9bab785add2c0bb17e6be/src/main/java/com/google/devtools/build/lib/rules/cpp/CppLinkAction.java#L228-L242.

I'm fine with something like this change. I'll approve the PR to get it running on CI.

brentleyjones (approved):

brentleyjones:
I assume you've tested this with SwiftUI previews? Also `examples/integration`?

polac24:
@brentleyjones, sorry for a late response: yes, I tested SwiftUI preview in `examples/integrations` - works fine. 
Let me know if I should update something - from what I see, errors in `Test - Bazel HEAD` is not introduced in this PR (saw failing on other PRs too).

brentleyjones:
Correct, those you can ignore.

brentleyjones:
Just need another approval.

luispadron (approved):

thiagohmcruz (approved):

PR #2880, opened 2024-01-25
Define _rules_xcodeproj_build config for all commands

polac24:
Define _rules_xcodeproj_build for all commands, not only build

### Problem

Prior to this change, other commands (e.g. `fetch`) fail with an error, like:
```
$ bazel run //:bazel-project --   "--generator_output_groups=all_targets" "fetch //..."
...
ERROR: Config value '_rules_xcodeproj_build' is not defined in any .rc file
```

### DIFF
--- a/xcodeproj/internal/templates/xcodeproj.bazelrc
+++ b/xcodeproj/internal/templates/xcodeproj.bazelrc
@@ -111,7 +111,7 @@ build:rules_xcodeproj_swiftuipreviews --swiftcopt=-Xfrontend --swiftcopt=-enable
 # `rules_xcodeproj` instead. This config exists solely to allow project-level
 # configs to work correctly.
 
-build:_rules_xcodeproj_build --config=rules_xcodeproj
+common:_rules_xcodeproj_build --config=rules_xcodeproj
 
 ### Project specific configs
 %project_configs%
--- a/examples/integration/test/fixtures/bwb.xcodeproj/rules_xcodeproj/bazel/xcodeproj.bazelrc
+++ b/examples/integration/test/fixtures/bwb.xcodeproj/rules_xcodeproj/bazel/xcodeproj.bazelrc
@@ -111,7 +111,7 @@ build:rules_xcodeproj_swiftuipreviews --swiftcopt=-Xfrontend --swiftcopt=-enable
 # `rules_xcodeproj` instead. This config exists solely to allow project-level
 # configs to work correctly.
 
-build:_rules_xcodeproj_build --config=rules_xcodeproj
+common:_rules_xcodeproj_build --config=rules_xcodeproj
 
 ### Project specific configs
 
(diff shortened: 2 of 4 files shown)


xcodeproj/internal/templates/xcodeproj.bazelrc
```diff
@@ -111,7 +111,7 @@ build:rules_xcodeproj_swiftuipreviews --swiftcopt=-Xfrontend --swiftcopt=-enable
 # `rules_xcodeproj` instead. This config exists solely to allow project-level
 # configs to work correctly.
 
-build:_rules_xcodeproj_build --config=rules_xcodeproj
+common:_rules_xcodeproj_build --config=rules_xcodeproj
```
thiagohmcruz, line 114: I wonder if the `_build` suffix stops making sense since this change proposes to make it available to all? Maybe the original intent was to have it only in `build`? I'd confirm with @brentleyjones

brentleyjones: Adjusting https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/2063abaca378b5d6600f45bcfe78ae21e6898d33/xcodeproj/internal/templates/runner.sh#L213 is how I've handled this before. I guess I would want to know why this isn't preferable? And yes, the `_build` one is supposed to specifically be for when we are building, versus all the other actions. Those should use the baseline config instead.

polac24: For maintenance: bazel has 20+ commands and there might be new commands in the future that might require exclusion. 
A feature to ignore unknown params in common was added quite recently, in [6.3.0](https://github.com/bazelbuild/bazel/releases/tag/6.3.0) and I think it will simplify `rules_xcodeproj` codebase. To not mislead about `_build`, we could rename to `_rules_xcodeproj` but that will be a breaking change if someone relied on that private API. 

I can add `fetch` and `query`. those 2 I need but I can imagine someone might want e.g. `license` or `version`, which are not working now). 

If you are curious why do I need that for `fetch` - in my implementation to populate cache, I have two strategies (`standard` and `rules_xcodeproj`) and the flow calls `fetch` as a first step. Yes, I can fallback to the standard strategy for `fetch` command, but I am afraid that it would be hard to reason about (yet doable).

#### Another alternative
If we want to apply that only to build, then what about inverting the if and apply that only to "buildable"? However, the [list](https://bazel.build/run/bazelrc) is also quite long and not forward-compatible:
```
build, test, run, clean, mobile-install, info, print_action, config, cquery, aquery, coverage
```

brentleyjones: When I split these configs it was to ensure that people didn’t accidentally add changes to build only (it’s not a public config that people should depend on). That was before I added the command-line API. I need to think on how to best approach this.

brentleyjones: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2889

PR #2885, opened 2024-01-29
Remove `--noincompatible_objc_provider_remove_linking_info`

brentleyjones:
Not a valid option at HEAD, and no longer needed.


mattrobmattrob (approved):

brentleyjones:
Failure is an issue with difference between Bazel 6 and 7 for the fixtures. I'm looking to see if there is anything we can do about it.

PR #2898, opened 2024-02-09
Stop sending duplicate build target ids to `xcschemes`

brentleyjones:


brentleyjones:
I'm not sure this is worth the increased Starlark CPU (over the current increased `xcschemes` CPU time). It probably isn't,

thiagohmcruz (approved):

brentleyjones:
Not doing this for now.

PR #2899, opened 2024-02-12
Remove `platform_mappings`

brentleyjones:
The file was only required during the rules_apple 3.0 RC cycle. Removing this works around issues testing both Bazel 6 and Bazel HEAD.


brentleyjones:
The failures related to `@@rules_xcodeproj~override` versus `@@rules_xcodeproj~` are a Bazel HEAD change unrelated to this PR. I'll address that with a different change.

PR #2900, opened 2024-02-12
Strip version from fixture files

brentleyjones:
This causes them to match what is produced in Bazel HEAD, while allowing us to test Bazel 6 and 7.


luispadron (approved):

brentleyjones:
Going a different route, since it's hard to get the paths to match up correctly in the fixtures: https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2901.

PR #2905, opened 2024-02-13
Fix tree artifact `File` handling in incremental generation

### LINKED ISSUE #2904: Bug: 'ERROR: Internal precondition failure' raised when generating a project containing a target with more than one dynamically generated localizable resource
### Description

When generating a localizable resources directory (e.g. **fr.lproj**) using a custom rule, project generation works if the generated directory contains a single file, but fails if the generated directory contains more than one file.

A setup with a single file inside the generated directory works as expected:

```
fr.lproj
|- Localizable.strings
```

While a setup with more than one file inside the generated directory fails with the `Internal precondition failure` error:

```
fr.lproj
|- Localizable.strings
|- MyOtherLocalizable.strings
```

In both cases, directly building the app target which contains the generated localizable resource directory using `bazel build //Path/To:AppTarget` works as expected.

The error message looks like some kind of incorrect argument offset:

```
tools/generators/lib/PBXProj/src/ConsumeArgs.swift:168: "/Users/<username>/Developer/rules_xcodeproj/examples/integration/iOSApp/bazel-output-base/rules_xcodeproj.noindex/build_output_base/execroot/_main/bazel-out/darwin_arm64-dbg/bin/external/rules_xcodeproj~override~internal~rules_xcodeproj_generated/generator/xcodeproj-incremental-bazel-sim_arm64/xcodeproj-incremental-bazel-sim_arm64_pbxproj_partials/target_arguments_files/3": Failed to parse "iOSApp/Source/es.lproj/unprocessed.json" as Int.Type for <folder-resources-count>
```

### Reproduction steps

Issue is consistently reproducible in the example on [PR 2903](https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2903).

Steps to reproduce:

1. `cd examples/integration/iOSApp`
2. `bazel run //:xcodeproj-incremental-bazel-sim_arm64`

### Expected behavior

We should be able to successfully generate a project which uses a custom rule that generates more than one localizable resource file enclosed in a standard localizable resources directory like this:

```
fr.lproj
|- Localizable.strings
|- MyOtherLocalizable.strings
```

### rules_xcodeproj version

e181453a45b1e64674110aa06fa81c30dec6ed19

### Xcode version

15.1

### Bazel version

7.0.3

### rules_apple version

2.0.0

### rules_swift version

1.6.0

### Additional information

_No response_

brentleyjones:
Fixes #2904.

We fix our handling by using a null-character terminator instead of encoding the number of paths. This is because at analysis time we can’t rely on the number of elements in a `depset`, as they might expand to more elements at execution if any of the elements are tree artifacts.

Added an example by Wiley that exercises the previous failure case.

### DIFF
--- a/tools/generators/pbxnativetargets/src/Generator/TargetArguments.swift
+++ b/tools/generators/pbxnativetargets/src/Generator/TargetArguments.swift
@@ -80,19 +80,22 @@ extension Dictionary<TargetID, TargetArguments> {
                 try rawArgs.consumeArg("has-c-params", as: Bool.self, in: url)
             let hasCxxParams =
                 try rawArgs.consumeArg("has-cxx-params", as: Bool.self, in: url)
-            let srcs =
-                try rawArgs.consumeArgs("srcs", as: BazelPath.self, in: url)
-            let nonArcSrcs = try rawArgs.consumeArgs(
+            let srcs = try rawArgs.consumeArgsUntilNull(
+                "srcs",
+                as: BazelPath.self,
+                in: url
+            )
+            let nonArcSrcs = try rawArgs.consumeArgsUntilNull(
                 "non-arc-srcs",
                 as: BazelPath.self,
                 in: url
             )
-            let resources = try rawArgs.consumeArgs(
+            let resources = try rawArgs.consumeArgsUntilNull(
                 "resources",
                 as: BazelPath.self,
                 in: url
             )
-            let folderResources = try rawArgs.consumeArgs(
+            let folderResources = try rawArgs.consumeArgsUntilNull(
                 "folder-resources",
                 as: BazelPath.self,
                 in: url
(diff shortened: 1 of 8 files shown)


tools/generators/pbxnativetargets/src/Generator/TargetArguments.swift
```diff
@@ -80,19 +80,22 @@ extension Dictionary<TargetID, TargetArguments> {
...
+                as: BazelPath.self,
+                in: url
+            )
+            let nonArcSrcs = try rawArgs.consumeArgsUntilNull(
                 "non-arc-srcs",
                 as: BazelPath.self,
                 in: url
             )
-            let resources = try rawArgs.consumeArgs(
+            let resources = try rawArgs.consumeArgsUntilNull(
```
brentleyjones, line 93: I'll try to change other uses of `consumeArgs` to `consumeArgsUntilNull` in a future change. It has lower Starlark CPU use. After that, it might be the only version of `consumeArgs` remaining, in which I'll rename `consumeArgsUntilNull` to `consumeArgs`.

thiagohmcruz (approved):

luispadron (approved):

PR #2907, opened 2024-02-15
Add support for argument lists in `swiftc_stub`

marcosgriselli:
Xcode 15.3 Beta 2 introduced an argument list file for previews:

```
Fixed: Previews fail when Swift files require more than 4096 compiler arguments (92630764) (FB10004742)
```

This PR adds support for extracting the arguments from the argument list file while still supporting the legacy method:

image


tools/swiftc_stub/main.swift
```diff
@@ -166,11 +183,15 @@ if args.count == 2, args.last == "-v" {
 }
 
 for arg in args {
-    if arg.hasSuffix(".preview-thunk.swift"),
+    if arg.hasSuffix(".preview-thunk.swift") || isPreviewsArgumentList(arg: arg),
         !argsSet.contains("-output-file-map")
     {
+        var arguments = args
+        if isPreviewsArgumentList(arg: arg) {
+            arguments = extractArgumentsFromArgumentListFile(arg)
```
brentleyjones, line 191: Why do we have to extract it? Can't we just pass through the arguments as they are passed (which means just passing through the argument list file)? Isn't the main issue that we couldn't detect that it was a thunk compilation, and we just need to adjust to to check for the existence of this file in addition to the check we had before?

(I ask all of this without running this change, which I can do later.)

brentleyjones: I believe if we extract the arguments we can run into the same command-line length limit that they are fixing.

brentleyjones: (We could also expand all response files and leave the check as is, but against the expanded args, but then pass through the original args.)

marcosgriselli: Thanks for pointing these out. I updated the PR to only expand the arguments from the list file when attempting to find the sdk path. Everything still works on the latest Xcode stable release and 15.3 beta 3.

brentleyjones: Thanks!

If you don't mind I would like to tweak it some before merging, but that won't require any additional work on your end.

marcosgriselli: Sure! Feel free to modify it as much as you see fit

brentleyjones (approved):

PR #2920, opened 2024-02-26
Fix tvOS Xcode preview framework code signing with incremental generation

brentleyjones:
Seems tvOS frameworks need to be codesigned for Xcode Previews.

### DIFF
--- a/tools/generators/pbxnativetargets/src/Generator/CalculateSharedBuildSettings.swift
+++ b/tools/generators/pbxnativetargets/src/Generator/CalculateSharedBuildSettings.swift
@@ -99,15 +99,17 @@ extension Generator.CalculateSharedBuildSettings {
                     value: label.description.pbxProjEscaped
                 )
             )
-        } else {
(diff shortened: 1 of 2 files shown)


tools/generators/pbxnativetargets/src/Generator/CalculateSharedBuildSettings.swift
```diff
@@ -99,15 +99,17 @@ extension Generator.CalculateSharedBuildSettings {
...
-        } else {
-            // Used to work around CODE_SIGNING_ENABLED = YES for resource
-            // bundles in Xcode 14+
-            buildSettings.append(
-                .init(key: "CODE_SIGNING_ALLOWED", value: "NO")
-            )
         }
 
-        if productType == .uiTestBundle {
+        if productType == .framework {
```
thiagohmcruz, line 104: Just to clarify. You mentioned tvOS but IIUC this is being applied to any platform correct?

Btw, TIL that code signing was needed 🤔

thiagohmcruz (approved):

PR #2922, opened 2024-02-26
Add support for visionOS

### LINKED ISSUE #2352: Feature Request: Support visionOS
https://github.com/bazelbuild/rules_apple/issues/2012

chriscraws:
Resolves #2352.

Some notes: apple is a little inconsistent with "visionos" vs "xros" but this worked for generating a test project for myself.


test/utils.bzl
```diff
@@ -6,6 +6,8 @@ _PLATFORM_NAME = {
     "MACOS": "macosx",
     "TVOS_DEVICE": "appletvos",
     "TVOS_SIMULATOR": "appletvsimulator",
+    "XROS_DEVICE": "xros",
+    "XROS_SIMULATOR": "xrsimulator",
```
brentleyjones, line 10: Pretty sure some of these (most of them except those that line up with SDK name?) should be `visionos` instead. For example, the key here should match https://github.com/bazelbuild/bazel/blob/6889d08a2f0b64eb9fe9518793ce572489b6011b/src/main/java/com/google/devtools/build/lib/rules/apple/ApplePlatform.java#L40-L41

brentleyjones: Hmm, a new project from Xcode in 15.2.0 still shows a lot of xrOS stuff. Odd. Minimally we need the Bazel names to match, and in the code I want to refer to it as `visionOS`.

brentleyjones: (So in this example, `xrsimulator` stays, but `XROS_SIMULATOR` needs to be `VISIONOS_SIMULATOR`.)

chriscraws: Makes sense, and thanks for the pointer, I wasn't sure where those strings were coming from. I went ahead and updated everything to visionOS where possible. If this looks good I can take a stab at updating the tests.

brentleyjones: Yep, looks good.

brentleyjones (approved):
Approval to run CI

brentleyjones:
You'll need to sign-off on the commit(s) to pass the DCO check.

chriscraws:
> You'll need to sign-off on the commit(s) to pass the DCO check. I recommend squashing all of these existing ones into a single commit for that.

Done, also realized I hadn't updated the transition code, so there are a few more changes - see 

xcodeproj/internal/fixtures.bzl
xcodeproj/internal/xcodeproj_macro.bzl
xcodeproj/internal/xcodeproj_runner.bzl
xcodeproj/internal/xcodeproj_transitions.bzl

xcodeproj/internal/xcodeproj_macro.bzl
```diff
@@ -306,6 +308,28 @@ def xcodeproj(
...
+            **Warning:** Changing this value will affect the Starlark transition
+            hash of all transitive dependencies of the targets specified in the
+            `top_level_targets` argument with the `"device"`
+            `target_environment`, even if they aren't visionOS targets.
+        visionos_simulator_cpus: Optional. The value to use for `--visionos_cpus`
+            when building the transitive dependencies of the targets specified
+            in the `top_level_targets` argument with the `"simulator"`
+            `target_environment`.
+
+            If no value is specified, it defaults to the simulator cpu that goes
```
brentleyjones, line 325: There is only an arm64 simulator, so this should not have the defaulting logic like the others.

xcodeproj/internal/xcodeproj_transitions.bzl
```diff
@@ -62,6 +66,13 @@ def make_xcodeproj_target_transitions(
             else:
                 tvos_cpus = "x86_64"
 
+        visionos_cpus = attr.visionos_simulator_cpus
+        if not visionos_cpus:
+            if cpu_value == "darwin_arm64":
+                visionos_cpus = "sim_arm64"
```
brentleyjones, line 72: Same comment. The default value should just be `sim_arm64`.

xcodeproj/internal/xcodeproj_runner.bzl
```diff
@@ -570,6 +572,8 @@ xcodeproj_runner = rule(
         "tvos_simulator_cpus": attr.string(),
         "unfocused_labels": attr.string_list(default = []),
         "unowned_extra_files": attr.string_list(),
+        "visionos_device_cpus": attr.string(mandatory = True),
+        "visionos_simulator_cpus": attr.string(),
```
brentleyjones, line 576:
```suggestion
        "visionos_simulator_cpus": attr.string(mandatory = True),
```

brentleyjones (approved):

brentleyjones:
Seems like we need to use `minimum_os_version` instead of `visionos_minimum_os`.

chriscraws:
> Seems like we need to use `minimum_os_version` instead of `visionos_minimum_os`.

updated

brentleyjones (approved):

xcodeproj/internal/xcodeproj_runner.bzl
```diff
@@ -570,6 +572,8 @@ xcodeproj_runner = rule(
         "tvos_simulator_cpus": attr.string(),
         "unfocused_labels": attr.string_list(default = []),
         "unowned_extra_files": attr.string_list(),
+        "visionos_device_cpus": attr.string(mandatory = True),
+        "visionos_simulator_cpus": attr.string(mandatory = True),
```
thiagohmcruz, line 576: 👋 Mind clarifying why the other `*_simulator_cpus` are not mandatory and this one is?

brentleyjones: Mandatory means that it has to have a value. The others don't need a value because `None` is used to determine the simulator CPU based on the host CPU. That doesn't happen for visionOS, since it only has an arm64 simulator.

thiagohmcruz: Oh right, that makes sense to me, thx I just wanted to sanity check I was understanding the changes

thiagohmcruz (approved):

brentleyjones:
I'm looking at the CI failures

chriscraws:
> > Seems like we need to use `minimum_os_version` instead of `visionos_minimum_os`.
> 
> updated

Actually, does it make sense to include this at all? I updated without thinking much, but seems "minimum_os_version" with a default of 1.0 is not particularly useful to 

> I'm looking into the CI failures

Fixed a couple issues - I do have a question though about the "minimum_os_version", I'm not familiar with how the fixtures are used but it seems like it's a bit blunt to set that value to 1.0 for generating the fixtures generally. Does it make sense to include it at all?

brentleyjones (approved):

brentleyjones:
It makes sense, because soon once 1.1 is available and the default, we want the fixtures to set it to 1.0.

brentleyjones (approved):

brentleyjones:
Thanks @chriscraws!

chriscraws:
> It makes sense, because soon once 1.1 is available and the default, we want the fixtures to set it to 1.0.

I see - just seems like that flag would apply to more than visionOS, but the others have coverage over everything else so I'm sure it's fine. Thank you for reviewing and helping me through this!

PR #2928, opened 2024-02-28
Set `explicitFileType` for bazel and bzl extensions

andre-alves:
BUILD.bazel and .bzl files are not being recognized as a text.script.python in Xcode when using the incremental generator.

Replacing lastKnownFileType with explicitFileType for these extensions fixed the issue.

| Before  | After |
| ------------- | ------------- |
| image  | image |

### DIFF
--- a/tools/generators/files_and_groups/test/ElementCreator/CreateFileElementTests.swift
+++ b/tools/generators/files_and_groups/test/ElementCreator/CreateFileElementTests.swift
@@ -178,9 +178,9 @@ final class CreateFileElementTests: XCTestCase {
     func test_element_content_lastKnownFileType_file() {
         // Arrange
 
-        let name = "node_name.bazel"
-        let ext = "bazel"
-        let bazelPath = BazelPath("a/bazel/path/node_name.bazel")
+        let name = "node_name.foo"
+        let ext = "foo"
+        let bazelPath = BazelPath("a/bazel/path/node_name.foo")
         let createAttributes = ElementCreator.CreateAttributes.stub(
             elementAttributes: ElementAttributes(
                 sourceTree: .group, name: nil, path: "a_path"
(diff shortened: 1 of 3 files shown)


mattrobmattrob (commented):
Is it possible to make the `BUILD.bazel` files hit this same check here? That seems more in-line with the current functionality.

https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/3edd5d90510a2d592f7814dfacf639d152cdc2f3/tools/generators/files_and_groups/src/ElementCreator/CreateFileElement.swift#L93-L95

tools/generators/files_and_groups/test/ElementCreator/CreateFileElementTests.swift
```diff
@@ -175,21 +175,21 @@ final class CreateFileElementTests: XCTestCase {
 
     // MARK: element.content - lastKnownFileType
 
-    func test_element_content_lastKnownFileType_file() {
```
brentleyjones, line 178: Instead of deleting this, can you change the extension to something else?

brentleyjones:
Similar to @mattrobmattrob's question, maybe we change it up a bit:

```swift
let impliedExtension = ext ?? Xcode.impliedExtension(basename: name)
let fileTypeType = calculateFileTypeType(basename: name, extension: impliedExtension)
let fileType = Xcode.pbxProjEscapedFileType(extension: impliedExtension)!
```

`Xcode.impliedExtension` would turn `BUILD` and `Podfile` into `bazel` and `rb`. `calculateFileTypeType` would turn `bazel` and `bzl` extensions or `Podfile` basename into `"explicitFileType"` otherwise `"implicitFileType"`.

Thoughts?

tools/generators/files_and_groups/test/ElementCreator/CreateFileElementTests.swift
```diff
@@ -178,9 +178,9 @@ final class CreateFileElementTests: XCTestCase {
     func test_element_content_lastKnownFileType_file() {
         // Arrange
 
-        let name = "node_name.bazel"
-        let ext = "bazel"
-        let bazelPath = BazelPath("a/bazel/path/node_name.bazel")
+        let name = "node_name.foo"
```
brentleyjones, line 181: nit: Sorry, but can we also add a test that has a known extension, so we verify the non-`"file"` path?

andre-alves: Sure, added one for ruby!

brentleyjones (approved):

PR #2929, opened 2024-03-01
Work around when bazelisk is called recursively during project generation

brentleyjones:
Since bazelisk only supports one path for the wrapper script, and `BAZELISK_SKIP_WRAPPER` is only set when the wrapper script is used, we currently can safely default this value.


thiagohmcruz (approved):
We've got a lot going on in our `tools/bazel` so ran some tests with this branch but nothing came up. LGTM!

mattrobmattrob (approved):

PR #2930, opened 2024-03-01
Stop setting `--experimental_remote_download_regex` by default with command-line API

brentleyjones:
Also fixes the regex used to match what we changed `build_build.sh` to use.

This is to optimize bandwidth usage of cache warming calls.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -316,6 +316,23 @@ For example, this will build all targets the same way that Xcode does:
 bazel run //:xcodeproj -- --generator_output_groups=all_targets build
 ```
 
+### `--download_intermediates`
+
+By default, when using
+[`--generator_output_groups`][#--generator_output_groups],
+intermediate outputs (such as generated files and index stores) won't be
+downloaded from a remote cache if `--remote_download_toplevel` or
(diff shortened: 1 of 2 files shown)


docs/usage.md
```diff
@@ -316,6 +316,23 @@ For example, this will build all targets the same way that Xcode does:
 bazel run //:xcodeproj -- --generator_output_groups=all_targets build
 ```
 
+### `--download_intermediates`
```
brentleyjones, line 319: I'm not set on this name, if something else sounds better.

thiagohmcruz: Seems like a reasonable name to me

thiagohmcruz (approved):

mattrobmattrob (approved):

mattrobmattrob:
Have any sense for the impact on a larger build's cache download before vs. after this change? I don't have a sense for how much these intermediates add in terms of download for a ~fully cached build.

brentleyjones:
I don't. But the number of files (e.g. index stores) can be slow to write regardless. Just trying to make cache warming as efficient as possible.

PR #2936, opened 2024-03-07
Remove `TARGET_DEVICE_PLATFORM_NAME` matching from `Info.plist` touching

sebastianv1:
It appears that the `TARGET_DEVICE_PLATFORM_NAME` check didn't account for `iphonesimulator` platform name which was causing unit tests running on a test host to not reflect changes made in source code.


brentleyjones:
Hmm, this was specifically set to that because the change was only needed to fix a device deployment issue. Are you saying that the same issue exists on the simulator?

brentleyjones:
Does this issue also happen for tvOS? Maybe we need to remove the `TARGET_DEVICE_PLATFORM_NAME` check all together?

sebastianv1:
@brentleyjones Thats correct; it appears to happen on simulator as well. Devs have been reporting that source code changes aren’t reflected in test host target apps.

sebastianv1:
> Does this issue also happen for tvOS? Maybe we need to remove the `TARGET_DEVICE_PLATFORM_NAME` check all together?

Haven't tried with tvOS, but logically it makes sense to apply to all platforms?

brentleyjones:
Yes please. Also you need to sign-off on your git commit to pass the `DCO` check.

brentleyjones (approved):

mattrobmattrob (approved):

PR #2950, opened 2024-03-12
Add example for `*_build_test` in `xcschemes.top_level_anchor_target`

brentleyjones:
Also works around a bug for libraries that merged into other targets.


brentleyjones:
I'm filing a bug after this is merged.

thiagohmcruz (approved):

PR #2956, opened 2024-03-18
Reduce work that incremental generation mode installer does for generated directories

brentleyjones:
A couple of optimizations:

- We create the list of generated directories with Bazel, allowing it to be cached
- The generated directories are the unique leaf directories, no longer including any intermediate parent directories. This is accomplished using a trie.


brentleyjones:
@MobileNativeFoundation/rules_xcodeproj-maintainers PTAL

thiagohmcruz (approved):

PR #2960, opened 2024-03-21
Stop collecting uncategorized files for unfocused targets

brentleyjones:
Besides for being more memory efficient, this actually fixes an edge case where before some extra files were added for unfocused unsupported targets (e.g. a modulemap like rule that is referenced in a categorized attribute).

In the future we should probably rethink the “uncategorized files become extra files if the target is listed in a categorized attribute downstream” logic. I believe it was originally to show inputs of custom code gen rules, but I think that can be handled differently.


ycho-sc (approved):
Thanks for the fix!

yongjincho92 (approved):

PR #2964, opened 2024-03-22
Set `--noincompatible_struct_has_no_methods` until our dependencies are fixed

brentleyjones:


bazel_8.bazelrc
```diff
@@ -0,0 +1,2 @@
+# Disabled until new rules_aplpe and rules_swift releases
```
mattrobmattrob, line 1:
```suggestion
# Disabled until new rules_apple and rules_swift releases
```

mattrobmattrob (approved):

PR #2969, opened 2024-03-22
Stop collecting uncategorized files for source-less library targets

brentleyjones:
`collect_uncategorized_files` is determined by the type of rule, not related to if we should create an Xcode target. Related, we should rename `is_supported` to `can_generate_xcode_target` or something in a future change.


thiagohmcruz (approved):
I don't have a better suggestion than `can_generated_xcode_target`, regardless I think this name states clearly what it does so sounds reasonable.

PR #2972, opened 2024-03-22
Collect `additional_*_inputs` as extra files in incremental generation mode

brentleyjones:


mattrobmattrob (approved):
Should we have a test for this functionality since it's somewhat niche?

brentleyjones:
We don't have unit tests around input collection. But it all falls under the extra_files collection which is exercised in our examples.

PR #2985, opened 2024-04-02
Document `xcodeproj.generation_mode`

brentleyjones:


mattrobmattrob (approved):
Is BwX already removed/deprecated? If not, please add that caveat between the two modes.

mattrobmattrob (approved):

PR #2992, opened 2024-04-04
Fix don't attempt to get extension_host for launch_path

luispadron:
The `launch_path` does not have an `extension_host` attr

### DIFF
--- a/xcodeproj/internal/xcschemes/xcscheme_infos.bzl
+++ b/xcodeproj/internal/xcschemes/xcscheme_infos.bzl
@@ -72,6 +72,15 @@ def _make_launch_target(
     )
 
 def _make_same_as_run_launch_target(run_launch_target):
+    if run_launch_target.is_path == TRUE_ARG:
+        return struct(
+            is_path = TRUE_ARG,
+            path = run_launch_target.path,
+            post_actions = EMPTY_LIST,
+            pre_actions = EMPTY_LIST,
+            working_directory = run_launch_target.working_directory,
+        )
+
     return struct(
         extension_host = run_launch_target.extension_host,
         id = run_launch_target.id,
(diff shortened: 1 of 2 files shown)


xcodeproj/internal/xcschemes/xcscheme_infos.bzl
```diff
@@ -73,7 +73,7 @@ def _make_launch_target(
 
 def _make_same_as_run_launch_target(run_launch_target):
     return struct(
-        extension_host = run_launch_target.extension_host,
+        extension_host = None if run_launch_target.is_path else run_launch_target.extension_host,
```
thiagohmcruz, line 76: Nit: Sorry if I misunderstood the intent here but IIUC the idea is to set it only if `extension_host` exists correct?  In that case wouldn't `hasattr` be a more direct check here instead of an indirect check via `.is_path`?

thiagohmcruz (approved):

xcodeproj/internal/xcschemes/xcscheme_infos.bzl
brentleyjones: That's more expensive though. `is_path` is a type check. We did a similar thing elsewhere.

thiagohmcruz: Oh ok SGTM, thx for clarifying!

luispadron:
🤔 somehow a private PR mentioning this PR closed it? Weird... Working on tests now

xcodeproj/internal/xcschemes/xcscheme_infos.bzl
luispadron: updated this

brentleyjones (approved):

PR #2999, opened 2024-04-06
Fix Xcode Previews in App Extensions

andre-alves:
Fix #2997.

PR #3005, opened 2024-04-09
Adjust target dependencies for transitive Xcode Preview targets instead of schemes

### LINKED ISSUE #2997: Bug: Xcode Previews failing in extensions
### Description

I'm not sure if it's a recent regression or not, but Xcode Previews is not working with extensions:

```
The app extension ”WidgetExtension.appex” needs to be embedded in an app in order to use previews
```

image

### Reproduction steps

It's possible to reproduce using the integration example without any changes:

- `cd examples/integration`
- `bazel run //:xcodeproj-incremental-bazel-sim_arm64`
- `open Integration.xcodeproj`
- Select  scheme`iOSApp` or `WidgetExtension in iOSApp` 
- Open `WidgetExtension.swift` and check previews

### Expected behavior

Xcode Previews to display.

### rules_xcodeproj version

main (de3089fd2612fc97299e87955975e1dc34049b67)

### Xcode version

15.2

### Bazel version

7.1.0

### rules_apple version

3.5.0

### rules_swift version

1.18.0

### Additional information

_No response_

brentleyjones:
Fixes #2997.

This fixes errors similar to `The app extension ”WidgetExtension.appex” needs to be embedded in an app in order to use previews`. The same number of targets are “built” in Xcode, but we tell Xcode about it in a different way.


xcodeproj/internal/xcodeproj_macro.bzl
```diff
@@ -19,6 +19,7 @@ def _normalize_build_setting(flag):
 def xcodeproj(
         *,
         name,
+        adjust_dependencies_for_xcode_previews = True,
```
brentleyjones, line 22: Should this even be configurable? For schemes it made sense, I'm not sure it makes sense here.

luispadron (approved):

xcodeproj/internal/xcodeproj_macro.bzl
luispadron: Would it be possible to set a global default without it? fwiw we haven't needed to set this to `False` any longer need to use this as we fixed our top-level target issues

brentleyjones: I'm proposing that there would be no need to disable this. Unlike the scheme way, I can't think of a reason, outside of potential small performance gains, to disable it.

andre-alves:
Just tested here and it fixed Xcode Previews for me, but I noticed the same issue I had [here](https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2999#issuecomment-2041125265) in regards to focus mode.

brentleyjones:
Ahh, I forgot about that. I'll address that.

tools/generators/pbxtargetdependencies/src/Generator/ConsolidateTargets.swift
```diff
@@ -140,11 +140,11 @@ extension Generator.ConsolidateTargets {
...
             _ depID: TargetID,
             for id: TargetID
-        ) throws -> PotentialConsolidatedTargetKey {
+        ) throws -> PotentialConsolidatedTargetKey? {
             guard let dependencyKey = targetIDMapping[depID] else {
-                throw PreconditionError(message: """
-Target "\(id)" dependency on "\(depID)" not found in `targetIDMapping`
-""")
+                // We don't throw here because the dependency was probably a
+                // "potential test host" that became unfocused
```
brentleyjones, line 146: I believe this is more performant than popping a depset on the Starlark side. I think I might be able to optimize this though, instead of having it done in 3 places.

brentleyjones:
@andre-alves unfocusing should work now.

andre-alves:
Fixed!

tools/generators/pbxtargetdependencies/src/Generator/ConsolidateTargets.swift
brentleyjones: It's too much of a hassle to pre-filter it. Going to leave it like this.

Issue #3009, opened 2024-04-10
Bug: Only the one associated extra file for a target is added to the project


CognitiveDisson:
### Description

If you specify multiple associated extra files with one target, only the last one will be presented in the Xcode project.
Fo example:
```
ASSOCIATED_EXTRA_FILES = {
    "//iOSApp/Source:iOSApp": ["//iOSApp:ownership.yaml", "//iOSApp:info.yaml"],
}
```
The Xcode project will contain info.yaml, but it won't include ownership.yaml.

### Reproduction steps

You can reproduce this on example project.

1. Create the `examples/integration/iOSApp/ownership.yaml` file.
2. Export it in `examples/integration/iOSApp/BUILD` : `exports_files(["ownership.yaml", "info.yaml"])`.
3. Add it to the `ASSOCIATED_EXTRA_FILES` list: `"//iOSApp/Source:iOSApp": ["//iOSApp:ownership.yaml", "//iOSApp:info.yaml"],`.
4. Generate a project using the command: `bazelisk run --config=cache //:xcodeproj-incremental-bazel-sim_arm64`.
5. Open the generated project: `xed Integration.xcodeproj`.
6. Check that the project does not include `ownership.yaml`.

### Expected behavior

All associated extra files presented in the project.

### rules_xcodeproj version

516c3dd1f2eff13a3a2077c02a9e5aa914854aba

### Xcode version

15.1

### Bazel version

7.1

PR #3010, opened 2024-04-10
Fix the issue with missing associated extra files

CognitiveDisson:
Fix for #3009. 
While this approach of converting a depset to a list and back to a depset may not be optimal, it seems the only way.
Dictionary comprehension is also an option, although it is less readable. As far as I know, `defaultdict` is not available in Starlark.

I don't insist on merging this implementation if there's a better way. Feel free to close this PR.


brentleyjones:
Thanks for the PR. I'm going to fix it in a slightly different way soon.

PR #3019, opened 2024-04-23
Convert `calculate_output_groups.py` into a Swift binary

pennig:
Performance is better than (but comparable to) the python script. Perhaps more importantly, the program looks specifically for the _build-request.json_ file and uses its modification date to compare against the _build_marker_ file's date, instead of comparing the parent _.xcbuilddata_ directory. Not sure if that'll make a difference, but it's a distinction worth calling out.


thiagohmcruz:
👋 Quick ping to check if you intend to continue the work here? Asking because we're seeing build delays due this `.json` file not being ready yet and I believe this fix will improve things for us. Happy to help landing this too if you need!

PR #3020, opened 2024-04-29
Respect "Continue building after errors" setting

dflems:
Add the `--keep_going` flag to builds when the "Continue building after errors" flag is set. Maybe there's a better place to add this, but I think it would be nice to respect the setting in the IDE.

pasted_image_4_29_24__11_45___am_720

### DIFF
--- a/xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
+++ b/xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
@@ -140,6 +140,8 @@ fi
 continue_building_value="$(defaults read com.apple.dt.Xcode IDEBuildingContinueBuildingAfterErrors 2>/dev/null)"
 if [ "$continue_building_value" == "1" ]; then
   build_pre_config_flags+=("--keep_going")
+else
+  build_pre_config_flags+=("--nokeep_going")
 fi
 
 # Runtime Sanitizers
--- a/xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
+++ b/xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
@@ -136,6 +136,12 @@ else
   readonly config="_${BAZEL_CONFIG}_build"
 fi
 
+# Respect the "Continue building after errors" setting
+continue_building_value="$(defaults read com.apple.dt.Xcode IDEBuildingContinueBuildingAfterErrors 2>/dev/null)"
+if [ "$continue_building_value" == "1" ]; then
+  build_pre_config_flags+=("--keep_going")
+fi
+
 # Runtime Sanitizers
 if [[ $apply_sanitizers -eq 1 ]]; then
   if [ "${ENABLE_ADDRESS_SANITIZER:-}" == "YES" ]; then


xcodeproj/internal/templates/bazel_build.sh
```diff
@@ -113,13 +113,18 @@ if [[ -n "${TOOLCHAINS-}" ]]; then
   fi
 fi
 
+# Respect "Continue building after errors" setting
+continue_building_value="$(defaults read com.apple.dt.Xcode IDEBuildingContinueBuildingAfterErrors 2>/dev/null)"
+[ "$continue_building_value" == "1" ] && continue_building_flag="--keep_going" || continue_building_flag="--nokeep_going"
```
brentleyjones, line 118: Can you add this to `build_pre_config_flags` instead?

xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
```diff
@@ -136,6 +136,12 @@ else
   readonly config="_${BAZEL_CONFIG}_build"
 fi
 
+# Respect the "Continue building after errors" setting
+continue_building_value="$(defaults read com.apple.dt.Xcode IDEBuildingContinueBuildingAfterErrors 2>/dev/null)"
+if [ "$continue_building_value" == "1" ]; then
+  build_pre_config_flags+=("--keep_going")
+fi
+
```
dflems, line 144: @brentleyjones Is this the right place?

xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
```diff
@@ -136,6 +136,12 @@ else
   readonly config="_${BAZEL_CONFIG}_build"
 fi
 
+# Respect the "Continue building after errors" setting
+continue_building_value="$(defaults read com.apple.dt.Xcode IDEBuildingContinueBuildingAfterErrors 2>/dev/null)"
+if [ "$continue_building_value" == "1" ]; then
+  build_pre_config_flags+=("--keep_going")
```
luispadron, line 142: What happens if someone is already setting this in their bazelrc?

xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
brentleyjones: Yep

brentleyjones (approved):

xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
dflems: `--keep_going` / `--no_keep_going` seems to be one of those flags you can set as many times as you want (last in wins). Since this is set before the bazelrcs are applied, whatever is in the bazelrc should win.

brentleyjones: It overrides globals, but allows the `rules_xcodeproj` config to override it.

dflems:
@brentleyjones Re-pushed with the signoff thing

brentleyjones (approved):

xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
luispadron: Would it be possible then to set this Xcode setting the other way then? Right now the setting will only match if it's set in Xcode right?

luispadron (approved):

xcodeproj/internal/bazel_integration_files/generate_bazel_dependencies.sh
dflems: I could add an `else` with `--nokeep_going` if you'd prefer @luispadron

brentleyjones: Yeah, I think we should have it set both states, so it matches exactly what Xcode says, unless someone overrides it in the `rules_xcodeproj` config.

dflems:
Can someone help explain why the commit message style GH action is [unhappy](https://github.com/MobileNativeFoundation/rules_xcodeproj/actions/runs/8885279497/job/24396278201?pr=3020) with how I worded my message? 😅

It looks like "respect" should be allowed?

https://github.com/mristin/opinionated-commit-message/blob/f3b9cec249cabffbae7cd564542fd302cc576827/src/mostFrequentEnglishVerbs.ts#L721

PR #3026, opened 2024-05-13
Fix Bazel wrapper check for case-sensitive filesystem

### LINKED ISSUE #3025: Bug: `bazel_path` breaks when used in a case-sensitive file system when Bazel wrapper is in `Tools/bazel`
### Description

We keep our wrapper in a **uppercased** `Tools/bazel` directory. In a case-insensitive file system the check done here: https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/7683faee63343c3410c4c40ea13f5adc5bae43b7/xcodeproj/internal/templates/runner.sh#L93 works fine. However, on a case sensitive file system the logic breaks and leads to other issues (we validate this is set in our wrapper and fail the build)

### Reproduction steps

- Have `Tools/bazel` wrapper in a case sensitive file system
- Set `bazel_path` to `./Tools/bazel`
- See issue when generating a project.

### Expected behavior

This works for both `Tools/bazel` and `tools/bazel`

### rules_xcodeproj version

2.2.0

### Xcode version

15.1

### Bazel version

6.5.0

### rules_apple version

_No response_

### rules_swift version

_No response_

### Additional information

_No response_

luispadron:
Fixes #3025

### DIFF
--- a/xcodeproj/internal/templates/runner.sh
+++ b/xcodeproj/internal/templates/runner.sh
@@ -88,18 +88,26 @@ if [[ -z "$bazel_path" ]]; then
   exit 1
 fi
 
+# Check that either `tools/bazel` or `Tools/bazel` exist
+bazel_wrapper_path=""
+if [[ -f "tools/bazel" ]]; then
+  bazel_wrapper_path="./tools/bazel"
+elif [[ -f "Tools/bazel" ]]; then
+  bazel_wrapper_path="./Tools/bazel"
+fi
+
 if [[ "${BAZEL_REAL:-}" == "$bazel_path" && \
       -n "${BAZELISK_SKIP_WRAPPER:-}" && \
-      -f "tools/bazel" ]]
+      -n "$bazel_wrapper_path" ]]
 then
   echo "WARNING: 'bazelisk' was called recursively while a \"tools/bazel\"" \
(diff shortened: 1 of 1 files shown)


thiagohmcruz (approved):

luispadron:
Updated this a bit more

xcodeproj/internal/templates/runner.sh
```diff
@@ -88,18 +88,26 @@ if [[ -z "$bazel_path" ]]; then
...
+if [[ -f "tools/bazel" ]]; then
+  bazel_wrapper_path="./tools/bazel"
+elif [[ -f "Tools/bazel" ]]; then
+  bazel_wrapper_path="./Tools/bazel"
+fi
+
 if [[ "${BAZEL_REAL:-}" == "$bazel_path" && \
       -n "${BAZELISK_SKIP_WRAPPER:-}" && \
-      -f "tools/bazel" ]]
+      -n "${bazel_wrapper_path:-}" ]]
```
brentleyjones, line 101: Since we always assign to this, we don't need the conditional expansion:
```suggestion
      -n "$bazel_wrapper_path" ]]
```

luispadron: Ah i thought an empty string is still undefined? Similar to `foo=()`

luispadron: nvm just tested, and its fine

PR #3028, opened 2024-05-16
Fix mistranslation of external sources

rockbruno:
Remapping of external sources in the index is done by replacing the path with `xcodeOutputBase`, which is built by removing paths from `xcodeExecutionRoot`. However, when `split()` is called, the leading `/` from `xcodeExecutionRoot` is lost, which was causing `index-import` to not translate these files correctly.

Screenshot 2024-05-16 at 15 57 02

### DIFF
--- a/tools/import_indexstores/ImportIndexstores.swift
+++ b/tools/import_indexstores/ImportIndexstores.swift
@@ -127,7 +127,7 @@ Intermediates\.noindex/Previews/[^/]*/Intermediates\.noindex
             }
 
             let xcodeOutputBase = xcodeExecutionRoot
-                .split(separator: "/")
+                .split(separator: "/", omittingEmptySubsequences: false)
                 .dropLast(2)
                 .joined(separator: "/")
 


tools/import_indexstores/ImportIndexstores.swift
```diff
@@ -126,11 +126,18 @@ Intermediates\.noindex/Previews/[^/]*/Intermediates\.noindex
...
+            let baseXcodeExecutionRoot = xcodeExecutionRoot
                 .split(separator: "/")
                 .dropLast(2)
                 .joined(separator: "/")
 
+            let xcodeOutputBase: String
+            if xcodeExecutionRoot.hasPrefix("/") {
+                xcodeOutputBase = "/" + baseXcodeExecutionRoot
+            } else {
+                xcodeOutputBase = baseXcodeExecutionRoot
```
rockbruno, line 138: Unsure if this condition is ever possible, but added it just in case. Otherwise we can simplify this even further

brentleyjones:
You'll need to sign-off on your git commit to pass the DCO check.

luispadron (approved):

BalestraPatrick (approved):

brentleyjones:
Thanks @rockbruno!

PR #3046, opened 2024-06-21
Fix incremental project generation swift debugging

yongjincho92:
Incremental project generation mode was not setting swift debug settings when top level target (e.g. ios application) is an Obj-c target. This change should fix the issue. 

Before : lldb > `settings show target.swift-extra-clang-flags` is empty 
After : lldb > `settings show target.swift-extra-clang-flags` is not empty

### DIFF
--- a/tools/generators/target_build_settings/src/Generator/ProcessSwiftArgs.swift
+++ b/tools/generators/target_build_settings/src/Generator/ProcessSwiftArgs.swift
@@ -104,6 +104,50 @@ extension Generator.ProcessSwiftArgs {
         frameworkIncludes: OrderedSet<String>,
         swiftIncludes: OrderedSet<String>
     ) {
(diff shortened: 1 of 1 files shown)


tools/generators/target_build_settings/src/Generator/ProcessSwiftArgs.swift
```diff
@@ -114,9 +158,8 @@ extension Generator.ProcessSwiftArgs {
         guard let tool = try await iterator.next(),
               tool != Generator.argsSeparator
         else {
-            return (false, [], [], [])
+            return (false, [], [], [], [])
         }
-        _ = try await iterator.next()
```
brentleyjones, line 119: Why was this removed?

yongjincho92: oops, removed by mistake. let me put this back

brentleyjones (approved):

PR #3049, opened 2024-07-08
Co-locate generated files

### LINKED ISSUE #3037: Feature Request: Co-locate Bazel Generated Files with targets

brentleyjones:
Resolves #3037.


tools/generators/files_and_groups/src/Generator/CalculatePathTree.swift
```diff
@@ -1,97 +1,183 @@
...
+        var generatedPathsByPackageAndConfig:
+            [BazelPath: [String: [BazelPath]]] = [:]
+        for generatedPath in generatedPaths {
+            generatedPathsByPackageAndConfig[
+                generatedPath.package,
+                default: [:]
+            ][generatedPath.config, default: []].append(generatedPath.path)
+        }
+
+        // FIXME: Do this in parallel
```
sebastianv1, line 18: Is this still relevant/are you looking to parallelize the for loop? For reference, I did some basic benchmarking and found this branch the same and even marginally faster.

brentleyjones: Yeah, I'll still probably do this is parallel. Just as one of the last FIXME. And I might do

erikkerber:
This fixes those "top-level" generated folders from our internal upstream project, and looks pretty good.

We have enough configs with sim+device times 3 product variants that there are 6 generated folders per library, so it's not a silver arrow for making things drop-dead-easy for Apple devs not fluent in Bazel, but it's also a big improvement over what we had previously.

sebastianv1:
What if we added an option to map a configuration names to something more idiomatic/familiar? `ios_arm64-dbg-ios-arm64-min15.0-applebin_ios-ST-a79286d91b1b: iOS 15 Device Build`. It's manual & brittle as configurations change over time but is pretty flexible in naming?

tools/generators/files_and_groups/src/Generator/CalculatePathTree.swift
```diff
@@ -175,7 +175,13 @@ extension Generator {
                         children: children
                     )
                 case .generatedFiles(let generatedFiles):
-                    node = .generatedFiles(generatedFiles)
+                    switch generatedFiles {
+                    case .multipleConfigs(let configs):
+                        // Sort configs under same package
+                        node = .generatedFiles(.multipleConfigs(configs.sorted { $0.name < $1.name } ))
```
brentleyjones, line 181: This is the wrong place to do this

sebastianv1: >Putting the sort here seemed best since other path tree sorting happens here.

Dropped this note in the commit, would you rather see it happen in `calculatePathTree` or inside `CreateInlineBazelGeneratedFiles`? I imagine you're thinking of performance optimizations

brentleyjones: I would do it up at line 54 above, you can add the sort just before the `map`.

sebastianv1: Updated

test/internal/pbxproj_partials/write_files_and_groups_tests.bzl
```diff
@@ -64,8 +68,16 @@ def _write_files_and_groups_test_impl(ctx):
         struct(
             path = path,
             is_source = False,
+            owner = Label(values[0]),
```
brentleyjones, line 71: Instead of a label, which possibly does some odd resolution, I would create a struct that has the fields that we look at (and if we don't, a simple string).

test/internal/pbxproj_partials/write_files_and_groups_tests.bzl
```diff
@@ -185,8 +197,8 @@ write_files_and_groups_test = unittest.make(
         "file_paths": attr.string_list(mandatory = True),
         "files": attr.string_list(mandatory = True),
         "folders": attr.string_list(mandatory = True),
-        "generated_files": attr.string_list(mandatory = True),
-        "generated_folders": attr.string_list(mandatory = True),
+        "generated_files": attr.string_list_dict(mandatory = True),
```
brentleyjones, line 200: Should this instead be `attr.string_dict` so we can express the 1:1 nature? We can't have multiple owners for a file.

sebastianv1: The ultimately gets transformed into a `struct` to represent a `File` and my intent was to make this extensible in the future to enable mocking other fields (aside from just `owner`) if the implementation expands to use other fields on `File`. wdyt?

test/internal/pbxproj_partials/write_files_and_groups_tests.bzl
sebastianv1: Ah good to know! fixed

test/internal/pbxproj_partials/write_files_and_groups_tests.bzl
brentleyjones: I don't think we need to future proof it.

tools/generators/files_and_groups/test/ElementCreator/CreateInlineBazelGeneratedConfigGroupElementTests.swift
```diff
@@ -0,0 +1,58 @@
...
+            bazelPath: bazelPath,
+            childIdentifiers: [],
+            createIdentifier: createIdentifier.mock
+        )
+
+        XCTAssertNoDifference(
+            createIdentifier.tracker.called,
+            expectedCreateIdentifierCalled
+        )
+        XCTAssertEqual(result.object.identifier, stubbedIdentifier)
```
brentleyjones, line 39: We should also check the contents as well, right?

tools/generators/files_and_groups/test/ElementCreator/CreateInlineBazelGeneratedFilesElementTests.swift
```diff
@@ -0,0 +1,21 @@
...
+        
+        let result = ElementCreator.CreateInlineBazelGeneratedFilesElement.defaultCallable(
+            path: "bazel-out/ios-sim-123",
+            childIdentifiers: [],
+            createIdentifier: .stub(identifier: "id-123")
+        )
+        
+        XCTAssertEqual(result.name, "Bazel Generated")
+        XCTAssertEqual(result.sortOrder, .inlineBazelGenerated)
+        XCTAssertEqual(result.object.identifier, "id-123")
```
brentleyjones, line 19: Same: we should check contents right?

brentleyjones:
@erikkerber @sebastianv1 Can you approve of it works for you locally?

sebastianv1 (approved):
Thank you for the help!

erikkerber (approved):

PR #3061, opened 2024-07-29
Include headers in remote_download_regex

mjburghard:


xcodeproj/internal/templates/runner.sh
```diff
@@ -226,7 +226,7 @@ else
   if [[ $cmd == "build" && -n "${generator_output_groups:-}" ]]; then
     if [[ $download_intermediates -eq 1 ]]; then
       pre_config_flags=(
-        "--experimental_remote_download_regex=.*\.indexstore/.*|.*\.(a|c|C|cc|cl|cpp|cu|cxx|c++|m|mm|swift|swiftdoc|swiftmodule|swiftsourceinfo)$"
+        "--experimental_remote_download_regex=.*\.indexstore/.*|.*\.(a|c|C|cc|cl|cpp|cu|cxx|c++|h|m|mm|swift|swiftdoc|swiftmodule|swiftsourceinfo)$"
```
brentleyjones, line 229: I think we want all of these: https://github.com/bazelbuild/bazel/blob/691149afd591078dc616e0143bc92bed96deedb3/src/main/java/com/google/devtools/build/lib/rules/cpp/CppFileTypes.java#L91-L92

mjburghard: @brentleyjones  The flag allows multiple specifications, should we split it for better readability?

brentleyjones: I wouldn't. It's slightly less performant that way.

brentleyjones (approved):

PR #3070, opened 2024-08-09
Make indexbuild macos QoS 'utility' by default

AngelaGuardia:
**What changed?**
- Adds startup flag to set index builds MacOS QoS to `utility` by default

**Why did it change?**
- Developers have reported excessive battery usage related to background indexing

**How is it tested?**
- Added debug statements to the `bazel_build.sh` to see the startup flag being added for index builds
- I was unable to directly confirm (via my macbook) that the java process was actually set to utility - open to suggestions on how to do this.


AngelaGuardia:
Tried confirming that this change actually changes the qos of the bazel process by using this command `sudo powermetrics --show-process-qos --show-process-qos-tiers --show-cpu-qos --samplers tasks -n 1` but I still don't see concrete proof that the flag actually changes the qos

brentleyjones:
With the merged output bases I don't believe we can, or would want to, take this change?

PR #3074, opened 2024-08-19
Use the same output base for Xcode Build and Index Build

brentleyjones:
By using `--preemptible` Bazel flag on Index Builds, normal builds will cause them to cancel instead of block for them. This along with Xcode’s default interrupt means that Index Builds are able to use the same output base with minimal impact to normal builds.

This change has numerous benefits:

- Indexing completes faster since it can reuse any normal builds that have happened, including analysis
- Normal builds complete faster because it can leverage the work done by background Index Builds
- Quickly open won’t show two source paths anymore for generated files, since there is now only one path
- With only one output base, disk usage is drastically cut down
- There is one less Bazel server, so memory usage is drastically cut down


thiagohmcruz:
Amazing! I'll run some tests on this branch today using our repo!

brentleyjones:
This means that any settings that are different between normal and index builds can cause reanalysis when switching between the two. I'm debugging if there are any changes on the ruleset side that are needed to add with this, as interally we are def running into an issue with that.

brentleyjones:
Needs #3075.

PR #3079, opened 2024-08-20
Fix target merging involving source-less library targets

brentleyjones:


thiagohmcruz (approved):
This fixes an issue I hit in our project, thx!

PR #3088, opened 2024-09-10
Expand "Make" variables inside of values in the env dictionary on test rules

### LINKED ISSUE #3056: Bug: "Make" variable expansion in the `env` of test bundles does not work.
### Description

"Make" variable expansion support was added to rules_apple in [this commit](https://github.com/bazelbuild/rules_apple/commit/3c4111a2b2defdba8086a245d60af565273f2738) and shipped with version 3.6.0. This allows for Starlark like this to work without requiring a `--test_env` in a bazelrc.

```starlark
string_flag(
    name = "my_make_var",
    build_setting_default = "",
    make_variable = "MY_MAKE_VAR",
)

objc_library(
    name = "make_var_unit_test_lib",
    srcs = ["make_var_unit_test.m"],
)

ios_unit_test(
    name = "MakeVarUnitTest",
    deps = [":make_var_unit_test_lib"],
    env = {
        "MY_MAKE_VAR": "$(MY_MAKE_VAR)",
    },
    toolchains = [":my_make_var"],
)
```

This works well enough when building with Bazel's CLI. With Xcode, however, the scheme that is generated for this test produces the uninterpolated value for the `env`. Is this possible to support in the xcodeproj generator? 

### Reproduction steps

1. Create a `string_flag` with a `make_variable` value
2. Create a `*_unit_test` that takes the `string_flag` as one of its `toolchains` and expands the value in its `env` attr.
3. Create an `xcodeproj` that takes the unit test as a top-level target.
4. Generate the xcodeproj.

### Expected behavior

Observe that the generated scheme has the literal `$(MY_MAKE_VAR)` (or whatever your `make_variable` was named), and not the interpolated value from bazelrc or even the default value. 

### rules_xcodeproj version

2.5.1

### Xcode version

15F31d

### Bazel version

7.2.1

### rules_apple version

3.6.0

### rules_swift version

2.0.0

### Additional information

I am using incremental generation and automatic scheme generation.

aaronsky:
Fixes #3056

This PR implements a gap introduced when bazelbuild/rules_apple#2476 was introduced, where expandable env settings on test rules would not be expanded into the xcscheme during generation. There are a couple of things that reviewers should be aware of before proceeding with this change:

- This change is not considerate of the rules_apple being used, meaning that it assumes the developer is using at least version [3.6.0](https://github.com/bazelbuild/rules_apple/releases/tag/3.6.0) (when the breaking changes in #2476 were introduced). This is not good and should be considered a blocker, but I don't know how to address this with the tools offered by the ruleset or Bazel.
- This change as-implemented requires regeneration of the xcschemes if the configured value of the "Make" variable ever changes.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -10,13 +10,13 @@ bazel_dep(name = "bazel_features", version = "1.3.0")
 bazel_dep(name = "bazel_skylib", version = "1.3.0")
 bazel_dep(
     name = "rules_swift",
-    version = "1.6.0",
+    version = "1.18.0",
     max_compatibility_level = 2,
     repo_name = "build_bazel_rules_swift",
 )
 bazel_dep(
     name = "rules_apple",
-    version = "2.0.0",
+    version = "3.6.0",
     repo_name = "build_bazel_rules_apple",
 )
 bazel_dep(name = "rules_python", version = "0.27.1")
--- a/xcodeproj/internal/incremental_xcodeprojinfos.bzl
+++ b/xcodeproj/internal/incremental_xcodeprojinfos.bzl
@@ -258,6 +258,7 @@ def _target_info_fields(
 
 def _make_skipped_target_xcodeprojinfo(
         *,
+        ctx,
         automatic_target_info,
         rule_attr,
         skip_type,
@@ -269,6 +270,7 @@ def _make_skipped_target_xcodeprojinfo(
     forwards them on, not collecting any information for the current target.
 
     Args:
+        ctx: The aspect context.
         automatic_target_info: The `XcodeProjAutomaticTargetProcessingInfo` for
             `the target.
         rule_attr: `ctx.rule.attr`.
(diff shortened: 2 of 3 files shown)


brentleyjones:
> This change is not considerate of the rules_apple being used, meaning that it assumes the developer is using at least version [3.6.0](https://github.com/bazelbuild/rules_apple/releases/tag/3.6.0) (when the breaking changes in [Fix label encoding in incremental generators #2476](https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/2476) were introduced). This is not good and should be considered a blocker, but I don't know how to address this with the tools offered by the ruleset or Bazel.

I'm fine with bumping our minimum required version of **rules_apple** to 3.6.0. I agree there is no good way to detect this otherwise. Though, we could add a flag to go back to the old behavior? I would rather just bump our dep though.

> This change as-implemented requires regeneration of the xcschemes if the configured value of the "Make" variable ever changes.

I don't think there is a better way?

aaronsky:
I've increasingly come to regret the way I introduced support for this feature in rules_apple. I think the most flexible behavior from a UX standpoint is a way to map Bazel settings to project-level xcconfigs, so they're available to the scheme without having to regenerate them, but then you'd need to regenerate the xcconfig instead.

I can put up a new commit that bumps the minimum rules_apple to 3.6.0.

congt:
Any progress on this? We encounter exactly the same problem.

brentleyjones (approved):

aaronsky:
@brentleyjones anything else we need to do to get this landed?

luispadron (approved):

luispadron:
@congt would y'all be able to test this change in Square and see if it fixes the issue

congt:
> @congt would y'all be able to test this change in Square and see if it fixes the issue

Just verified it works for us

brentleyjones:
Thanks everyone!

PR #3112, opened 2024-12-10
Fix to Bazel 8

lucasromanomr:
I made some changes, but I still believe that some adjustments need to be made to make it work or that it might break some tests. Initially, it was just for a start.

One of the things I had to do was export the internal provider from rules_apple. I might have to adjust that in the rules.

need to adjust the visibility of [apple/internal/providers/apple_dynamic_framework_info.bzl](https://github.com/bazelbuild/rules_apple/blob/master/apple/internal/providers/apple_dynamic_framework_info.bzl)

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -7,16 +7,16 @@ module(
 )
 
 bazel_dep(name = "bazel_features", version = "1.3.0")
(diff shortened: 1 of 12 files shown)


brentleyjones:
#3117

PR #3115, opened 2024-12-12
Create param files with absolute path 

yongjincho92:
This change would make the index-compile for single source file (.m and .mm) to not depend on params file generated by index-build of the target containing the source file.

### DIFF
--- a/xcodeproj/internal/pbxproj_partials.bzl
+++ b/xcodeproj/internal/pbxproj_partials.bzl
@@ -1,6 +1,7 @@
 """Actions for creating `PBXProj` partials."""
 
 load("//xcodeproj/internal/files:files.bzl", "join_paths_ignoring_empty")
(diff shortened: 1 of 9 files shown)


xcodeproj/internal/pbxproj_partials.bzl
```diff
@@ -1092,6 +1094,9 @@ def _write_target_build_settings(
         swift_args,
         swift_debug_settings_to_merge = EMPTY_DEPSET,
         team_id = None,
+        xcode_config,
```
yongjincho92, line 1097: @brentleyjones Got a question on handling this param in the unit test for this function. Right now, `apple_support.run()` would try to validate `xcode_config` and fails in unit test. 

I would like to mock `XcodeVersionConfig`, but can we even mock it?

PR #3125, opened 2025-01-13
Fix non-experimental remote_download_regex flag

luispadron:
This flag in Bazel 7.0 is no longer experimental so you get the following warning:

```
WARNING: Option 'experimental_remote_download_regex' is deprecated: Use --remote_download_regex instead
```

We can update to the non-experimental flag as we now only support Bazel 7+.


brentleyjones (approved):

jszumski (approved):

thiagohmcruz (approved):

luispadron:
@brentleyjones same `HEAD` issue here but should otherwise be ready

PR #3135, opened 2025-03-06
Generate non recursive LIBRARY_SEARCH_PATHS paths

karim-alweheshy:
For transitive deps with .a output


karim-alweheshy:
Closing this proof of concept as superseded.

Its static-library Preview-linking work was absorbed into the broader later experiments, while #3325 is now the current Xcode 26 XOJIT implementation with substantially stronger tests and current generator paths. Since this branch also conflicts with the path renames/removals merged through #3192, rebasing it would recreate obsolete machinery rather than preserve a clean independent change.

Any remaining gap should be split into a focused follow-up on top of #3325/current `main`.

PR #3139, opened 2025-03-18
Add custom toolchain rule

karim-alweheshy:
This PR implements a `custom_toolchain` rule to copy, stub and link a toolchain that points directly to the default toolchain. This eventually will be used instead of `SWIFT_USE_INTEGRATED_DRIVER` that does not work with SwiftUIPreviews

kudos to [Nick](https://github.com/longsview) for setting it up

### DIFF
--- a/test/internal/custom_toolchain/BUILD
+++ b/test/internal/custom_toolchain/BUILD
@@ -1,0 +1,27 @@
+# buildifier: disable=bzl-visibility
+load("//xcodeproj/internal:custom_toolchain.bzl", "custom_toolchain")
+
+# Example swiftc override for testing
+filegroup(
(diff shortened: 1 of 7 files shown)


karim-alweheshy:
Closing this as superseded/obsolete rather than publishing the local rebase.

All four commits from this branch are patch-identical to commits already carried by the broader #3140 experiment, and the active Preview implementation has since moved to the Xcode 26 XOJIT approach in #3325. The focused custom-toolchain test path is also no longer valid on current `main` because the old native `sh_test` setup was removed.

Any custom-toolchain behavior still worth preserving should be extracted as a new, independently tested PR against current `main`.

PR #3141, opened 2025-03-24
Fix file permissions on macOS 15.4

adincebic:
There have been changes to `rsync` on macOS 15.4 which require us to have write permissions here.

### DIFF
--- a/xcodeproj/internal/bazel_integration_files/copy_outputs.sh
+++ b/xcodeproj/internal/bazel_integration_files/copy_outputs.sh
@@ -31,6 +31,7 @@ if [[ "$ACTION" != indexbuild ]]; then
       ln -sfh "$PWD/$BAZEL_OUTPUTS_PRODUCT_BASENAME" "$TARGET_BUILD_DIR/$PRODUCT_NAME"
     else
       # Product is a bundle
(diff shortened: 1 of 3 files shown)


xcodeproj/internal/bazel_integration_files/copy_outputs.sh
```diff
@@ -31,6 +31,7 @@ if [[ "$ACTION" != indexbuild ]]; then
       ln -sfh "$PWD/$BAZEL_OUTPUTS_PRODUCT_BASENAME" "$TARGET_BUILD_DIR/$PRODUCT_NAME"
     else
       # Product is a bundle
+      chmod -R +w "$BAZEL_OUTPUTS_PRODUCT_BASENAME" "$TARGET_BUILD_DIR" # Starting on MacOS 15.4, rsync will fail if it does not have write access on the src or dst directory
```
brentleyjones, line 34: Does this adjust permissions in `bazel-out`? If so, we can't do that.

adincebic:
> This seems odd. rsync shouldn't have to have write access to where its copying from 🧐

Makes sens. Didn't give it much thought, we had it as a patch for quite some time, I'll take a look at it again.

brentleyjones:
Adjusted in https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/3157.

PR #3147, opened 2025-03-25
Don’t require `xcode-select` or `DEVELOPER_DIR` to run

brentleyjones:
Needed to support building on Linux.

### DIFF
--- a/xcodeproj/internal/templates/runner.sh
+++ b/xcodeproj/internal/templates/runner.sh
@@ -183,7 +183,7 @@ common:rules_xcodeproj --repo_env=XCODE_VERSION=%xcode_version%
 EOF
 
 if command -v /usr/bin/xcode-select >/dev/null 2>&1; then
-  developer_dir=$(xcode-select -p)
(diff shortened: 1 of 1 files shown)


xcodeproj/internal/templates/runner.sh
```diff
@@ -169,24 +169,33 @@ if [[ -s "$extra_flags_bazelrc" ]]; then
...
-
 # Work around https://github.com/bazelbuild/bazel/issues/8902
 # \`USE_CLANG_CL\` is only used on Windows, we set it here to cause Bazel to
 # re-evaluate the cc_toolchain for a different Xcode version
 common:rules_xcodeproj --repo_env=USE_CLANG_CL=%xcode_version%
 common:rules_xcodeproj --repo_env=XCODE_VERSION=%xcode_version%
 EOF
 
+if command -v /usr/bin/xcode-select >/dev/null 2>&1; then
+  developer_dir=$(xcode-select -p)
```
mattrobmattrob, line 186: Want to use same command path as was checked above here?

mattrobmattrob (approved):

PR #3150, opened 2025-03-25
Create xcodeproj_extra_files aspect hint

jflan-dd:
Create an aspect hint to allow dependencies to propagate up extra files that should be include in the Xcode navigator

### DIFF
--- a/docs/BUILD
+++ b/docs/BUILD
@@ -6,6 +6,7 @@ load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
 # Generating
 
 _DOC_COMPONENTS = [
+    "aspect_hints",
     "project_options",
     "providers",
     "top_level_target",
@@ -35,6 +36,7 @@ genrule(
         ":xcschemes",
         ":xcode_build_settings",
         ":providers",
+        ":aspect_hints",
     ],
     outs = ["bazel.generated.md"],
     cmd = """\
--- a/docs/bazel.md
+++ b/docs/bazel.md
@@ -50,6 +50,8 @@ load("@rules_xcodeproj//xcodeproj:xcodeproj.bzl", "xcodeproj")
 - [Providers](#providers)
   - [`XcodeProjAutomaticTargetProcessingInfo`](#XcodeProjAutomaticTargetProcessingInfo)
   - [`XcodeProjInfo`](#XcodeProjInfo)
+- [Aspect Hints](#aspect-hints)
+  - [`xcodeproj_extra_files`](#xcodeproj_extra_files)
 
 # Core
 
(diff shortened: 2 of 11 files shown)


xcodeproj/defs.bzl
```diff
@@ -34,6 +35,7 @@ project_options = _project_options
 top_level_target = _top_level_target
 top_level_targets = _top_level_targets
 xcodeproj = _xcodeproj
+xcodeproj_extra_files = _xcodeproj_extra_files
```
brentleyjones, line 38: We are moving away from a single file for loads, so let’s instead just have the aspect hint live one level up.

docs/BUILD
```diff
@@ -36,6 +37,7 @@ genrule(
         ":xcode_schemes",
         ":xcode_build_settings",
         ":providers",
+        ":aspect_hints",
```
jflan-dd, line 40: This is giving me an error. I tried to mimic the setup for the other sets of documentation, but I _think_ the issues is that this is the first one where the implementation is at the root of `xcodeproj/` rather than in `xcodeproj/internal/`

```
ERROR: /Users/johnflanagan/Projects/src/rules_xcodeproj/docs/BUILD:19:12: in deps attribute of starlark_doc_extract rule //docs:aspect_hints.extract: missing bzl_library targets for Starlark module(s) //xcodeproj:xcodeproj_extra_files.bzl. Since this rule was created by the macro 'stardoc', the error might have been caused by the macro implementation
ERROR: /Users/johnflanagan/Projects/src/rules_xcodeproj/docs/BUILD:19:12: Analysis of target '//docs:aspect_hints.extract' failed
ERROR: Analysis of target '//docs:update' failed; build aborted: Analysis failed
```

xcodeproj/defs.bzl
jflan-dd: Can you help me understand how I should structure the dependencies if I move the aspect hit up a level?

I need `xcodeproj/internal/files/incremental_input_files.bzl` to depend on the new aspect hit which I currently have defined in `xcodeproj/xcodeproj_extra_files.bzl` which I think is were you're suggesting I put it in this comment.

This results in `xcodeproj_extra_files.bzl` being included in the sources glob for the`//xcodeproj:xcodeproj` `bzl_library`, but since the `xcodeproj` library depends on `//xcodeproj/internal` I don't know how to resolve the dependency for `xcodeproj/internal/files/incremental_input_files.bzl`.

jflan-dd: I moved things back to the way I could get the dependency tree working. Let me know if there's a better way to structure this.

docs/bazel.md
```diff
@@ -814,3 +816,49 @@ Provides information needed to generate an Xcode project.
...
+<pre>
+xcodeproj_extra_files(<a href="#xcodeproj_extra_files-name">name</a>, <a href="#xcodeproj_extra_files-files">files</a>)
+</pre>
+
+This rule is used to surface extra files that should be included in the Xcode
+project navigator, but otherwise aren't inputs to a target. The provider
+created by this rule should be attached to the related target via an aspect
+hint.
+
+This is only used when xcodeproj.generation_mode = "incremental" is set.
```
jflan-dd, line 836: Fixing...

PR #3156, opened 2025-03-31
Add `compile_only` aspect

luispadron:
Adds an aspect that can be used to skip actions which are not compilation related. This is useful for warming projects in CI to maintain a Bazel remote cache for users to use. It is also useful when compilation actions are the only thing that need to be validated as it allows skipping potentially longer actions like bundling, signing, etc.

Example usage: `bazel run //:LocalDevCacheWarming_xcodeproj -- 'build --aspects=//:compile_only_aspect.bzl%compile_only_aspect --output_groups=compiles $_GENERATOR_LABEL_'`

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -1,3 +1,5 @@
+- [Bazel aspects](#bazel-aspects)
+  - [`compile_only_aspect`](#compile_only_aspect)
 - [Bazel configs](#bazel-configs)
   - [`rules_xcodeproj`](#rules_xcodeproj)
   - [`rules_xcodeproj_generator`](#rules_xcodeproj_generator)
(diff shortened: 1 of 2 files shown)


xcodeproj/defs.bzl
```diff
@@ -39,3 +40,6 @@ xcode_provisioning_profile = _xcode_provisioning_profile
 # Re-exporting APIs
 xcode_schemes = _xcode_schemes
 xcschemes = _xcschemes
+
+# Re-exporting aspects
```
brentleyjones, line 44: Can we just have it be a top-level file? I don’t want to add new symbols to this bzl file.

brentleyjones (commented):
Can you adjust the example in `Usage.md` to use this aspect?

PR #3158, opened 2025-04-01
Fix legacy fixture generation

brentleyjones:
Regressed in 017297ba0b16b9c1635c06188b69e314e63fe5e9.


xcodeproj/internal/templates/runner.sh
```diff
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 
+set -x
+
```
brentleyjones: Will revert

PR #3163, opened 2025-04-03
Update apple_support to 1.21.0

luispadron:
Should fix Bazel HEAD CI


buildbuddy.yaml
```diff
@@ -23,7 +23,7 @@ x_templates:
         USE_BAZEL_VERSION: 8.x
     - &bazel_head
       env:
-        USE_BAZEL_VERSION: last_green
+        USE_BAZEL_VERSION: 4c2d91e762ab6e492853b021408129dd93fb5904 # TODO: Switch back to last_green once get_cpu has been removed
```
luispadron, line 26: Can be removed once https://github.com/bazelbuild/rules_apple/pull/2670 lands

luispadron: hmm this breaks other things 🙃 might just need to wait for that rules_apple change to land so we can pull apple_support + rules_apple upgrades

PR #3165, opened 2025-04-06
Allow empty string value for xcschemes.env_value

ghugues:
Xcode allows defining environment variables without a value. At runtime, sometimes only the presence of the environment variable is checked, not its value (e.g. `CG_NUMERICS_SHOW_BACKTRACE`, `CGBITMAP_CONTEXT_LOG_ERRORS`, etc.).

### DIFF
--- a/xcodeproj/internal/xcschemes/xcschemes.bzl
+++ b/xcodeproj/internal/xcschemes/xcschemes.bzl
@@ -1142,7 +1142,7 @@ def _env_value(value, *, enabled = True):
             checked in the scheme. An unchecked checkbox means Xcode won't
             include that environment variable when running a target.
     """
-    if not value:
+    if not value and value != "":
         fail("""
 `value` must be provided to `xcschemes.env_value`.
 """)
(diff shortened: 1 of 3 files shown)


adincebic:
@ghugues I think you need to sign off the commit.

brentleyjones (approved):

xcodeproj/internal/xcschemes/xcschemes.bzl
```diff
@@ -1142,7 +1142,7 @@ def _env_value(value, *, enabled = True):
             checked in the scheme. An unchecked checkbox means Xcode won't
             include that environment variable when running a target.
     """
-    if not value:
```
brentleyjones, line 1145: It would be nice to adjust the tests for this.

ghugues: This was not covered by unit tests so I add added a new suite with a couple tests. Let me know if this is what you had in mind.

PR #3172, opened 2025-04-23
Skip targets without "compiles" output group in compile_only aspect

adincebic:
We are running some additional targets which are not Apple / Swift specific but still required and the build was failing since there was no "compiles" output group.

I could add those rules to the list in the aspect definition. However I fear that it would set the bad example and the list would probably grow a lot over time.

### DIFF
--- a/xcodeproj/compile_only_aspect.bzl
+++ b/xcodeproj/compile_only_aspect.bzl
@@ -66,17 +66,13 @@ def _compile_only_aspect_impl(target, ctx):
     else:
         return []
 
-    for dep in deps:
-        if OutputGroupInfo in dep and not hasattr(dep[OutputGroupInfo], "compiles"):
-            fail(target, dep)
-
     return [
         OutputGroupInfo(
             compiles = depset(
                 transitive = outs + [
                     dep[OutputGroupInfo].compiles
                     for dep in deps
-                    if OutputGroupInfo in dep
+                    if OutputGroupInfo in dep and hasattr(dep[OutputGroupInfo], "compiles")
                 ],
             ),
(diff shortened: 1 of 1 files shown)


mattrobmattrob (commented):
The `*_KINDS` hashes are made public so you can append to them in your ~repo~ workspace, right? Why is `_compile_only_aspect_impl` made public?

adincebic:
My mistake, I will make zhem private again.

xcodeproj/compile_only_aspect.bzl
```diff
@@ -65,18 +63,16 @@ def _compile_only_aspect_impl(target, ctx):
         )
     else:
         return []
-
     for dep in deps:
         if OutputGroupInfo in dep and not hasattr(dep[OutputGroupInfo], "compiles"):
-            fail(target, dep)
-
+            continue
```
mattrobmattrob, line 68: If this isn't to signal the failure anymore, it can just be removed.
```suggestion
```

mattrobmattrob (approved):
Please just remove the whitespace deletions as well.

xcodeproj/compile_only_aspect.bzl
adincebic: Ah yes, makes sense.

mattrobmattrob (approved):
LGTM. PTAL @luispadron for the merging approval. Can you fix the commit message style check too, @adincebic?

mattrobmattrob (approved):

luispadron (approved):

brentleyjones (approved):

PR #3173, opened 2025-04-23
Fix argument length error for WriteSwiftDebugSettings

jszumski:
Fix argument length errors by switching WriteSwiftDebugSettings to use a param file when needed.

Previously when generating a very large project with 7000+ targets:
```
ERROR: /private/var/tmp/_bazel_jszumski/399f2da8a1f92cab020446a488e7bec6/rules_xcodeproj.noindex/build_output_base/external/rules_xcodeproj_generated/generator/tools/rules/xcodeproj/bazel-project/BUILD:12:10: Generating bazel-project.xcodeproj Debug-swift_debug_settings.py failed: (Exit -1): universal_swift_debug_settings failed: error executing WriteSwiftDebugSettings command (from target @@rules_xcodeproj_generated//generator/tools/rules/xcodeproj/bazel-project:bazel-project) 
  (cd /private/var/tmp/_bazel_jszumski/399f2da8a1f92cab020446a488e7bec6/rules_xcodeproj.noindex/build_output_base/execroot/register && \
  exec env - \
  bazel-out/darwin_x86_64-opt-exec-macos-x86_64-min13.0-applebin_macos-ST-5b758c1c0c80/bin/external/rules_xcodeproj/tools/generators/swift_debug_settings/universal_swift_debug_settings 1 {...lots of pairs here...}
# Configuration: 794c86670c8dbf3648f6dfd018247dc9402b8c0a923a1790920ea9bf34028e6a
# Execution platform: @@internal_platforms_do_not_use//host:host
Action failed to execute: java.io.IOException: Cannot run program "/var/tmp/_bazel_jszumski/install/fd0e803cd377e5d6c999b9f309de8848/process-wrapper" (in directory "/private/var/tmp/_bazel_jszumski/399f2da8a1f92cab020446a488e7bec6/rules_xcodeproj.noindex/build_output_base/execroot/register"): error=7, Argument list too long
```

### DIFF
--- a/tools/generators/swift_debug_settings/src/SwiftDebugSettings.swift
+++ b/tools/generators/swift_debug_settings/src/SwiftDebugSettings.swift
@@ -1,4 +1,5 @@
 import Darwin
+import Foundation
 import ToolCommon
 
 @main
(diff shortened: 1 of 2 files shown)


jszumski:
> Does this affect tests/fixtures in any way? If not should we add one?

Nope, it's conditional based on the specific args:
> Bazel will decide whether the arguments need to be spilled based on your system and arg length.
I'm happy to add a test but I couldn't find an example of any other param file call site that had one.

luispadron (approved):

tools/generators/swift_debug_settings/src/SwiftDebugSettings.swift
```diff
@@ -12,16 +13,32 @@ struct SwiftDebugSettings {
...
+            
+            // Check for a params file
+            var arguments: ArraySlice<String>
+            if let arg = rawArguments.first, rawArguments.count == 1,
+               arg.starts(with: "@")
+            {
+                arguments = try await parseParamsFile(String(arg.dropFirst()))
+            } else {
+                arguments = ArraySlice(rawArguments)
+            }
```
brentleyjones, line 26: We have `parseAsRootSupportingParamsFile` for this, why can't we use that?

jszumski: It seemed more invasive to transition this to `ArgumentParser` and then declare a generic property to take in arbitrary key & file pairs to pass to the `Generator`. Similar logic applies to `TargetBuildSettings` too.

I'm happy to take that route though if you think it'll be cleaner.

brentleyjones (approved):

PR #3175, opened 2025-05-06
Use `--perms` with `rsync`

brentleyjones:
This should fix writable permissions issues on all versions of macOS that we support.


brentleyjones:
I'm not able to test on macOS 15.3 and below.

mattrobmattrob (approved):

luispadron (approved):

PR #3182, opened 2025-05-17
wip: Support StoreKit Configuration setting in xcscheme Run action

aaronsky:
Fixes #3130

PR #3185, opened 2025-05-20
Add StoreKit Configuration support to Xcode Schemes.

yanks:
This change satisfies requirements of #3182, allowing the user to supply a relative path to a storekit configuration file based on the workspace root:

```
xcodeproj(
<snip>
    xcschemes = [
        xcschemes.scheme(
            name = "MyApp",
            run = xcschemes.run(
                launch_target = ":MyApp",
                storekit_configuration = "path/to/StoreKitConfig.storekit",
            ),
        ),
</snip>
)
```

The result is this written to the relevant xcscheme:
```
      <StoreKitConfigurationFileReference
         identifier = "../../path/to/StoreKitConfig.storekit">
      </StoreKitConfigurationFileReference>
```

A few notes:
* I'm pushing this out a bit early, because I couldn't figure out a great way to resolve a label to a given filegroup to a reasonable path relative to the project root. `xcschemes_json` encodes as a string into the rule, and it seems a bit unwieldy to have to resolve this issue for something that will never be modeled within bazel anyway (as this is an Xcode-only feature). I would love input on this! I think the tradeoff of just requiring the file to be relative to the workspace root to be sufficient, but happy to discuss.
* There are a few more tests I'd like to write, and some guardrails on the file itself (project generation should fail when it can't resolve the path to storekit file)


test/internal/xcode_schemes/model_tests.bzl
```diff
@@ -256,6 +256,7 @@ def _launch_action_test(ctx):
         diagnostics = None,
         env = env,
         working_directory = None,
+        storekit_configuration = None,
```
brentleyjones, line 259: Please make this alphabetical.

test/internal/xcschemes/info_constructors_tests.bzl
```diff
@@ -590,6 +590,7 @@ def info_constructors_test_suite(name):
             env_include_defaults = "1",
             launch_target = xcscheme_infos_testable.make_launch_target(),
             xcode_configuration = "",
+            storekit_configuration = "",
```
brentleyjones, line 593: Same, and everywhere.

tools/generators/xcschemes/src/Generator/URL+Extensions.swift
```diff
@@ -0,0 +1,26 @@
...
+        
+        // Add "../" for each level to go up
+        result.append(contentsOf: Array(repeating: "..", count: sourceComponents.count - commonPrefixCount))
+        
+        // Add remaining destination components
+        result.append(contentsOf: destComponents[commonPrefixCount...])
+        
+        return result.joined(separator: "/")
+    }
+}
```
brentleyjones, line 26:
```suggestion
}

```

tools/generators/xcschemes/src/Generator/URL+Extensions.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+extension URL {
+    func relativize(from source: URL) -> String {
```
brentleyjones, line 3: We haven’t needed this until now. I question if we still need it. Is there a way to make sure it comes in correctly from Starlark instead?

brentleyjones (approved):
Approve only to get CI to run.

tools/generators/xcschemes/src/Generator/URL+Extensions.swift
yanks: I struggled with this a lot. I suppose I could relativize this within starlark and provide a relative path there, but there is no great way to get a true path to the file via the rule itself. The schemes are encoded via json and supplied here: https://github.com/MobileNativeFoundation/rules_xcodeproj/blob/main/xcodeproj/internal/xcodeproj_incremental_rule.bzl#L835

And other things within the scheme match up based on identifier encodings, but this requires a path that is relative to the resultant scheme directory itself in the generated xcodeproj from what I can tell. I am happy to do it differently, but I think the outcome ends up being something considerably more complex. I'm not particularly experienced in working with rules_xcodeproj, so after banging my head on this for a little bit I threw up my hands and went with this. Open to suggestions!

chrisballinger:
Oh perfect, thank you for diving into this!

tools/generators/xcschemes/src/Generator/URL+Extensions.swift
brentleyjones: Mainly we want to be able to "cache" things in Bazel. Doing this work in the generator means it's re-done every time the generator runs. I'll leave it up to the other @MobileNativeFoundation/rules_xcodeproj-maintainers on what they think about this.

PR #3212, opened 2025-07-15
Support new convention for WMO module deps artifacts naming in Xcode 26 beta 3

aaronsky:
Xcode 26 beta 3 [seems to have changed a convention](https://github.com/swiftlang/swift-build/blob/release/6.2/Sources/SWBCore/SpecImplementations/Tools/SwiftCompiler.swift#L2249) around the naming of various dependencies files when belonging to a WMO module, from using a `-master` suffix to `-primary`. 

As an aside, it's not clear to me why we can't use the fallback case in this method, since it appears as though it would help. Maybe @brentleyjones can weigh in.


mattrobmattrob (approved):

aaronsky:
Can we merge this? The check failures are unrelated.

PR #3213, opened 2025-07-21
Fix Metal toolchains in Xcode 26 betas

adincebic:
Fixes:

```
Error: TOOLCHAINS was set to 'com.apple.dt.toolchain.Metal.32023' but the default toolchain was found, that likely means a matching toolchain isn't installed
```

Apparently `$toolchain` either was never set for Metal toolchains in earlier Xcode versions or Xcode didn't complain when it was set in `action_env`. I tested on Xcode 16 and it was never there while it definitely shows up in Xcode 26 Beta 3.

I think version check may not be needed here at all. If yes, please do let me know what would be suggested way of performing the check. I tried calling `xcodebuild -version` but that seem to randomly fail for some people so I removed it.

### DIFF
--- a/xcodeproj/internal/templates/bazel_build.sh
+++ b/xcodeproj/internal/templates/bazel_build.sh
@@ -97,9 +97,20 @@ readonly base_pre_config_flags=(
 # Custom Swift toolchains
 
 if [[ -n "${TOOLCHAINS-}" ]]; then
-  toolchain="${TOOLCHAINS%% *}"
-  if [[ "$toolchain" == "com.apple.dt.toolchain.XcodeDefault" ]]; then
-    unset toolchain
+  # We remove all Metal toolchains from the list first
+  toolchains_array=($TOOLCHAINS)
+  filtered_toolchains=()
+  for tc in "${toolchains_array[@]}"; do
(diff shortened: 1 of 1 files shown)


xcodeproj/internal/templates/bazel_build.sh
```diff
@@ -103,6 +103,13 @@ if [[ -n "${TOOLCHAINS-}" ]]; then
   fi
 fi
 
+# We unset Metal toolchains since Xcode 26+ complains when that's passed as action_env
+# Older Xcode versions don't seem to populate this variable with Metal toolchain
+# This is why there is no version check
+if [[ "${toolchain:-}" == *"com.apple.dt.toolchain.Metal"* ]]; then
+  unset toolchain
```
brentleyjones, line 110: Would it be better to remove the Metal toolchain instead? Unsetting it will prevent custom toolchains from working if there are multiple toolchains set, won't it?

brentleyjones (approved):

brentleyjones:
Please squash and sign off on the git commit to pass DCO. Then we can merge this.

adincebic:
> Please squash and sign off on the git commit to pass DCO. Then we can merge this.

@brentleyjones done.

PR #3219, opened 2025-08-11
Add --remote_download_outputs=toplevel to xcodeproj.bazelrc

erneestoc:
As titled. There's an issue when remote download minimal is generating empty app bundles when using Xcode.

```
error: Failed to read Info.plist of app /Users/ernesto/Library/Developer/Xcode/DerivedData/InstacartBazel-eqgbkfaxbhrobdfmvkzzkbbafnzs/Build/Products/Debug-iphonesimulator/bazel-out/ios_sim_arm64-dbg-ios-sim_arm64-min16.0-applebin_ios-ST-1de1fcec1e3b/bin/Instacart.app: The operation couldn't be completed. (SWBUtil.PropertyListConversionError error 1.) (in target 'InstacartDev' from project 'InstacartBazel')
```

### DIFF
--- a/xcodeproj/internal/templates/xcodeproj.bazelrc
+++ b/xcodeproj/internal/templates/xcodeproj.bazelrc
@@ -65,6 +65,9 @@ common:rules_xcodeproj --show_result=0
 common:rules_xcodeproj --noworker_sandboxing
 common:rules_xcodeproj --spawn_strategy=remote,worker,local
 
+# Output group outputs aren't downloaded with minimal, we need to use toplevel with Xcode
+build:rules_xcodeproj --remote_download=toplevel
+
 ### `--config=rules_xcodeproj_generator`
 #
 # Used when generating a project.


xcodeproj/internal/templates/xcodeproj.bazelrc
```diff
@@ -65,6 +65,9 @@ common:rules_xcodeproj --show_result=0
 common:rules_xcodeproj --noworker_sandboxing
 common:rules_xcodeproj --spawn_strategy=remote,worker,local
 
+# Output group outputs aren't downloaded with minimal, we need to use toplevel with Xcode
+build:rules_xcodeproj --remote_download=toplevel
```
aaronsky, line 69:
```suggestion
build:rules_xcodeproj --remote_download_outputs=toplevel
```
assuming we want this and not `--remote_download_toplevel`

aaronsky (changes requested):

brentleyjones:
> Don't think we can set this globally yet as not every remote cache supports BwoB.

The default for Bazel is this, so I think we are safe re-enforcing the default? Then anyone that needs to override Bazel's default will need to do it for **rules_xcodeproj** as well? At this point remote caches really should support this, and it's more likely that someone wants this default than accidentally breaking their Xcode builds?

luispadron (approved):
Overriding (kinda forgot this was possible) and matching the defaults in Bazel make sense

PR #3220, opened 2025-08-26
Use DEVELOPER_DIR as an --xcode_version

ma-oli:
In the exceptional case where the user has multiple versions of Xcode on his system with the same Xcode version, we want to make sure that Xcode is invoking bazel by referring to itself.

As bazel supports passing DEVELOPER_DIR to --xcode_version, we're using that variable when invoked through Xcode instead.

### DIFF
--- a/xcodeproj/internal/templates/bazel_build.sh
+++ b/xcodeproj/internal/templates/bazel_build.sh
@@ -79,7 +79,7 @@ readonly bazel_cmd
 
 readonly base_pre_config_flags=(
   # Be explicit about our desired Xcode version
-  "--xcode_version=$XCODE_PRODUCT_BUILD_VERSION"
+  "--xcode_version=$DEVELOPER_DIR"
 
   # Set `DEVELOPER_DIR` in case a bazel wrapper filters it
   "--repo_env=DEVELOPER_DIR=$DEVELOPER_DIR"


xcodeproj/internal/templates/bazel_build.sh
```diff
@@ -79,7 +79,7 @@ readonly bazel_cmd
 
 readonly base_pre_config_flags=(
   # Be explicit about our desired Xcode version
-  "--xcode_version=$XCODE_PRODUCT_BUILD_VERSION"
```
brentleyjones, line 82: Is using `$DEVELOPER_DIR` a new capability for Bazel? I don't recall this working before.

Since `DEVELOPER_DIR` could just be `/Applications/Xcode.app/Contents/Developer`, I believe this has the issue of Bazel caching which version this resolves to, so if the Xcode installed at that spot changes, Bazel still thinks it's compiling with a different version.

Also, `--xcode_version=` impacts analysis, and this will result in a cache miss from the version used in `xcodeproj/internal/templates/runner.sh`.

ma-oli: But yes; I agree with your comment about bazel not noticing Xcode changing. That could be a problem for the general case actually 🤔 let me think about it. I wish there was a way to override it somehow

PR #3222, opened 2025-09-05
Improve `compile_only_aspect`

brentleyjones:
* Collect `CppCompile`
* Add a new `resources` output group, which will cause the resources that will be bundled to be processed (e.g. asset catalog compiles, `Info.plist` processing, entitlement processing, etc.)


mattrobmattrob (approved):

luispadron (approved):
👑

PR #3242, opened 2025-11-04
Add support for viewing and edit xcmappingmodel files

lucasromanomr:
Updates the handling of `.xcmappingmodel` files in the file and group element creation logic, ensuring they are treated correctly as file references rather than groups. It also adds a dedicated unit test to verify this behavior.

Before implementation, only the .xml file was visible, inside a regular folder. In addition to only being able to view the XML content and not the Xcode preview,
image (1)

Now there is interpretation via Xcode. 
image

In addition to displaying correctly in the file structure, when opened, it shows the Xcode preview and not just the XML as before.


adincebic (approved):
This is great, thanks @lucasromanomr

PR #3243, opened 2025-11-06
Allow separate/merged index build output base to be configured

aaronsky:
We've run into stability issues that we believe can be traced back to the point where the build and indexbuild output bases were merged, particularly in the area of sudden pcm corruption when changing branches that require a module cache busting. I'm investigating whether or not this helps to ease the pain. 

TODO:
- [x] docs
- [x] changelog

### DIFF
--- a/tools/generators/pbxproj_prefix/src/Generator/PBXProjectBuildSettings.swift
+++ b/tools/generators/pbxproj_prefix/src/Generator/PBXProjectBuildSettings.swift
@@ -200,12 +200,12 @@ extension Generator {
                 ),
             ])
         } else {
-        buildSettings.append(
-        .init(
-            key: "PROJECT_DIR",
-            value: projectDir.pbxProjEscaped
-        )
-        )
+            buildSettings.append(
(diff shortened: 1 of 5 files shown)


tools/generators/pbxproj_prefix/src/Generator/PBXProjectBuildSettings.swift
```diff
@@ -165,6 +174,39 @@ extension Generator {
...
+                    value:
+                        #""$(INDEXING_PROJECT_DIR__$(INDEX_ENABLE_BUILD_ARENA))""#
+                ),
+            ])
+        } else {
+        buildSettings.append(
+        .init(
+            key: "PROJECT_DIR",
+            value: projectDir.pbxProjEscaped
+        )
```
brentleyjones:
```suggestion
            .init(
                key: "PROJECT_DIR",
                value: projectDir.pbxProjEscaped
            )
```

brentleyjones (approved):

PR #3279, opened 2026-02-19
Delete unused files

john-flanagan:
Is it correct that these aren't used?


thiagohmcruz:
From a quick scan this seems to be unused indeed, IIUC this was moved to Swift but I'd confirm with @brentleyjones

brentleyjones (approved):

brentleyjones (approved):

Issue #3134, opened 2025-02-23
Bug: Writing lldbinit configuration to home directory interferes with all non-rules_xcodeproj Xcode projects


tinder-cfuller:
### Description

Writing lldbinit configuration to home directory interferes with all non-rules_xcodeproj Xcode projects.

To avoid rules_xcodeproj interfering with all Xcode projects, lldbinit configuration should be project specific.

This issue is exacerbated by the fact that `bazelisk clean --expunge` removes the directories referenced by the lldbinit config resulting in errors in all non-rules_xcodeproj Xcode projects.

### Reproduction steps

1. Use rules_xcodeproj as normal
2. (optional) Verify that `.lldbinit-Xcode` and `.lldbinit-rules_xcodeproj` have been created in home directory
3. (optional) Verify that `/var/tmp/_bazel_[REDACTED]/189ed9e41c694ca49d4b373f9608df74` directory exists (using actual username and hash – see contents of `.lldbinit-rules_xcodeproj` for exact path)
4. Run `bazelisk clean --expunge`
5. (optional) Verify that `/var/tmp/_bazel_[REDACTED]/189ed9e41c694ca49d4b373f9608df74` directory no longer exists (using actual username and hash – see contents of `.lldbinit-rules_xcodeproj` for exact path)
6. Create a new empty directory for a new Swift package
7. Make the new empty directory the current working directory
8. Create a new Swift package using `swift package init`
9. Open the new Swift package using `xed Package.swift`
10. Run the unit tests (press command-u for example)
11. See the following errors in the log:

```
error: the replacement path doesn't exist: "/var/tmp/_bazel_[REDACTED]/189ed9e41c694ca49d4b373f9608df74/rules_xcodeproj/build_output_base/execroot/__main__/bazel-out"
error: the replacement path doesn't exist: "/var/tmp/_bazel_[REDACTED]/189ed9e41c694ca49d4b373f9608df74/rules_xcodeproj/indexbuild_output_base/execroot/__main__/bazel-out"
error: the replacement path doesn't exist: "/var/tmp/_bazel_[REDACTED]/189ed9e41c694ca49d4b373f9608df74/rules_xcodeproj/build_output_base/external"
error: the replacement path doesn't exist: "/var/tmp/_bazel_[REDACTED]/189ed9e41c694ca49d4b373f9608df74/rules_xcodeproj/indexbuild_output_base/execroot/__main__/external"
[...]

PR #3291, opened 2026-03-17
Fix Xcode navigation for stack trace paths with leading slash

tinder-hemanthprasad:
Fixes #3290

Stack traces in Xcode show workspace-relative paths with a leading slash (e.g., `/Projects/Module/File.swift:123`). These paths are not clickable because the lldbinit source-map configuration only handles paths starting with `./` and not `/`.

This adds a source-map entry for the `/` prefix to `create_lldbinit.sh`, mapping it to the workspace root (`$execution_root`).

## Testing

- Manually verified: Added the source-map entry to `~/.lldbinit-rules_xcodeproj`, restarted Xcode, and confirmed clicking stack trace paths now navigates correctly

## Related

Related to #3134 (other lldbinit configuration concerns)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/xcodeproj/internal/bazel_integration_files/create_lldbinit.sh
+++ b/xcodeproj/internal/bazel_integration_files/create_lldbinit.sh
@@ -29,6 +29,10 @@ echo "settings append target.source-map ./external/ \"$BAZEL_EXTERNAL\""
 # `external` when set from swiftsourcefile
 echo "settings append target.source-map ./external/ \"$build_external\""
 
+# Workspace-relative paths with leading slash (e.g., /Projects/...)
+# These appear in debug symbols and need to be mapped to the workspace root
+echo "settings append target.source-map \"/\" \"$execution_root/\""
+
 if [[ "${BAZEL_SEPARATE_INDEXBUILD_OUTPUT_BASE:-}" == "YES" ]]; then
   readonly output_base="${execution_root%/*/*}"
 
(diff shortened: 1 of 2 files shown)


adincebic (approved):

xcodeproj/internal/bazel_integration_files/create_lldbinit.sh
```diff
@@ -29,6 +29,10 @@ echo "settings append target.source-map ./external/ \"$BAZEL_EXTERNAL\""
 # `external` when set from swiftsourcefile
 echo "settings append target.source-map ./external/ \"$build_external\""
 
+# Workspace-relative paths with leading slash (e.g., /Projects/...)
+# These appear in debug symbols and need to be mapped to the workspace root
+echo "settings append target.source-map \"/\" \"$execution_root/\""
```
brentleyjones, line 34: This feels like it will hit too broad? There might be actual absolute paths as well that will now be mapped to the execution_root.

tinder-hemanthprasad: Addressed in b7b69fb1: replaced the broad `/` mapping with mappings generated only for top-level directories that actually exist in the workspace, so genuine absolute paths such as `/Users/...` retain their meaning. Verified with `bash -n` and a generated-output fixture that maps `/Projects/` and `/Sources/` while emitting neither `/` nor `/Users/`; the upstream Bazel/buildifier target was unavailable because this machine's `bazelisk` is a Tinder-workspace-only wrapper.

PR #3305, opened 2026-05-01
Resolve `bazel_env` earlier

brentleyjones:
This makes setting `"PATH": None` less painful by preventing dynamic values from busting the analysis cache. Before this change `defs.bzl` could change resulting in a lot of loading and analysis. Now just a single `bazel_env.sh` file changes, which is passed off to the installer and not part of the inner analysis.


aaronsky:
Does this mean that `xcodeproj(...)` doesn't take a `bazel_env` attr at all now? I couldn't see where else it was being fed in.

brentleyjones:
The API hasn't changed. It's just how those values are plumbed into the nested build and Xcode builds.

aaronsky:
Sure, but I can't see where that attr is actually used anymore now. I set a handful of env items in all of our xcodeproj rules to ensure certain bazel wrapper defaults are set – is that going to be affected by that?

xcodeproj/internal/xcodeproj_runner.bzl
brentleyjones, line 355: @aaronsky It's still set here.

xcodeproj/internal/templates/runner.sh
```diff
@@ -178,12 +178,23 @@ else
...
 fi
 
-touch "$pre_xcodeproj_bazelrc_dir/pre_xcodeproj.bazelrc"
+touch "$installer_inputs_dir/pre_xcodeproj.bazelrc"
+
+readonly bazel_env_file="$installer_inputs_dir/bazel_env.sh"
+{
+  echo "# Generated by rules_xcodeproj. Do not edit."
+  echo "envs=("
+  for env in "${envs[@]}"; do
```
brentleyjones, line 187: And written into `bazel_env.sh` here.

xcodeproj/internal/templates/bazel_build.sh
```diff
@@ -86,10 +86,12 @@ for var in "${allowed_vars[@]}"; do
   fi
 done
 
+source "$BAZEL_INTEGRATION_DIR/bazel_env.sh"
+
 bazel_cmd=(
   env -i
   "${passthrough_env[@]}"
-%bazel_env%
+  "${envs[@]}"
```
brentleyjones, line 94: And called/consumed here.

aaronsky (approved):

xcodeproj/internal/templates/bazel_build.sh
aaronsky: thanks

PR #3310, opened 2026-05-11
Add `generator_bazel_env`

brentleyjones:
Allows setting environment variables that affect only the nested generator call, but not the Xcode based calls. An example use case is setting `tools/bazel` output affecting env variables (e.g. color or verbosity).


xcodeproj/xcodeproj.bzl
```diff
@@ -365,6 +373,7 @@ def xcodeproj(
     if not bazel_path:
         bazel_path = "bazel"
     bazel_env = dict(bazel_env) if bazel_env else {}
+    generator_bazel_env = dict(generator_bazel_env) if generator_bazel_env else {}
```
thiagohmcruz, line 380: IIUC this is so you can mutate de dict a few lines below correct?

xcodeproj/xcodeproj.bzl
```diff
@@ -391,6 +400,12 @@ def xcodeproj(
         key: "\0" if value == None else value
         for key, value in sorted(bazel_env.items())
     }
+    generator_bazel_env = {
+        # Null character is used to represent `None`, since `attr.string_dict`
+        # requires non-`None` values.
+        key: "\0" if value == None else value
```
thiagohmcruz: Nit: feel free to ignore but should we make this `"\0"` char a constant since is being used in two places? I read first the usage in `xcodeproj_runner.bzl‎` and it wasn't clear to me the intent until I read this comment.

thiagohmcruz (approved):

adincebic (approved):

xcodeproj/xcodeproj.bzl
brentleyjones: Yep

brentleyjones:
This had a small bug. I'll open another PR to fix it.

brentleyjones:
https://github.com/MobileNativeFoundation/rules_xcodeproj/pull/3312

PR #3315, opened 2026-05-20
Buildable folders PoC

iMostfa:


adincebic:
Recently I built this internally but the performance was really bad. Adding / removing files would make Xcode hang for about 5-6 seconds.

iMostfa:
@adincebic interesting, i will test it for more time, you were using focused targets in that case ?

adincebic:
> @adincebic interesting, i will test it for more time, you were using focused targets in that case ? 

Nope, those are inconveniemt for us.

PR #3317, opened 2026-06-03
Fix address sanitizer being incorrectly applied to host tools

adincebic:
Enabling Address Sanitizer caused the app to fail to build because sanitizer flags were also being applied to host tools. Since host tools do not need to be sanitized when sanitizing the app, we use Bazel features to disable those flags for host tools while leaving the app target unchanged.

### DIFF
--- a/xcodeproj/internal/templates/generate_bazel_dependencies.sh
+++ b/xcodeproj/internal/templates/generate_bazel_dependencies.sh
@@ -147,32 +147,13 @@ fi
 # Runtime Sanitizers
 if [[ $apply_sanitizers -eq 1 ]]; then
   if [ "${ENABLE_ADDRESS_SANITIZER:-}" == "YES" ]; then
-    build_pre_config_flags+=(
-      --copt=-fno-omit-frame-pointer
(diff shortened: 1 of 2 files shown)


adincebic:
Perhaps we should not disable the features for host.

xcodeproj/internal/templates/generate_bazel_dependencies.sh
```diff
@@ -147,32 +147,13 @@ fi
...
-    build_pre_config_flags+=(
-      --copt=-fno-omit-frame-pointer
-      --copt=-fno-sanitize-recover=all
-      --copt=-fsanitize=address
-      --linkopt=-fsanitize=address
-      --%swiftcopt%=-sanitize=address
-      --copt=-Wno-macro-redefined
-      --copt=-D_FORTIFY_SOURCE=0
-    )
+    build_pre_config_flags+=(--features=asan --host_features=-asan)
```
brentleyjones, line 150: There is historical reasons for using the raw flags instead of the features. I don't recall what is it, but I think Git history will tell.

PR #3323, opened 2026-07-29
Replace bundled rsync with a clonetool based copy_tool

john-flanagan:
Take advantage of APFS clones when copying files out of the Bazel output base.

Also breaks our dependency on either system or bundled `rsync`.


xcodeproj/internal/bazel_integration_files/copy_tool.py
```diff
@@ -0,0 +1,216 @@
...
+
+        # Fix dest_dir's permissions and properties now that copying is done.
+        shutil.copystat(src_dir, dest_dir)
+        _chmod(dest_dir)
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--delete", action="store_true")
+    parser.add_argument("--exclude-from", default=None)
```
john-flanagan, line 197: We always pass `--delete` so I could remove this argument, but it seemed weird to me to have the `--exclude-from` argument without a explicit `--delete` argument too 🤷

xcodeproj/internal/templates/installer.sh
```diff
@@ -92,22 +92,17 @@ mkdir -p "$dest_xcschemes"
...
-  [[ -n "$rsync" ]] || fail \
-    "Failed to locate host rsync." \
-    "rules_xcodeproj's bundled rsync is only supported on macOS project generation."
 fi
 
-"$rsync" \
-  --archive \
-  --perms \
-  --chmod=u+w,F-x \
+python3 "$src_copy_tool" \
```
aaronsky: can we not rely on the tool’s own preference for how to be launched?

xcodeproj/internal/bazel_integration_files/BUILD
```diff
@@ -23,6 +23,11 @@ filegroup(
     visibility = ["//visibility:public"],
 )
 
+exports_files(
```
aaronsky: can this be a py_binary instead?

xcodeproj/internal/bazel_integration_files/copy_tool.py
```diff
@@ -0,0 +1,216 @@
...
+import errno
+import fnmatch
+import os
+import shutil
+import sys
+from ctypes import CDLL, c_char_p, c_int, get_errno
+
+_CLONEFILE = None
+_USE_CLONEFILE = sys.platform == "darwin"
+def _load_clonefile():
```
aaronsky: you’re mixing indentation styles in this file, which is unusual. prefer 4 spaces for indents, per PEP-8

xcodeproj/internal/bazel_integration_files/copy_tool.py
```diff
@@ -0,0 +1,216 @@
...
+
+        if result != 0:
+            # Recover fom `EXDEV` (cross-device) and `ENOTSUP` (destination
+            # filesystem doesn't support cloning) by disabling clonefile and
+            # falling back to copy2.
+            if get_errno() in (errno.EXDEV, errno.ENOTSUP):
+                _USE_CLONEFILE = False
+                shutil.copy2(src, dest)
+            else:
+                raise Exception(f"failed to clonefile {src} to {dest}")
```
aaronsky: prefer a custom or context-specific exception to Exception

xcodeproj/internal/bazel_integration_files/copy_tool.py
aaronsky, line 1: type hints in this file would be appreciated, as would tests for it.

aaronsky (changes requested):

xcodeproj/internal/bazel_integration_files/copy_tool.py
john-flanagan: Good catch. I copied this function [from rules_apple](https://github.com/bazelbuild/rules_apple/blob/af05e8bfb35c2f723b92d6b17ca218a0d5db385d/tools/bundletool/bundletool_experimental.py#L61-L72) and didn't update the indentation to match the rest of the file.

aaronsky (approved):
It looks fine but I haven't tested it on our project yet. Approving to unblock CI.

Issue #1953, opened 2023-04-03
Feature Request: Have a build mode that leverages an XCBBuildService proxy for a more integrated experience


brentleyjones:
We should support a new `xcodeproj.build_mode` value of `"proxy"`, which causes Xcode to use a custom XCBBuildService proxy.

Differences from BwB mode (not all of this has to land in the initial version):

- `BazelDependencies` target only shows Bazel output, output from Bazel targets that don't translate to Xcode targets, and possibly output from unfocused targets
  - Probably will be renamed
- `calculate_output_groups.py` isn't needed any more
  - Similar functionality will be part of of the proxy, just loaded into memory
- BEP is parsed/translated to create a Build Report
  - Each target is represented as an actual target
  - Dia files are parsed and sent over as diagnotics (#400 and #402), fixing Fix-Its
  - Swift parsable output is parsed (when a new version of https://github.com/bazelbuild/rules_swift/pull/304 lands), showing per-file progress
- Where possible, we no longer copy products to `DerivedData` (e.g. non-zipped binaries or bundles)
- If possible, `IndexingInfoResponse` points to `bazel-out/` per-library indexing data, preventing the need for `index-import`
  - This can only work if `-file-prefix-map` is used

Some open questions:

- How do we want to integrate the proxy with Xcode? And do we make some of these options for the user to decide?
  - Launch `Xcode.app/Contents/MacOS/Xcode` from `$SRCROOT` with `XCBBUILDSERVICE_PATH` set. This has the side benefit of fixing some relative path issues (e.g. #1284). Ideally we could modify the name and/or icon of the Xcode instance in the Dock, otherwise this could be confusing (though we could kill all versions of the same Xcode that are open...).
  - Modify the `Xcode.app` bundle. This has permissions/security issues.
- Is this the new default?
  - I assume no, at least to start, and possibly ever, because of the above integration issues
- How are unfocused targets represented in the Build Report?
    - We might be able to give them an ID that isn't associated with a target in the project. If that's the case, they aren't handled any differently. Otherwise we can have them be sub-outputs on `BazelDependencies`

This issue will serve as an Epic. Smaller focused issues will be made as needed.

thiagohmcruz:
Looks great at a high level! Some Qs/thoughts:

1. Are we planning to handle progress bar messages here as well?
[...]

PR #3328, opened 2026-08-08
Add a versioned build proxy manifest

karim-alweheshy:
# Summary

- Add a deterministic schema-v2 build-proxy manifest to generated Xcode projects.
- Bind project identity, target GUIDs, Bazel labels, target IDs, variants, operation capabilities, and output groups.
- Keep projects without proxy configuration unchanged.

# Stack

- Part of #1953.
- 1 of 5 in the rules_xcodeproj build-proxy stack.
- Parent: `main`.
- Next: the generated Bazel adapter layer.

# Validation

- `git diff --check` passes at this exact endpoint.
- The combined stack passes the repository test matrix; exact-layer CI is expected on this draft.

# Testing

Run the manifest analysis and generation tests, regenerate the Sanitizers fixture twice, and verify byte-identical schema-v2 output plus unchanged generation when proxy attributes are absent.

### DIFF
--- a/test/internal/build_proxy_manifest/BUILD
+++ b/test/internal/build_proxy_manifest/BUILD
@@ -1,0 +1,7 @@
+load("@rules_python//python:defs.bzl", "py_test")
+
+py_test(
+    name = "build_proxy_manifest_tests",
+    srcs = ["build_proxy_manifest_tests.py"],
+    deps = ["//xcodeproj/internal:build_proxy_manifest_lib"],
+)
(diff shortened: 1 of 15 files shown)


karim-alweheshy:
Rebased and normalized the complete build-proxy review stack onto current `main`.

For this root slice specifically, the old branch-only dependency/archive duplicates were dropped where already present on `main`; the three actual manifest commits were replayed patch-identically. The resulting head is `626d5f9d9dde4d9717c7145c992946bef0e0dd68` and GitHub reports it mergeable against current `main`.

The downstream review chain is now a single linear 23-commit DAG through #3339. All 12 intended edge relationships, changed-path sets, commit counts, dropped duplicates, and stable patch IDs were mechanically verified. Four focused build-proxy test targets pass on the final endpoint.

Because this repository uses squash-only merges, I am marking only this root slice ready. The downstream slices will remain draft and mechanically mergeable against their updated stack bases until each parent is promoted; then they can be retargeted/replayed one at a time.

PR #3338, opened 2026-08-11
Expose Bazel action starts to the build proxy

karim-alweheshy:
## Summary\n\n- emit a private allowlisted action-start stream for proxy-enabled builds\n- suppress raw subcommand environment and command blocks from Xcode-facing output\n- preserve direct and non-proxy rules_xcodeproj behavior\n\n## Stack\n\n- Base: the live Bazel progress PR\n- This PR adds live action lifecycle edges without changing the build manifest schema\n\n## Validation\n\n- focused launcher tests\n- shell syntax, ShellCheck, and Buildifier\n- live Xcode Reddit validation observed compile and link rows while Bazel was still running\n\n## Testing\n\nRun a clean proxy-enabled Xcode build and verify action rows open before completion, close exactly once, and do not expose raw environment values.


karim-alweheshy:
Closing this alias PR to keep one canonical review path.

Its branch is now an exact alias of #3336's endpoint and the same one-commit lifecycle delta is reviewed there. Keeping both open would duplicate the merge slice and previously caused shared-SHA check pollution. Continue with #3336, then #3339.

PR #3343, opened 2026-09-17
Fix Xcode navigation to symlinked external source files

adincebic:
Some libraries like [RxSwift](https://github.com/reactivex/rxswift) ship symlinks in their repositories. We noticed that it causes problems when navigating to the symlinked sources when they are in Bazel's external repositories directory via "jump to definition". Typically it works the first time and second time Xcode refuses to jump.

I traced the issue to the  resolution of the symlinks where we resolved the file symlink but leave the parent directory unresolved, which causes Xcode to open sources at paths that don't match their build settings and break navigation.

so we now resolve the parent directory symlinks too.


karim-alweheshy (approved):
LGTM

